PR #3, opened 2022-10-25
adding data description and tests

dyf:
Bringing in data description, incorporating validation logic on the names of raw and derived assets.


saskiad:
> @saskiad You'll probably want to weigh in on the inclusion of `acquisition_date` and `acquisition_time`. I had to add those to be able to generate/validate conventions for `DataDescription.name`.

what are the conventions for 'DataDescription.name'? Where did this come from?

I don't think we want acquisition date and time in the data description

src/aind_data_schema/data_description.py
```diff
@@ -0,0 +1,266 @@
...
+
+    ecephys = "ecephys"
+    ExASPIM = "ExASPIM"
+    SmartSPIM = "SmartSPIM"
+    mesoSPIM = "mesoSPIM"
+    ophys = "ophys"
+
+
+def datetime_tostring(d, t):
+    """take a date and time object, format it a as string"""
```
jtyoung84, line 67: This might be trickier than it seems. We should either include timezone info in the string representation, enforce formatting the string into utc, or use millis.

dyf: yeah good point. there are two concerns here - one is how we name data.  here I don't think we want to include TZ, unless there is a succinct way of doing that (the names are already getting long). that's what I'm using this function for.

within the JSON files I think we should use a standard string representation that includes TZ for sure. I haven't checked how pydantic renders datetime objects to strings.  let me look.

src/aind_data_schema/data_description.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+class Modality(Enum):
+    """see field description in DataDescription"""
+
+    ecephys = "ecephys"
+    ExASPIM = "ExASPIM"
+    SmartSPIM = "SmartSPIM"
+    mesoSPIM = "mesoSPIM"
+    ophys = "ophys"
```
jtyoung84, line 63: I may have been formatting my enums incorrectly, but I took a quick look at the python guide, and it suggests making enum keys all in screaming snake case: https://docs.python.org/3/library/enum.html#creating-an-enum

dyf: good idea, doing that

src/aind_data_schema/data_description.py
jtyoung84: I think if we convert all the date time objects to use utc, we can either add a 'Z' or we can document that the file name date times are in utc.

dyf: I will make these functions more specific `..._to_name_string` or something.

dyf: hmm. my suspicion is that if we switch to UTC in the data names that will be super confusing for our scientists. @saskiad?

jtyoung84: It might be fine as long as the datetime string isn't expected to be used anywhere

dyf: it's used to construct a unique name `{modality}_{subject}_{datestr}_{timestr}`

src/aind_data_schema/data_description.py
```diff
@@ -0,0 +1,231 @@
...
+class DataDescription(BaseModel):
+    schema_version: str = Field("0.1.0", title="Schema Version", const=True)
+    license: str = Field("CC-BY-4.0", title="License", const=True)
+    describedBy: str = Field(
+        "https://github.com/AllenNeuralDynamics/data_schema/blob/main/schemas/data_description.py",
+        title="Described by",
+        const=True,
+    )
+
+    acquisition_time: time = Field(
```
saskiad: replace with creation_time

src/aind_data_schema/data_description.py
```diff
@@ -0,0 +1,231 @@
...
+        ...,
+        description="Time of day that data acquisition started",
+        title="Acquisition Time",
+    )
+    acquisition_date: date = Field(
```
saskiad: replace with creation_date

saskiad: description should be "Date that ..."
and should it be when file is created? Eg. for derived data we want this to reflect when it was derived, correct?

src/aind_data_schema/data_description.py
```diff
@@ -0,0 +1,231 @@
...
+        ...,
+        description="Day that data acquisition started",
+        title="Acquisition Date",
+    )
+    label: str = Field(
+        ...,
+        description="Generic label for method of acquisition",
+        title="Label",
+    )
```
saskiad, line 106: how is this different from modality?

src/aind_data_schema/data_description.py
saskiad: also "Time of day" feels colloquial. We want this it very specific time, so we should specify if it's local time or UTC

src/aind_data_schema/data_description.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+class RawDataDescription(DataDescription):
+    modality: str = Field(
+        ...,
+        regex=DataRegex.NO_UNDERSCORES.value,
+        description="A short name for the specific manner, characteristic, pattern of application, or the employment of any technology or formal procedure to generate data for a study",
+        title="Modality",
+    )
+    subject_id: str = Field(
```
saskiad, line 229: I'm not sure this lives in data description.

src/aind_data_schema/data_description.py
```diff
@@ -0,0 +1,231 @@
...
+            label=label,
+            acquisition_date=acquisition_date,
+            acquisition_time=acquisition_time,
+            input_data=input_data,
+            **kwargs,
+        )
+
+
+class RawDataDescription(DataDescription):
+    modality: str = Field(
```
saskiad, line 223: see comment above about label. How is this different? And why unique to raw data?

src/aind_data_schema/data_description.py
```diff
@@ -0,0 +1,266 @@
...
+    license: str = Field("CC-BY-4.0", title="License", const=True)
+    describedBy: str = Field(
+        "https://github.com/AllenNeuralDynamics/aind-data-schema/blob/main/schemas/data_description.py",
+        title="Described by",
+        const=True,
+    )
+
+    creation_time: time = Field(
+        ...,
+        description="Time of day that data was created, used to uniquely identify the data",
```
saskiad, line 94: be specific about time. UTC? local time zone? Time of day is too colloquial

src/aind_data_schema/data_description.py
```diff
@@ -0,0 +1,266 @@
...
+        title="Creation Time",
+    )
+    creation_date: date = Field(
+        ...,
+        description="Day that data was created, used to uniquely identify the data",
```
saskiad, line 99: *date* that data was created

src/aind_data_schema/data_description.py
saskiad: the problem I see with UTC is that a use case is to analyze data relative to circadian cycle, and UTC makes that difficult.

src/aind_data_schema/data_description.py
dyf: For raw data, `label` = `modality_subject_id`.  I agree that label is a weird concept.  Let me think about it.

src/aind_data_schema/data_description.py
saskiad: I think we can leave TZ out of the date-time string, but include it as the unit for the time

src/aind_data_schema/data_description.py
dyf: For derived data, label would be e.g. `spikesorting-ks25`.  Do you want to also include subject id there?

src/aind_data_schema/data_description.py
dyf: I vote for UTC - okay?

src/aind_data_schema/data_description.py
dyf: how about "date that data was created"?

src/aind_data_schema/data_description.py
dyf: I need it to construct and validate names.  I could make it "private" so that it doesn't get saved out?

src/aind_data_schema/data_description.py
saskiad: that's good.

src/aind_data_schema/data_description.py
saskiad: UTC provided we have some information of what the local time zone is - eg. in the session metadata

saskiad (approved):

src/aind_data_schema/data_description.py
dyf: k, fixed

src/aind_data_schema/data_description.py
dyf: will fix this in `ephys_session`

PR #4, opened 2022-10-25
adding procedures schema

dyf:
Only minor changes here.  I moved `water_restriction` into its own subschema, but otherwise everything should be untouched.

### DIFF
--- a/src/aind_data_schema/procedures.py
+++ b/src/aind_data_schema/procedures.py
@@ -15,7 +15,7 @@ class ProtectiveMaterial(Enum):
     Duragel = "Duragel"
     SORTA_clear = "SORTA-clear"
     Kwik_Cast = "Kwik-Cast"
-    Other___see_notes = "Other - see notes"
+    Other = "Other - see notes"
 
 
 class Craniotomy(BaseModel):
@@ -279,13 +279,13 @@ class Procedures(BaseModel):
     """description of all procedures applied to subject"""
 
     describedBy: str = Field(
-        "https://github.com/AllenNeuralDynamics/data_schema/blob/main/schemas/procedures.py",
+        "https://github.com/AllenNeuralDynamics/aind-data-schema/blob/main/src/aind-data-schema/procedures.py",
         description="The URL reference to the schema.",
         title="Described by",
         const=True,
     )
     schema_version: str = Field(
-        "0.2.0", description="schema version", title="Version", const=True
+        "0.3.0", description="schema version", title="Version", const=True
     )
     specimen_id: str = Field(
         ...,
--- a/src/aind_data_schema/procedures.py
+++ b/src/aind_data_schema/procedures.py
@@ -29,12 +29,14 @@ class Craniotomy(BaseModel):
     )
     craniotomy_protocol_id: str = Field(..., title="Craniotomy protocol ID")
     craniotomy_coordinates_ml: float = Field(
-        ..., title="Craniotomy coordinate ML (mm)"
+        ..., title="Craniotomy coordinate ML (mm)", units="mm"
     )
     craniotomy_coordinates_ap: float = Field(
-        ..., title="Craniotomy coordinates AP (mm)"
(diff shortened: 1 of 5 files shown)


src/aind_data_schema/procedures.py
```diff
@@ -0,0 +1,319 @@
...
+
+    date: date = Field(..., title="Date")
+    experimenter_full_name: str = Field(
+        ...,
+        description="First and last name of the experimenter.",
+        title="Experimenter full name",
+    )
+    craniotomy_protocol_id: str = Field(..., title="Craniotomy protocol ID")
+    craniotomy_coordinates_ml: float = Field(
+        ..., title="Craniotomy coordinate ML (mm)"
```
saskiad, line 32: is there no way to declare units is pydantic?

dyf: there is, the converter just missed them.  adding back...

src/aind_data_schema/procedures.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+class HeadframeMaterial(Enum):
+    """headframe materials"""
+
+    Titanium = "Titanium"
+    Steel = "Steel"
+
+
+class WellType(Enum):
```
saskiad, line 55: let's remove the enum until we actually have real things here

src/aind_data_schema/procedures.py
```diff
@@ -0,0 +1,319 @@
...
+class MriScan(BaseModel):
+    """information about MRI scan"""
+
+    date: date = Field(..., title="Date")
+    experimenter_full_name: str = Field(
+        ...,
+        description="First and last name of the experimenter.",
+        title="Experimenter full name",
+    )
+    scan_sequence: MriScanSequence = Field(..., title="Scan sequence")
```
saskiad, line 164: why'd we change the name? not opposed, just curious why

src/aind_data_schema/procedures.py
```diff
@@ -0,0 +1,319 @@
...
+    Gelation = "Gelation"
+
+
+class TissuePrep(BaseModel):
+    """information about tissue prep procedure"""
+
+    name: TissuePrepName = Field(..., title="Name")
+    date_started: date = Field(..., title="Date started")
+    date_ended: Optional[date] = Field(None, title="Date ended")
+    experimenter_full_name: Optional[str] = Field(
```
saskiad, line 192: should not be optional

src/aind_data_schema/procedures.py
```diff
@@ -0,0 +1,319 @@
...
+    Immunostaining = "Immunostaining"
+    Gelation = "Gelation"
+
+
+class TissuePrep(BaseModel):
+    """information about tissue prep procedure"""
+
+    name: TissuePrepName = Field(..., title="Name")
+    date_started: date = Field(..., title="Date started")
+    date_ended: Optional[date] = Field(None, title="Date ended")
```
saskiad, line 191: probably shouldn't be optional. For some procedures it might be the same as date_started, but that's okay.
Actually, these should maybe the date-time?

src/aind_data_schema/procedures.py
dyf: resolved in the latest commit

saskiad (approved):
biggest concern is the units. Others are small fixes we can do later if necessary

src/aind_data_schema/procedures.py
dyf: Personal editorialization - I normally see T1/T2/RARE called sequences.  I can revert if you want.

src/aind_data_schema/procedures.py
dyf: fixed in latest commit

src/aind_data_schema/procedures.py
dyf: fixed

src/aind_data_schema/procedures.py
dyf: fixed

PR #5, opened 2022-10-25
adding imaging schemas

dyf:
Schemas for `Acquisition` and `Instrument`.

I created a base `Device` that is reused for everything with a model/manufacturer/serial#. Otherwise no other major changes.


adamkglaser:
Amazing, this looks good to me.

src/aind_data_schema/imaging/acquisition.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+    voxel_size: float = Field(..., title='Voxel size (um)')
+    volume_size: float = Field(
+        ...,
+        description='Size of the volume for this dimension.',
+        title='Volume size (um)',
+    )
+
+
+class Laser(BaseModel):
```
saskiad, line 60: there's a bunch of other fields for the laser that we moved into acquisition - I think this is based on outdated version

dyf: I think this matches the latest JSON schemas.  If they need changing let me know and I will re-generate this.

saskiad:
> I created a base Device that is reused for everything with a model/manufacturer/serial#. Otherwise no other major changes.

There was already a device object for additional devices - I don't see that in here.

src/aind_data_schema/imaging/acquisition.py
saskiad: this definitely isn't the latest JSON schema - have you had a chance to figure out what happened?

saskiad (approved):
I'll fix things when it's merged.

dyf:
> There was already a device object for additional devices - I don't see that in here.

Ah, that's what it was.  I repurposed that `Device` for all the other objects with the same device properties.

PR #26, opened 2022-10-31
making light cycle optional

saskiad:
since we've made light cycle the time on/off, there's no option of "unknown" so I made it optional.


dyf (approved):
looks good.  I bumped the version numbers FYI.

PR #52, opened 2022-11-04
Feat 24 coords

### LINKED ISSUE #24: update ephys coordinates
Update the definition of `EphysSession` coordinates to have `*_ml` `*_ap` `*_dv` in the name.

Acceptance Criteria:
- [ ] add a subclass that has the three named anatomical direction coordinate values in it
- [ ] do the same for `FieldCoordinates3dMm`
- [ ] update the tests
- [ ] patch version bump

mekhlakapoor:
Closes #24

### DIFF
--- a/src/aind_data_schema/ephys/ephys_session.py
+++ b/src/aind_data_schema/ephys/ephys_session.py
@@ -77,12 +77,12 @@ class Laser(BaseModel):
         ...,
         title="Targeted CCF coordinates",
     )
-    targeted_lab_coordinates: List[Field3dCoordinatesMm] = Field(
+    targeted_lab_coordinates: Field3dCoordinatesMm = Field(
         ...,
         description="Targeted coordinates relative to the headframe",
         title="Targeted lab coordinates",
     )
-    manipulator_coordinates: List[Field3dCoordinatesMm] = Field(
+    manipulator_coordinates: Field3dCoordinatesMm = Field(
         ...,
         title="Manipulator coordinates",
     )
--- a/src/aind_data_schema/ephys/ephys_session.py
+++ b/src/aind_data_schema/ephys/ephys_session.py
@@ -35,20 +35,24 @@ class CcfVersion(Enum):
 
     CCFv3 = "CCFv3"
 
-
+'''
 class Direction(Enum):
     """Axis direction name (TODO: combine with imaging axis)"""
 
     X = "X"
     Y = "Y"
     Z = "Z"
-
+'''
+class Direction(BaseModel):
+    X = "X"
+    Y = "Y"
+    Z = "Z"
 
 class Field3dCoordinatesMm(BaseModel):
     """Description of 3d coordinates in mm"""
 
-    direction: Direction = Field(..., title="Direction")
-    value: float = Field(..., title="Value (mm)", units="mm")
+    direction: Direction = Field(..., title="Direction Value", units="mm")
+    #value: float = Field(..., title="Value (mm)", units="mm")
 
 
 class AnatomicalDirection(Enum):
--- a/src/aind_data_schema/ephys/ephys_session.py
+++ b/src/aind_data_schema/ephys/ephys_session.py
@@ -6,7 +6,7 @@ from datetime import datetime
 from enum import Enum
 from typing import List, Optional
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, validator
 
 
 class SessionType(Enum):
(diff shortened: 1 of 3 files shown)


mekhlakapoor:
generates schema to look like this for coords: 

Screen Shot 2022-11-04 at 1 39 53 PM

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -35,35 +35,20 @@ class CcfVersion(Enum):
...
-    value: float = Field(..., title="Value (mm)", units="mm")
-
-
-class AnatomicalDirection(Enum):
-    """Anatomical direction name (TODO: combine with imaging direction)"""
-
-    ML = "ML"
-    AP = "AP"
-    DV = "DV"
+    X: float = Field(..., title="X (mm)", units="mm")
```
dyf, line 41: I think X/Y/Z/AP/DV/ML can be lower case

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -35,35 +35,20 @@ class CcfVersion(Enum):
...
+    Y: float = Field(..., title="Y (mm)", units="mm")
+    Z: float = Field(..., title="Z (mm)", units="mm")
 
 
 class CcfCoords(BaseModel):
```
dyf, line 46: now that it's organized like this, let's move the `ccf_version` field off of `EphysRg` and put it as a field on `CcfCoords`.  it can be required, default to 'CCFv3'.

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -35,35 +35,20 @@ class CcfVersion(Enum):
...
-
-class Direction(Enum):
-    """Axis direction name (TODO: combine with imaging axis)"""
-
-    X = "X"
-    Y = "Y"
-    Z = "Z"
-
-
 class Field3dCoordinatesMm(BaseModel):
```
dyf, line 38: I've been thinking the name `Field3dCoordinatesMm` is a little hokey. Can you rename it to `Coordinates3d`?

dyf: I'm thinking we may want to make `units` a field at some point, but let's do that later.

dyf (changes requested):
Looks good! A couple changes, please :).

src/aind_data_schema/ephys/ephys_session.py
dyf: (I guess it should actually default to `CcfVersion.CCFv3`

jtyoung84:
Other than the "describedBy" field, we want to resolve the jsonschema keys in snake case. This may require something like:
https://stackoverflow.com/questions/67995510/how-to-inflect-from-snake-case-to-camel-case-post-the-pydantic-schema-validation

dyf:
> (Maybe this can be opened as a separate issue) Other than the "describedBy" field, we want to resolve the jsonschema keys in snake case. This may require something like: https://stackoverflow.com/questions/67995510/how-to-inflect-from-snake-case-to-camel-case-post-the-pydantic-schema-validation

Ah yeah, this is default behavior for the `pydantic` `BaseModel.schema_json` method.  Let's investigate this in another thread.

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -88,27 +73,18 @@ class Laser(BaseModel):
...
-    targeted_ccf_coordinates: List[CcfCoords] = Field(
+    targeted_ccf_coordinates: CcfCoords = Field(
         ...,
-        max_items=3,
-        min_items=3,
         title="Targeted CCF coordinates",
-        unique_items=True,
     )
-    targeted_lab_coordinates: List[Field3dCoordinatesMm] = Field(
+    targeted_lab_coordinates: Field3dCoordinatesMm = Field(
```
saskiad: based on Josh's feedback we are removed the lab coordinates. I thought I had already removed that... did that get reverted? Either way, let's remove targeted_lab_coordinates.

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -121,27 +97,18 @@ class EphysProbe(BaseModel):
...
```
saskiad: same here for lab coordinates. We can remove later too if that's easier

saskiad (approved):

dyf (approved):
yay!

PR #73, opened 2022-11-15
centralizing device, manufacturer, etc

### LINKED ISSUE #39: Centralize `Device`
Right now the `Device` base class is a common concept between the `ephys` and `imaging` modules.  Let's move it up a level and share.

This will be true once `ophys` gets merged as well.

### LINKED ISSUE #40: Consistency between `ephys.EphysRig` and `imaging.Instrument`
Right now the `EphysRig` has an array of `Devices`, whereas `Instrument` has all of the devices at the top level.  Let's be consistent here.

### LINKED ISSUE #41: two `Laser` classes
Right now `ephys.ephys_session.Laser` and `imaging.instrument.Laser` have the same name.   These should probably have different names, and maybe share a common base class.

dyf:
resolves #41 
resolves #40 
resolves #39

### DIFF
--- a/src/aind_data_schema/ephys/ephys_rig.py
+++ b/src/aind_data_schema/ephys/ephys_rig.py
@@ -8,6 +8,8 @@ from typing import List, Optional
 
 from pydantic import BaseModel, Field
 
+from ..device import Device, DeviceManufacturer
+
 
 class HarpDevice(Enum):
     """Harp device name"""
@@ -28,13 +30,10 @@ class CameraName(Enum):
     FACE_CAMERA = "Face Camera"
 
 
-class Camera(BaseModel):
+class Camera(Device):
     """Description of camera"""
 
     name: CameraName = Field(..., title="Name")
-    manufacturer: str = Field(..., title="Manufacturer")
-    model: str = Field(..., title="Model")
-    serial_number: str = Field(..., title="Serial number")
     position_x: float = Field(..., title="Position X")
     position_y: float = Field(..., title="Position Y")
     position_z: float = Field(..., title="Position Z")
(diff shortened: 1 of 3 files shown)


dyf:
@saskiad You can see how as soon as `Device` becomes a shared concept, `Manufacturer` wants to as well. 

I did a (very fast) pass at updating ophys and ephys, there may be some mistakes.

dyf:
An additional thought is that we should remove `type` from the general `Device` class, since that is only ever getting overwritten by rig-specific classes.

src/aind_data_schema/ephys/ephys_rig.py
```diff
@@ -177,31 +170,12 @@ class ProbeType(Enum):
     MP_PHOTONIC_V1 = "MPI Photonic Probe (Version 1)"
 
 
-class Probe(BaseModel):
+class Probe(Device):
```
saskiad, line 173: this is something we need to figure out - because previous version, probe had name, type, serial_number. As a device they want manufacturer, model, serial_number...  Type can be model. But the list of types conflates manufacturer and model number... anyhow ...something to sort out.

saskiad (approved):

saskiad:
> An additional thought is that we should remove `type` from the general `Device` class, since that is only ever getting overwritten by rig-specific classes (and other times just doesn't exist).

Yeah, the DeviceType list is specific to microscopes and doesn't make sense being a field for a generic Device. We just need this list to get used appropriately.

src/aind_data_schema/ephys/ephys_rig.py
dyf: yes, I see. good to be consistent with how we use these terms if we can, I think.

saskiad (approved):

saskiad (approved):

PR #77, opened 2022-11-16
making label private

### LINKED ISSUE #75: DataDescription.label should be a private attribute
It's mostly a convenience for constructing names and such - it should not be exposed in the schema.

dyf:
resolves #75


dyf:
This takes `DataDescription.label` and makes it a private attribute so it doesn't show up in the schema.

saskiad (approved):

Issue #88, opened 2022-11-29
add filter position to filter spec class


adamkglaser:
````Lightsource```` class currently specifies matching filter by #. But this field does not also map over into the ````Filter```` class. So the filter position # cannot be linked the the actual filter specs.

Suggest added the same ````filter_wheel_index```` into the Filter class.

````python
class Filter(Device):
    """Description of a filter device"""

    type: FilterType = Field(..., title="Filter Type")
    diameter: float = Field(..., title="Size (mm)", ge=0)
    thickness: float = Field(..., title="Size (mm)", ge=0)
    filter_wheel_index: int = Field(..., title="Filter index")
    description: Optional[str] = Field(
        None, description="Where/how filter is being used", title="Description"
    )
````

Alternatives. Does it make sense to duplicate this same parameter into both classes? Or perhaps move the filter_wheel_index out of the ````Lightsource```` class and only into the ````Filter```` class.

**Comment by saskiad** [comment]
I guess I don't really know what the filter_wheel_index really means. I'm fine with moving it to filter. I take it that duplicating doesn't make sense?

**Comment by saskiad** [comment]
are there any filters that aren't in a filter wheel?

**Comment by dyf** [comment]
I think @adamkglaser is saying that he wants to associate a `Lightsource` with a `Filter`.  You could do that by having a common index between the two, or just moving the entire `Filter` description to `Lightsource`.  e.g.:

```
class Lightsource(Device):
   filter: Filter
```

Is that right @adamkglaser ?

dyf:
A `Filter` has a position on a filter wheel, and there is one filter wheel per `Lightsource` -- is this right?

If so, `filter_wheel_index` doesn't make sense on the `Lightsource` and should move to `Filter`.

saskiad:
perfect, I'll do this. And add a lightsource - filter item in the acquisition schema

adamkglaser:
great, once this and #87 and #86 are modified i will finalize and send over the schema

Issue #87, opened 2022-11-29
Modify specification of optical table size in imaging schema


adamkglaser:
Expand optical table dimension specification to tables with different width/lengths.

https://github.com/AllenNeuralDynamics/aind-data-schema/blob/e9e85c5732c3ec80779c9131fca531b72f84aaf6/src/aind_data_schema/imaging/instrument.py#L182-L183

Change spec from ````optical_table_size```` to ````optical_table_length```` and ````optical_table_width````.

Another option is to change ````optical_table_size```` to an array.

saskiad:
@adamkglaser what units do you specify table sizes in? meters? feet?

adamkglaser:
Sorry missed your response. For better or worse, the tables are usually specified in inches.

Issue #86, opened 2022-11-29
add additional devices to imaging schema


adamkglaser:
Motorized stage and DAQ are currently not exposed in a class. However they are a common item across SPIM systems and could be added as a dedicated class.

Suggested addition(s) to [instrument.py](https://github.com/AllenNeuralDynamics/aind-data-schema/blob/main/src/aind_data_schema/imaging/instrument.py)
````python
class MotorizedStage(Device):
    """Description of a motorized stage device"""

    manufacturer: str = Field(..., title="Manufacturer of device")  
    model: str = Field(..., title="Model of device")  
    axes: str = Field(..., title="Axes of device")
    travel: float = Field(..., title="Travel of device (mm)")

class DAQ(Device):
    """Description of a DAQ device"""

    manufacturer: str = Field(..., title="Manufacturer of device")  
    model: str = Field(..., title="Model of device")
    device_name: int = Field(..., title="PC device name")
    update_frequency: int = Field(..., title="Update frequency [hz]")
````

An alternative is to add them as an ````AdditionalImageDevice````. Although since they are common to all systems, I think this is not the best solution?

saskiad:
is it possible there will be multiple daq or motorizedstages for a given microscope? I believe no, but just want to make sure

saskiad:
"axes" refers to the axis that the motorized stage travels in? I'm wondering if we can use the Direction class in the acquisition schema here. Thoughts @dyf?

adamkglaser:
> is it possible there will be multiple daq or motorizedstages for a given microscope? I believe no, but just want to make sure

That is right, there would only be a single DAQ. For motorized stages, we could either bundle all of the axes into a single stage/device. Or separate them out into individual axes. The latter might make more sense since the travel would be different for all of them. Which if they were combined, would require travel to be an array.

saskiad:
I'm good making an array of stages - that's easy to do.

I'm concerned about the axes. How are we defining it? If it's relative to anatomy, that only makes sense in the context of an acquisition session where there is tissue. If it's relative to geometrical space, we need to figure out what that is in this schema.

adamkglaser:
The axes are confusing. I think the different permutations to choose from are.

1. Purely based on the hardware driver/devices. The stages from the manufacturer ASI all get assigned a lettered axis in the firmware. This is somewhat arbitrary though, and vary depending on how a specific instrument is assembled. This probably isn't a good option.
[...]

PR #90, opened 2022-11-30
Updating instrument and acquisition to better specify filters and stages, etc.

saskiad:
closes issues #88, #87, #86

### DIFF
--- a/src/aind_data_schema/imaging/instrument.py
+++ b/src/aind_data_schema/imaging/instrument.py
@@ -180,18 +180,18 @@ class DAQ(Device):
 
 class OpticalTable(Device):
     """Description of Optical Table"""
-    optical_table_length: Optional[float] = Field(
-        None, title="Optical table length (inches)", units="inches", ge=0
+    length: Optional[float] = Field(
+        None, title="Length (inches)", units="inches", ge=0
     )
-    optical_table_width: Optional[float] = Field(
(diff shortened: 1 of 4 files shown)


src/aind_data_schema/imaging/instrument.py
```diff
@@ -157,6 +157,36 @@ class AdditionalImagingDevice(Device):
...
+    update_frequency: float = Field(
+        ..., title="DAQ update frequency (Hz)", units="Hz"
+    )
+    daq_ao_names_to_channel: Optional[List[AnalogOutput]] = Field(
+        None, title="DAQ analog outputs", unique_items=True
+    )
+
+class OpticalTable(Device):
+    """Description of Optical Table"""
+    optical_table_length: Optional[float] = Field(
```
dyf, line 183: I think we could just call these `length` and `width`

dyf (changes requested):
minor change

dyf:
Oh also, go ahead and do a minor version bump of `Instrument` and `Acquisition`

src/aind_data_schema/imaging/instrument.py
dyf: Also, I think `model` is gone.  Is that on purpose?

dyf (approved):

PR #103, opened 2022-12-06
new examples, rejecting extra fields

dyf:
I updated the examples to include aind's smartspim and exaspim.

I also made the schemas reject extra fields (Adam made up some field names). This is a nontrivial change, so doing a minor version bump.


examples/exaspim_instrument.py
```diff
@@ -0,0 +1,162 @@
...
+            manufacturer="National Instruments",
+            device_name="Dev2",
+            update_frequency="10000",
+            number_active_channels=7,
+            serial_number="Unknown",
+        )
+    ],
+    motorized_stages=[
+        instrument.MotorizedStage(
+            axis="Detection axis",
```
dyf, line 100: @adamkglaser help

saskiad:
Should manufacturer here be "other" if this is a one-off collaboration? And then add the names in the notes?

dyf:
Sorry, this one got a little out of hand.  Turning on a strict "no extra fields" policy was trickier than I thought and broke something unnecessarily fancy I was doing in `DataDescription`.

examples/exaspim_instrument.py
```diff
@@ -0,0 +1,161 @@
...
+    ],
+    light_sources=[
+        instrument.Lightsource(
+            type="laser",
+            coupling="SMF",
+            wavelength=405,
+            max_power=200,
+            serial_number="LAS-08307",
+            manufacturer="Oxxius",
+            notes="Housed in commercial laser combiner, S/N L6CC-00513",
```
saskiad, line 41: Should this laser combiner be an additional device?

adamkglaser: This is a good idea I had overlooked. Most of the SPIM systems (also Smrithi's system for Ophys) will have all lasers together in some combiner. Same for Anna/Galen's lasers in ephys.

examples/exaspim_instrument.py
adamkglaser, line 99: I can either add this into the model name, i.e. "MS-8000-X" but there is actually some merit in keeping around the native stage axis, in addition to how we've defined the instrument axis. If not adding to the model name, another field such as hardware_stage_axis or native_stage_axis. The encoders PID settings etc. can be different and it would be useful if needed to backout what lettered axis was which.

adamkglaser, line 99: this is correct.

examples/exaspim_instrument.py
```diff
@@ -0,0 +1,162 @@
...
+        instrument.MotorizedStage(
+            axis="Detection axis",
+            travel=1000,
+            model="MS-8000",
+            manufacturer="Applied Scientific Instrumentation",
+            daq_channel=DaqChannel(index=2, type="Analog Output"),
+            serial_number="Unknown",
+        ),
+        instrument.MotorizedStage(
+            axis="Illumination axis",
```
adamkglaser, line 107: This should be perpendicular axis.

examples/exaspim_instrument.py
```diff
@@ -0,0 +1,162 @@
...
+            travel=1000,
+            model="MS-8000",
+            manufacturer="Applied Scientific Instrumentation",
+            serial_number="Unknown",
+        ),
+        instrument.MotorizedStage(
+            axis="Perpendicular axis",
```
adamkglaser, line 114: This should be illumination axis.

saskiad (approved):

Issue #106, opened 2022-12-07
Update ephys session types


dyf:
Let's get these up to date for the workshop

dyf:
closed by #109

PR #109, opened 2022-12-08
ephys updates

saskiad:
updating ephys schema based on feedback from Josh


saskiad:
also closes issue #106

src/aind_data_schema/ephys/ephys_rig.py
```diff
@@ -23,6 +23,11 @@ class HarpDevice(Enum):
     INPUT_EXPANDER = "Input Expander"
 
 
+class HarpDevice(Device):
+    """Describes a Harp device"""
+    name: HarpDeviceName = Field(..., title="Name")
+    device_version: str = Field(..., title="Device version")
```
dyf, line 29: do we want version on top level `Device`?

src/aind_data_schema/ephys/ephys_rig.py
```diff
@@ -176,6 +215,23 @@ class EphysProbe(Device):
...
+    probe_manipulator: Manipulator = Field(..., "Manipulator")
+    calibration_data: str = Field(
+        ..., title="Calibration data", 
+        description="Path to calibration data"
+    )
+    calibration_date: Optional[datetime] = Field(
+        None, title="Calibration date"
+    )
+
+class DAQ(Device):
```
dyf, line 227: I think this is identical to the `DAQ` in imaging.  Move this definition to `device.py` and share?

dyf (changes requested):
Couple small things.

src/aind_data_schema/ephys/ephys_rig.py
saskiad: I don't think so... I think for most devices this shows up as model numbers. Not sure why they are version numbers for Harp

src/aind_data_schema/ephys/ephys_rig.py
saskiad: yes

src/aind_data_schema/ephys/ephys_rig.py
dyf: bueno.

dyf (approved):

PR #114, opened 2022-12-10
Feat 69 default filename

### LINKED ISSUE #69: Map schema model to default json file name
# User story
As a user, I want a standardized name for my json file instances that validate against a schema, so I can ensure it is consistent across all projects.

# Acceptance criteria
- [x] Attach default json file name attribute to the Pydantic schema models
- [x] (Stretch) Add a json writer method to the BaseModel class that a use can do something like:
```
p = Processing(...)
p.write("directory to write to", "optional file name if default not desired")
```

# Sprint Ready Checklist 
1. - [ ] Acceptance criteria defined 
2. - [ ] Team understands acceptance criteria 
3. - [ ] Team has defined solution / steps to satisfy acceptance criteria 
4. - [ ] Acceptance criteria is verifiable / testable 
5. - [ ] External / 3rd Party dependencies identified
6. - [ ] Ticket is prioritized and sized

# Notes
- Let's make this a private attribute, so it doesn't show up in the JSONSchema
- Add the writer method and private attribute to `base.AindSchema`, now that it exists.

mekhlakapoor:
closes #69


examples/aind_smartspim_instrument.py
```diff
@@ -232,5 +232,5 @@
     temperature_control=False,
 )
 
-with open("aind_smartspim_instrument.json", "w") as f:
-    f.write(inst.json(indent=3))
+dir_path = os.path.dirname(os.path.realpath(__file__))
+inst.write_to_json(dir_path, "aind_smartspim")
```
dyf, line 236: Can we have the examples just do `write_to_json()`?

src/aind_data_schema/base.py
```diff
@@ -58,3 +58,23 @@ def __init_subclass__(cls, optional_fields=None, **kwargs):
         )
         field.field_info.const = True
         cls.__fields__.update({"describedBy": field})
+
+    def write_to_json(cls, dir=os.getcwd(), optional_name=''):
```
dyf, line 62: Let's have `dir` default to `'.'`, I think.

dyf (changes requested):
I'd like the examples to be very simple, so let's not include the stuff about checking dirpath.

examples/aind_smartspim_instrument.py
mekhlakapoor: sure, but it wont create the example files as expected in the examples directory (for ex the extra naming)

jtyoung84 (approved):

dyf (approved):

PR #125, opened 2022-12-21
updates based on feedback from ephys team

### LINKED ISSUE #122: more ephys changes

dyf:
resolves #122


src/aind_data_schema/device.py
```diff
@@ -45,12 +45,15 @@ class DaqChannelType(Enum):
...
     AO = "Analog Output"
+    DI = "Digital Input"
+    AI = "Analog Input"
 
 
 class DaqChannel(AindModel):
     """Description of a DAQ Channel"""
 
-    index: int = Field(..., title="index")
+    channel_name: str = Field(None, title="DAQ channel name")
```
saskiad, line 55: can this be channel_index?

src/aind_data_schema/ephys/ephys_rig.py
```diff
@@ -147,37 +197,30 @@ class LaserModule(Device):
...
-        ..., title="Angle pitch (deg)", units="deg", ge=0, le=360
-    )
-    angle_yaw: float = Field(
-        ..., title="Angle yaw (deg)", units="deg", ge=0, le=360
-    )
-    angle_roll: float = Field(
-        ..., title="Angle roll (deg)", units="deg", ge=0, le=360
-    )
+    refresh_rate: int = Field(..., title="Refresh rate", ge=60)
+    regresh_rate_unit: str = Field("hertz", title="Refresh rate unit")
```
saskiad, line 210: typo (refresh)

Also, I'm not sure I'm on board with separating out units like this. I find it really really clunky

saskiad: Not just clunky, but it's going to lead to having different units in different files, making it hard to do meaningful searches. We need to define the units, not ask for them.

src/aind_data_schema/ephys/ephys_rig.py
```diff
@@ -147,37 +197,30 @@ class LaserModule(Device):
...
+    size: Size2d = Field(..., title="Monitor size")
+    position: Coordinates3d = Field(..., title="Monitor position")
+    orientation: Orientation3d = Field(..., title="Monitor orientation")
+    viewing_distance: float = Field(..., title="Viewing distance")
+    viewing_distance_unit: SizeUnit = Field(SizeUnit.CM, title="Size unit")
```
saskiad, line 215: shouldn't title be viewing distance unit? Not size?

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -37,35 +38,32 @@ class CcfVersion(Enum):
...
-    wavelength: int = Field(..., title="Wavelength (nm)", units="nm")
-    power: float = Field(..., title="Power (mW)", units="mW")
-    targeted_structure: str = Field(..., title="Targeted structure")
-    targeted_ccf_coordinates: CcfCoords = Field(
-        ..., title="Targeted CCF coordinates",
-    )
+    wavelength: int = Field(..., title="Wavelength")
+    wavelength_unit: SizeUnit = Field(SizeUnit.NM, title="Wavelength unit")
+    power: float = Field(..., title="Power")
+    ower_unit: str = Field("milliwatt", title="Maximum power unit")
```
saskiad, line 58: typo: power

examples/aibs_smartspim_instrument.json
```diff
@@ -9,69 +9,69 @@
    "humidity_control": false,
    "optical_tables": [
       {
-         "manufacturer": "MKS Newport",
+         "name": null,
```
saskiad, line 12: I might be lost, but the name is to link to DAQ. And is optional. So for non-daq connected items, can we just not have a name rather than have it be null?

examples/aind_smartspim_instrument.py
```diff
@@ -86,7 +86,7 @@
     ],
     fluorescence_filters=[
         instrument.Filter(
-            type="Band pass",
+            filter_type="Bandpass",
```
saskiad, line 89: shouldn't this be "Band pass" as two words? (this is probably a comment for a different location....)

examples/exaspim_instrument.json
```diff
@@ -103,91 +91,143 @@
...
          "notes": null,
-         "device_name": "Dev2",
-         "update_frequency": 10000.0,
-         "number_active_channels": 7
+         "data_interface": "USB",
+         "computer_name": "Dev2",
+         "channels": [
+            {
+               "channel_name": "3",
+               "device_name": "LAS-08308",
```
saskiad, line 153: this doesn't map onto a name. This is a serial number. We need this to map to names, which are all null currently.

examples/exaspim_instrument.json
```diff
@@ -103,91 +91,143 @@
...
+               "channel_type": "Analog Output",
+               "port": null,
+               "channel_index": null,
```
saskiad, line 156: channel index can't be null. If there's a daq channel, it has an index

examples/exaspim_instrument.py
```diff
@@ -83,13 +79,21 @@
         )
     ],
     daqs=[
-        instrument.DAQ(
+        DAQDevice(
             model="PCIe-6738",
+            data_interface="USB",
+            computer_name="Dev2",
```
saskiad, line 85: this is a thought: do we need information on the computer?

src/aind_data_schema/device.py
```diff
@@ -11,64 +17,257 @@
...
+    PCIE = "PCIe"
+    PXI = "PXI"
+    ETH = "Ethernet"
+    OTHER = "Other"
+
+
+class FilterType(Enum):
+    """Filter type"""
+
+    BANDPASS = "Bandpass"
```
saskiad, line 77: These should be two words

src/aind_data_schema/subject.py
```diff
@@ -38,10 +38,14 @@ class LightCycle(AindModel):
     """Description of vivarium light cycle times"""
 
     lights_on_time: time = Field(
-        ..., description="Time in UTC that lights were turned on", title="Lights on time",
+        ...,
+        description="Time in UTC that lights were turned on",
```
saskiad, line 42: not something for this PR, but if light cycle is in UTC then we need all times in UTC, and I don't think that's how they are being used currently.

saskiad (commented):
I think things are okay - I had super minor comments that are things we can fix in the future. Oh, except the examples not having names for the daq. But maybe that can be fixed in the future too?

saskiad (approved):

jsiegle (approved):
Just made a minor change to the validator docstrings. Looks good now!

PR #135, opened 2022-12-23
fixing bug with data description

dyf:


src/aind_data_schema/__init__.py
```diff
@@ -1,7 +1,7 @@
 """ imports for AindModel subclasses
 """
 
-__version__ = "0.3.10"
+__version__ = "0.3.11"
```
jtyoung84, line 4: The version should be managed automatically now

dyf: Jon so I undo the version change?

jtyoung84 (changes requested):
No need for manual version bump

src/aind_data_schema/__init__.py
jtyoung84: yeah

dyf: (done)

jtyoung84 (approved):

PR #137, opened 2022-12-24
Bug 136 snakecase

### LINKED ISSUE #136: Get default file name should return snake case
**Describe the bug**
The `_get_default_filename()`  method should convert the CamelCase class names to snake_case.

**To Reproduce**
1. Import the `DataDescription` class.
2. Run: `rd = DataDescription.construct()` and then `rd._get_default_filename()`
3. Notice the output is `datadescription.json`

**Expected behavior**
The output should be `data_description.json` (snake case)

**Additional context**
- There's also an issue where the default filename for the `RawDataDescription` class should also be `data_description.json`
- `_get_default_filename()` should check superclasses for a direct subclass of `AindCoreModel` and use that class's name when generating the filename.

mekhlakapoor:
closes #136

### DIFF
--- a/src/aind_data_schema/base.py
+++ b/src/aind_data_schema/base.py
@@ -4,7 +4,7 @@ import inspect
 import os
 import urllib.parse
 import re
-import logging
+import warnings
 
 from pydantic import BaseModel, Extra
 from pydantic.fields import ModelField
@@ -63,11 +63,13 @@ class AindCoreModel(AindModel):
         Returns string
         """
         bases = list(cls.__bases__)
+        if not bases:
+            warnings.warn("Warning: empty list")
+            return cls.__name__
+
         while bases[0] is not AindCoreModel:
             cls = bases[0]
             bases = list(cls.__bases__)
-        # Handle weird case where this degenerates to an empty list
-        # log a warning or error and return cls?
         return cls.__name__
 
     @classmethod
--- a/src/aind_data_schema/base.py
+++ b/src/aind_data_schema/base.py
@@ -3,6 +3,7 @@
 import inspect
 import os
 import urllib.parse
+import re
 
 from pydantic import BaseModel, Extra
 from pydantic.fields import ModelField
@@ -40,7 +41,7 @@ class AindCoreModel(AindModel):
     describedBy: str
 
     def __init_subclass__(cls, optional_fields=None, **kwargs):
-        """add the describedby field to all subclasses"""
+        """Add the describedby field to all subclasses"""
         super().__init_subclass__(**kwargs)
 
         value = build_described_by(cls)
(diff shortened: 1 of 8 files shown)


src/aind_data_schema/base.py
```diff
@@ -54,16 +55,27 @@ def __init_subclass__(cls, optional_fields=None, **kwargs):
         field.field_info.const = True
         cls.__fields__.update({"describedBy": field})
 
+    def _get_direct_subclass(self, cls):
+        """
+        Check superclasses for a direct subclass of AindCoreModel
+        """
+        for base in cls.__bases__:
```
jtyoung84, line 62: It doesn't look like you need a for loop here. If it's assumed that AindCoreModel will eventually be the only item in the list, you can rewrite this as something like:
```
bases = cls.__bases__
if not bases:
  # Handle weird case where this degenerates to an empty list
  # log a warning or error and return cls?
elif AindCoreModel is bases[0]:
  return cls
else:
  return self._get_direct_subclass(bases[0])
```
This function is small and simple enough to leave it as it is. There are a lot of considerations when using recursion though (look up tail recursion for example and max iterations for recursion in python). This function would be easier to maintain if it was written as a while loop though. It'd have the added benefit of being a static method and returning a string instead of class.

tests/test_base.py
```diff
@@ -30,6 +30,14 @@ def test_get_default_filename(self):
         actual_default_filename = p._get_default_filename()
         self.assertEqual(expected_default_filename, actual_default_filename)
 
+    def test_get_default_filename_subclass(self):
+        """tests that default filename returns as expected"""
+
+        rd = RawDataDescription.construct()
```
jtyoung84, line 36: Can you add several more test cases? You can loop through them like:
```
test_models = [(RawDataDescription, "data_description.json"), (Subject, "subject.json"),...]
for model, expected_name in test_models:
   m = model.construct()
   actual_name = m._get_default_filename()
   self.assertEqual(expected_name, actual_name)
```

jtyoung84 (changes requested):
Add more examples in the unit test.
Handle edge case where list might be empty in `_get_default_filename()`
Reconsider using recursion, although I'll probably approve it if you don't want to change it.

src/aind_data_schema/base.py
```diff
@@ -54,16 +55,27 @@ def __init_subclass__(cls, optional_fields=None, **kwargs):
...
+            if base is AindCoreModel:
+                return cls
+            else:
+                return self._get_direct_subclass(base)
+
     def _get_default_filename(self):
```
jtyoung84, line 68: It'd be nice if this was a static method so that it can be called without constructing an instance.

jtyoung84 (approved):

src/aind_data_schema/base.py
```diff
@@ -54,23 +56,43 @@ def __init_subclass__(cls, optional_fields=None, **kwargs):
...
+            new_cls = pydantic_class
+            bases = new_cls.__bases__
+            first_base = bases[0]
+            while bases and AindCoreModel is not first_base:
+                new_cls = bases[0]
+                bases = new_cls.__bases__
+                first_base = bases[0]
+            return new_cls
+        except IndexError:
+            logging.error(f"Unable to construct filename for {pydantic_class}")
```
dyf, line 74: this should probably more like "Unable to find direct AindCoreModel subclass" or something.

dyf (changes requested):
small comment about log message

dyf (approved):
LGTM!

PR #138, opened 2022-12-30
Consolidate device classes; update ephys schemas

jsiegle:
Lots of changes here, happy to chat more about any of them! Since there are also some changes to `ophys_rig.py` and `instrument.py`, the ophys and imaging teams should be consulted before merging.

### Consolidating classes across projects

There were a number of classes (e.g. `Laser`, `Camera`) that were duplicated across modules. I moved these to `devices.py` and made sure they contained all of the original fields were present in the merged version. This should help ensure that schema are more consistent across projects. I would recommend putting each device type in a separate file (e.g. `lasers.py`, `cameras.py`), but I didn't want to make that change yet. There are also some classes that will be needed across projects that still live in the ephys rig schema (e.g. `HarpDevice`, `MousePlatform`). We should probably move these out sooner rather than later.

### Updated DAQ conventions

Rather than linking each device to one DAQ channel, I've changed it so each DAQ channel is associated with one device. I think this is better because there are some devices (e.g. lenses) that don't have DAQ channels, and some devices that have multiple DAQ channels. Ideally there could be some validation step to ensure that each DAQ channel refers to a device that exists elsewhere in the rig description, but this may be tricky.

### Device-specific manufacturer enums

Each device type in `devices.py` now has a separate manufacturer enum. This is not strictly necessary, but it will make it much easier to manually enter new devices. After many years of adding devices to rigs, the list of all manufacturers will become quite long.

### Naming devices

Devices such as probes and cameras are now named using open-ended strings rather than enums, because I don't think consecutive letter-based naming conventions are going to give us sufficient flexibility. This requires that whoever is creating the session JSON checks that data stream names match devices in the rig JSON. Eventually, it would be great if the form could ingest the rig JSON and display all of the valid stream names.

### Other changes

I tried to avoid using `type` as a field name, because this is a Python built-in function. Same with `filter`, but I didn't change that yet.

### Example files

Here are two example JSON files I generated with the latest changes:

[ephys_rig.json](https://github.com/AllenNeuralDynamics/aind-data-schema/files/10326500/ephys_rig.txt)
[ephys_session.json](https://github.com/AllenNeuralDynamics/aind-data-schema/files/10326501/ephys_session.txt)


src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -15,19 +15,7 @@ class SessionType(Enum):
...
-    """Names of data streams to expect in recording"""
-
-    NEUROPIXELS_PROBES = "Neuropixels probes"
-    BODY_CAMERA = "Body camera"
-    FACE_CAMERA = "Face camera"
-    EYE_CAMERA = "Eye camera"
-    BONSAI_FILE = "Bonsai file"
-    HARP_FILE = "Harp bin file"
-    OTHER = "Other"
+    RF_MAPPING = "Receptive field mapping"
```
saskiad, line 18: is this a data stream?

src/aind_data_schema/device.py
```diff
@@ -47,36 +140,151 @@ class DaqChannelType(Enum):
...
+    )
+    
+    x: Optional[float] = Field(None, title="Position X (mm)", units="mm")
+    y: Optional[float] = Field(None, title="Position Y (mm)", units="mm")
+    z: Optional[float] = Field(None, title="Position Z (mm)", units="mm")
+    
+    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
 
 
 class DeviceBase(AindModel):
```
dyf, line 163: @jsiegle I think the refactor of DAQ channels would mean that `Device` can be deleted and `DeviceBase` renamed to `Device`.  Is that correct?

src/aind_data_schema/device.py
```diff
@@ -47,36 +140,151 @@ class DaqChannelType(Enum):
...
+class Lens(DeviceBase):
+    """Lens used to focus light onto a camera sensor"""
+
+    # required fields
+    lens_manufacturer: LensManufacturer = Field(..., title="Lens manufacturer")
+
+    # optional fields
+    focal_length: Optional[float] = Field(None, title="Focal length of the lens", units="mm")
+    size: Optional[LensSize] = Field(None, title="Size (inches)")
+    optimized_wavelength_range: Optional[str] = Field(
```
dyf, line 199: do we want this to be a string?

src/aind_data_schema/device.py
```diff
@@ -47,36 +140,151 @@ class DaqChannelType(Enum):
...
+    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
 
 
 class DeviceBase(AindModel):
-    """Description of a general device"""
+    """Generic device"""
 
-    manufacturer: Manufacturer = Field(..., title="Manufacturer")
-    serial_number: str = Field(..., title="Serial number")
+    device_name: Optional[str] = Field(None, title="Device name")
```
dyf, line 166: I don't think `name` is a keyword - can we just use `name` throughout, unless a model has multiple names to keep track of?

src/aind_data_schema/device.py
```diff
@@ -17,27 +17,120 @@ class Manufacturer(Enum):
...
     VIEWORKS = "Vieworks"
     VORTRAN = "Vortran"
 
 
+class LensManufacturer(Enum):
+    """Camera manufacturer"""
+
+    EDMUND_OPTICS = "Edmund Optics"
+    BASLER = "Basler"
+    THORLABS = "Thorlabs"
```
dyf, line 43: Wondering if we can refer to `Manufacturer.THORLABS` here and other places where the same manufacturer is in multiple enums to reduce number of spelling mistakes.

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -53,46 +41,53 @@ class CcfCoords(AindModel):
...
     """Stream of data with a start and stop time"""
 
-    stream_start_time: datetime = Field(..., title="Stream start time")
-    stream_stop_time: datetime = Field(..., title="Stream stop time")
+    start_time: datetime = Field(..., title="Stream start time")
+    end_time: datetime = Field(..., title="Stream stop time")
     probes: List[EphysProbe] = Field(..., title="Probes", unique_items=True)
     lasers: List[LaserModule] = Field(..., title="Lasers", unique_items=True)
+    daqs: List[DAQ] = Field(..., title="DAQs", unique_items=True)
+    cameras: List[Camera] = Field(..., title="Cameras", unique_items=True)
```
dyf, line 90: This is just a list of names.  Let's just make `Stream.cameras` it a list of strings and get rid of `ephys_session.Camera`?

dyf (commented):
Saskia is approver, added some thoughts.

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -53,46 +41,53 @@ class CcfCoords(AindModel):
...
+
+    name: str = Field(..., title="Camera name (must match rig JSON)")
 
 
 class Stream(AindModel):
     """Stream of data with a start and stop time"""
 
-    stream_start_time: datetime = Field(..., title="Stream start time")
-    stream_stop_time: datetime = Field(..., title="Stream stop time")
+    start_time: datetime = Field(..., title="Stream start time")
```
saskiad, line 85: I think we still need to differentiate stream start time and session start time. Same for stop time. While it should be clear because this is within the stream class, I think it's better to disambiguate them

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -53,46 +41,53 @@ class CcfCoords(AindModel):
...
+    name: str = Field(..., title="Laser module name (must match rig JSON)")
+    
 
-class EphysProbe(AindModel):
-    """Description of an ephys probe"""
+class EphysProbe(ManipulatorModule):
+    """Probe recorded in a Stream"""
 
-    name: str = Field(..., title="Name")
-    tip_targeted_structure: str
```
saskiad, line 77: why are we removing the tip targeted structure?

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -53,46 +41,53 @@ class CcfCoords(AindModel):
...
-        None,
-        title="Targeted CCF coordinates",
-    )
-    manipulator_coordinates: Coordinates3d = Field(
-        ...,
-        title="Manipulator coordinates",
-    )
+    
+
+class DAQ(AindModel):
```
saskiad, line 70: I'm a little confused about why we're adding this here. there is no session specific configuration information, just which daq (and camera) is used. Correct? And that already lives in the rig schema. No?

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -53,46 +41,53 @@ class CcfCoords(AindModel):
...
 class Stream(AindModel):
     """Stream of data with a start and stop time"""
 
-    stream_start_time: datetime = Field(..., title="Stream start time")
-    stream_stop_time: datetime = Field(..., title="Stream stop time")
+    start_time: datetime = Field(..., title="Stream start time")
+    end_time: datetime = Field(..., title="Stream stop time")
     probes: List[EphysProbe] = Field(..., title="Probes", unique_items=True)
     lasers: List[LaserModule] = Field(..., title="Lasers", unique_items=True)
+    daqs: List[DAQ] = Field(..., title="DAQs", unique_items=True)
```
saskiad, line 89: see above comment

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -53,46 +41,53 @@ class CcfCoords(AindModel):
...
+    targeted_ccf_coordinates: Optional[CcfCoords] = Field(
+        None,
         title="Targeted CCF coordinates",
     )
     manipulator_coordinates: Coordinates3d = Field(
         ...,
         title="Manipulator coordinates",
     )
 
+class LaserModule(ManipulatorModule):
```
saskiad, line 57: this is missing wavelength and power

src/aind_data_schema/device.py
```diff
@@ -17,27 +17,121 @@ class Manufacturer(Enum):
...
-    TELEDYNE_FLIR = "Teledyne FLIR"
-    THORLABS = "Thorlabs"
     SEMROCK = "Semrock"
+    THORLABS = "Thorlabs"
     VIEWORKS = "Vieworks"
     VORTRAN = "Vortran"
 
 
+class LensManufacturer(Enum):
+    """Camera manufacturer"""
```
saskiad, line 39: Lens manufacturer?

src/aind_data_schema/device.py
```diff
@@ -17,27 +17,121 @@ class Manufacturer(Enum):
...
+    NI = "NI"
+    IMEC = "imec"
+    OEPS = "OEPS"
+
+
+class Coupling(Enum):
+    """Laser coupling type"""
+
+    FREE_SPACE = "Free-space"
+    SMF = "SMF"
```
saskiad, line 85: what is SMF and MMF?

src/aind_data_schema/device.py
```diff
@@ -47,36 +141,147 @@ class DaqChannelType(Enum):
...
-class Device(DeviceBase):
-    """Description of a device that communicates with a DAQ"""
+class Camera(DeviceBase):
+    """A device that acquires images and streams them to a PC"""
 
-    daq_channels: Optional[List[DaqChannel]] = Field(
-        None, title="DAQ channels"
+    # required fields
+    data_interface: DataInterface = Field(..., title="Type of connection to PC")
+    camera_manufacturer: CameraManufacturer = Field(..., title="Camera manufacturer")
```
saskiad, line 178: manufacturer already is part of the device class

src/aind_data_schema/device.py
```diff
@@ -47,36 +141,147 @@ class DaqChannelType(Enum):
...
+
+    # required fields
+    channel_name: str = Field(..., title="DAQ channel name")
+    device_name: str = Field(..., title="Name of connected device")
+    channel_type: DaqChannelType = Field(..., title="DAQ channel type")
+
+    # optional fields
+    port: Optional[int] = Field(None, title="DAQ port")
+    channel_index: Optional[int] = Field(None, title="DAQ channel index")
+    sample_rate: Optional[float] = Field(-1.0, title="DAQ channel sample rate (Hz); -1 = event-based sampling", units="Hz")
```
saskiad, line 246: we need a different way to indicate event based samplling.

jsiegle: How about adding a boolean field for `event_based_sampling`?

src/aind_data_schema/device.py
jsiegle: Yep, that's a mistake

src/aind_data_schema/ephys/ephys_session.py
jsiegle: This info is in the rig JSON (there is a list of Lasers associated with each LaserModule)

src/aind_data_schema/device.py
saskiad: oh I see ... this way it points to a specific list... hmmm

src/aind_data_schema/ephys/ephys_session.py
jsiegle: The idea is that the rig JSON contains information about all the physical connections on the rig, and the session JSON indicates which devices are being recorded in each Stream. So the DAQ/Camera/EphysProbe class in the session JSON just provide an ID that can be used to look up the relevant hardware info in the rig JSON.

src/aind_data_schema/ephys/ephys_session.py
saskiad: I think the rig json has maximum power, which is different from the actual power used. Wavelength is in both, but is optional in the rig for situations where it is tunable.

src/aind_data_schema/ephys/ephys_session.py
jsiegle: That's fine, I can make this change

src/aind_data_schema/ephys/ephys_session.py
jsiegle: I think that's fine for now. The reason to make it a class would be if the session JSON creator might eventually ingest the rig JSON and provide a list of device-type-specific names for the user to choose from.

src/aind_data_schema/device.py
jsiegle: I think so. I kept it as `DeviceBase` was to encourage people to make more specific classes for different device types.

src/aind_data_schema/device.py
jsiegle: I copied this from `ophy_rig.py` – we should consult with the ophys team about this, but it may be better to make this a list of `ints` (or perhaps have separate fields for upper + lower bounds)

src/aind_data_schema/device.py
jsiegle: Yeah, that should simplify things

src/aind_data_schema/device.py
jsiegle: When I was creating the rig JSON using the web interface, it was annoying to search through a long list of manufacturer names for every device. Not that terrible in the scheme of things, but having a shorter list would improve the user experience. If we do split it up, I agree with David that we should use existing `enum` values if possible.

src/aind_data_schema/device.py
saskiad: yeah that works

src/aind_data_schema/ephys/ephys_session.py
jsiegle: As David pointed out above, the ones in the session JSON don't necessarily have to be separate classes

src/aind_data_schema/device.py
jsiegle: I made this a separate field so we could use a camera-specific device `enum`. But it sounds like it's possible to just use a subset of the existing manufacturer `enum` instead.

src/aind_data_schema/device.py
saskiad: I'm mostly worried about maintaining different lists and adding Thorlabs to Filters but forgetting to add it to lenses, etc. But a long list is also a pain...

src/aind_data_schema/device.py
```diff
@@ -17,27 +17,121 @@ class Manufacturer(Enum):
...
 
+class LensManufacturer(Enum):
+    """Camera manufacturer"""
+
+    EDMUND_OPTICS = "Edmund Optics"
+    BASLER = "Basler"
+    THORLABS = "Thorlabs"
+
+
+class FilterManufacturer(Enum):
```
saskiad, line 46: if we do keep these separate, each list needs an Other option

src/aind_data_schema/device.py
```diff
@@ -47,36 +141,147 @@ class DaqChannelType(Enum):
...
+    # required fields
+    data_interface: DataInterface = Field(..., title="Type of connection to PC")
+    camera_manufacturer: CameraManufacturer = Field(..., title="Camera manufacturer")
+    max_frame_rate: float = Field(..., title="Maximum frame rate (Hz)", units="Hz")
+    pixel_width: int = Field(..., title="Width of the sensor in pixels", units="Pixels")
+    pixel_height: int = Field(..., title="Height of the sensor in pixels", units="Pixels")
+    chroma: CameraChroma = Field(..., title="Color or Monochrome")
+
+    # optional fields
+    sensor_format: Optional[str] = Field(None, title='Size of the sensor (e.g. 1/2.9")')
```
saskiad, line 186: what are the units for size?

src/aind_data_schema/device.py
```diff
@@ -47,36 +141,147 @@ class DaqChannelType(Enum):
...
+    device_name: str = Field(..., title="Name of connected device")
+    channel_type: DaqChannelType = Field(..., title="DAQ channel type")
+
+    # optional fields
+    port: Optional[int] = Field(None, title="DAQ port")
+    channel_index: Optional[int] = Field(None, title="DAQ channel index")
+    sample_rate: Optional[float] = Field(-1.0, title="DAQ channel sample rate (Hz); -1 = event-based sampling", units="Hz")
+    
 
 class DAQ(DeviceBase):
```
saskiad, line 249: Can we call this "DAQDevice" or something? Calling it "DAQ" implies to me that it is the DAQ itself, rather than a device that is connected via the DAQ.

src/aind_data_schema/device.py
jsiegle: It looks like there is an easy way to define a field as a subset of a larger `enum`: https://github.com/pydantic/pydantic/discussions/4232

I will make this change and see if it works

src/aind_data_schema/ephys/ephys_session.py
jsiegle: Actually, I think we should keep these as separate classes, because they may need to hold session-specific metadata (e.g. camera exposure time)

src/aind_data_schema/device.py
jsiegle: There is not a standard unit here...e.g. 1/2.9" and 35 mm sensors are both common

jsiegle:
@saskiad and @dyf, the most recent commit addresses many of the issues you raised, please take a look at the latest code

src/aind_data_schema/device.py
```diff
@@ -47,36 +141,147 @@ class DaqChannelType(Enum):
...
+    position_z: Optional[float] = Field(None, title="Position Z")
+    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
 
 
 class DeviceBase(AindModel):
     """Description of a general device"""
 
-    manufacturer: Manufacturer = Field(..., title="Manufacturer")
-    serial_number: str = Field(..., title="Serial number")
+    device_name: Optional[str] = Field(None, title="Device name")
```
saskiad: device_name is going to be a really tricky part to implement cleanly. I'm anticipating situations where we have "camera_1" and "camerab" and really messy/confusing names floating around. And this organization of having the list of DAQ channels (rather than specify the DAQ channel(s) within the device) really hinges on that. I think this is why I prefer putting the DAQ channels in the device instead - and we can definitely flesh that out so it's more detailed.

jsiegle: Can we chat about this after the workshop planning meeting tomorrow? I think sketching out the dependencies on a whiteboard would be helpful

src/aind_data_schema/device.py
saskiad: we need to add a units filed for this then - we can do it after the PR

src/aind_data_schema/device.py
saskiad: Can you add this boolean field?

jsiegle: Done!

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -106,9 +111,9 @@ class EphysSession(AindCoreModel):
         description="First and last name of the experimenter.",
         title="Experimenter full name",
     )
-    session_start_time: datetime = Field(..., title="Session start time")
-    session_end_time: datetime = Field(..., title="Session end time")
-    subject_id: int = Field(..., title="Subject ID")
+    stream_start_time: datetime = Field(..., title="Session start time")
+    stream_end_time: datetime = Field(..., title="Session end time")
+    subject_id: Optional[int] = Field(1, title="Subject ID; 1 = Test subject")
```
saskiad, line 116: remove the Test subject. I'm really against having secret codes hidden in there.

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -106,9 +111,9 @@ class EphysSession(AindCoreModel):
         description="First and last name of the experimenter.",
         title="Experimenter full name",
     )
-    session_start_time: datetime = Field(..., title="Session start time")
-    session_end_time: datetime = Field(..., title="Session end time")
-    subject_id: int = Field(..., title="Subject ID")
+    stream_start_time: datetime = Field(..., title="Session start time")
```
saskiad, line 114: this appears backwards - this should be session start time, right?

saskiad (changes requested):
There are a couple of small things to fix - all super minor

jsiegle:
Fixed!

src/aind_data_schema/ephys/ephys_session.py
saskiad: I'm confused how the subject ID was removed...

saskiad: nevermind - github was hiding that section from me.

saskiad (approved):

PR #140, opened 2023-01-04
feat: updated black line length

### LINKED ISSUE #128: increase black line length
I think 100 or 120 would be fine.

mekhlakapoor:
closes #128


jtyoung84 (changes requested):
Can you also have the config in the .flake8 file match the 120 line length config?

jtyoung84 (approved):

PR #143, opened 2023-01-08
Ephys example scripts

jsiegle:
Add examples scripts and generated JSON files for `ephys_session` and `ephys_rig`

### DIFF
--- a/examples/ephys_rig.py
+++ b/examples/ephys_rig.py
@@ -1,0 +1,137 @@
+from aind_data_schema.ephys.ephys_rig import *
+from aind_data_schema.device import *
+
+# Describes a rig with running wheel, 2 behavior cameras, one Harp Behavior board,
+# one dual-color laser module, one stick microscope, and 2 Neuropixels probes
+
+behavior_computer = "W10DT72941"
+ephys_computer = "W10DT72942"
+
+running_wheel = Disc(name="Running Wheel",radius=15)
+
+digital_out0 = DAQChannel(channel_name="DO0",
+                        device_name="Face Camera",
+                        channel_type="Digital Output")
+
+digital_out1 = DAQChannel(channel_name="DO1",
+                        device_name="Body Camera",
+                        channel_type="Digital Output")
+
+analog_input = DAQChannel(channel_name="AI0",
+                          device_name="Running Wheel",
+                          channel_type="Analog Input")
+
+harp = HarpDevice(name="Harp Behavior",
+                  harp_device_type="Behavior",
+                  harp_device_version="2.1",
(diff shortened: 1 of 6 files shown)


examples/ephys_session.py
```diff
@@ -0,0 +1,42 @@
...
+
+session = EphysSession(experimenter_full_name="Josh Siegle",
+                      subject_id=100001,
+                      session_start_time = datetime.now(),
+                      session_end_time=datetime.now(),
+                      session_type="Test",
+                      rig_id="323_EPHYS1",
+                      data_streams=[stream])
+
+with open("ephys_session.json", "w") as outfile:
```
dyf, line 41: Can you replace this with `session.write_standard_file()`?

examples/ephys_session.py
```diff
@@ -0,0 +1,42 @@
+from datetime import datetime
+from aind_data_schema.ephys.ephys_session import *
```
dyf, line 2: Can you `import ... as es` instead to be consistent with the other examples?

examples/ephys_rig.py
```diff
@@ -0,0 +1,137 @@
...
+
+rig = EphysRig(rig_id="323_EPHYS1",
+              ephys_modules=[ephys_moduleA, ephys_moduleB],
+              cameras=[camassm1, camassm2],
+              laser_modules=[laser_module],
+              daqs=[basestation, harp],
+              stick_microscopes=[microscope],
+              mouse_platform=running_wheel)
+
+with open("ephys_rig.json", "w") as outfile:
```
dyf, line 136: Can you replace this with `rig.write_standard_file()`?

dyf (changes requested):

examples/ephys_session.py
dyf: (the `import *` is why the CI is breaking right now)

dyf:
@jsiegle almost there - now you need to regenerate your .json files.

jsiegle:
Hmm...I regenerated but it's still failing. And the expected vs. actual JSON strings look the same. Can you tell what the difference is?

dyf:
@jsiegle It's a couple things.  
1) the datetimes don't match because you are using `datetime.now()`.  Can you fix the time like this? https://github.com/AllenNeuralDynamics/aind-data-schema/blob/83de3c9b189e7a5fd425af385bf79b3287ddb74b/examples/subject.py#L6
2) subject ID should be a string, not an integer.  I'm not sure why that's rendering out as an integer locally and a string in the CI.

jsiegle:
Thanks, it's passing now!

dyf (approved):

PR #149, opened 2023-01-12
Making craniotomy coordinates optional to allow whole hemisphere craniotomies

saskiad:


src/aind_data_schema/__init__.py
```diff
@@ -1,7 +1,7 @@
 """ imports for AindModel subclasses
 """
 
-__version__ = "0.8.0"
+__version__ = "0.8.`"
```
dyf, line 4: Don't need to do this anymore - you can put this back to `0.8.0`.

dyf (changes requested):

dyf (approved):

PR #151, opened 2023-01-12
Make stream-specific device lists optional in EphysSession class

jsiegle:
Previously, each stream added to an `EphysSession` needed to have at least one camera, laser, probe, etc. However, there are sessions in which some of these devices may not be recorded. These changes make these fields optional.

### DIFF
--- a/src/aind_data_schema/ephys/ephys_session.py
+++ b/src/aind_data_schema/ephys/ephys_session.py
@@ -40,7 +40,7 @@ class ManipulatorModule(AindModel):
     """A module connected to a 3-axis manipulator"""
 
     primary_targeted_structure: str = Field(..., title="Targeted structure")
-    targeted_ccf_coordinates: Optional[CcfCoords] = Field(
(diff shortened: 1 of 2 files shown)


src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -40,7 +40,7 @@ class ManipulatorModule(AindModel):
     """A module connected to a 3-axis manipulator"""
 
     primary_targeted_structure: str = Field(..., title="Targeted structure")
-    targeted_ccf_coordinates: Optional[CcfCoords] = Field(
+    targeted_ccf_coordinates: Optional[List[CcfCoords]] = Field(
```
dyf, line 43: Confirming that a single `ManipulatorModule` can target multiple CCF locations?

jsiegle: Yes, I changed this because there may be cases where there are multiple coordinates we're trying to hit with a single probe

dyf (approved):

PR #157, opened 2023-01-21
Stitching branch

sharmishtaa:
Would like to add a stitching schema so that output from stitching algorithms can conform. The modules output tiles and associated transformations with each tile. The following are the main changes:

1. I moved the code for the Tile class and transformations to a new file tile.py so that code can be shared between the acquisition schema and stitching schema. 
2. Setup a generic tile transform class so that coordinate transforms can be a list of those.
3. scale, translation, rotation and affine are some transformation classes added
4. added stitching class to processing.py (Saskia thought that would be the place for it).

Would appreciate your eyes on this!


jtyoung84 (commented):
We've been migrating towards using absolute imports instead of relative imports. Also, the class ordering in Python, unlike Java, is important. Some of the classes in modules are called before being defined.

src/aind_data_schema/imaging/tile.py
```diff
@@ -0,0 +1,59 @@
...
+class Affine3dTransform(Tile3dTransform):
+    """Values to be vector-added to a 3D position. Often needed to specify a Tile's origin."""
+
+    type: str = Field("affine", title="transformation type")
+    affinetransform: conlist(float, min_items=12, max_items=12) = Field(..., title="Affine transform matrix values (top 3x4 matrix)")
+
+class Tile(AindCoreModel):
+    """Description of an image tile"""
+
+    coordinate_transformations: List[Union[Tile3dTransform]] = Field(
```
dyf, line 49: You have to enumerate the actual subclasses here in the `Union`, not just list the base class. The data entry GUI will not figure out what subclasses you want to support this way.

src/aind_data_schema/imaging/tile.py
```diff
@@ -0,0 +1,59 @@
...
+    """Description of a channel"""
+
+    channel_name: str = Field(..., title="Channel")
+    laser_wavelength: int = Field(..., title="Wavelength", ge=300, le=1000)
+    laser_wavelength_unit: str = Field("nanometer", title="Laser wavelength unit")
+    laser_power: float = Field(..., title="Laser power", le=2000)
+    laser_power_unit: float = Field("milliwatt", title="Laser power unit")
+    filter_wheel_index: int = Field(..., title="Filter wheel index")
+
+class Tile3dTransform(AindCoreModel):
```
dyf, line 17: This is a model we can re-use outside the context of `Tile`s.  Let's just call this `CoordinateTransform`.

src/aind_data_schema/imaging/tile.py
```diff
@@ -0,0 +1,59 @@
...
+    translation: conlist(float, min_items=3, max_items=3) = Field(..., title="3D translation parameters")
+
+class Rotation3dTransform(Tile3dTransform):
+    """Values to be vector-added to a 3D position. Often needed to specify a Tile's origin."""
+
+    type: str = Field("rotation", title="transformation type")
+    rotation: conlist(float, min_items=9, max_items=9) = Field(..., title="3D rotation matrix values (3x3) ")
+
+class Affine3dTransform(Tile3dTransform):
+    """Values to be vector-added to a 3D position. Often needed to specify a Tile's origin."""
```
dyf, line 41: nitpick - "Values from a 3x4 affine transform matrix to be matrix-multiplied with a 3D position", or something?

dyf (changes requested):
Looks good, just a couple changes.

src/aind_data_schema/imaging/tile.py
sharmishtaa: yeah, this is something  I was going back and forth on. This would be done in a homogenous coordinate space where a row of [0,0,0,1] is appended to this matrix and the point to be multiplied is [x,y,z,1]^T . So saving the extra 0s and 1s is redundant. We would have to have code that processes this this way.

src/aind_data_schema/imaging/tile.py
sharmishtaa: So are you suggesting I take off the Union ?I feel like I we would want to be flexible about what the order of classes is going to be?

dyf: `Union[Scale3dTransform,Translation3dTransform,Affine3dTransform,etc]`

dyf (approved):

PR #162, opened 2023-01-24
Feat opendatabucket

dyf:


doc_template/source/opendata.rst
```diff
@@ -0,0 +1,75 @@
...
+S3 Access to AIND Open Data
+###########################
+
+The Allen Institute for Neural Dynamics (AIND) is committed to FAIR, Open, 
+and Reproducible science. We therefore share all of the raw and derived data 
+we collect publicly with rich metadata. We share data as near to the 
+time of collection as possible, including preliminary data collected 
+during methods development.
+
+In addition to sharing curated datasets with data archives like `DANDI <https://dandiarchive.org/>`_
```
saskiad, line 11: I'd point out that these are NIH repositories and maybe that they are modality specific?

doc_template/source/opendata.rst
```diff
@@ -0,0 +1,75 @@
...
+A raw extracellular ephys asset would look like this:
+
+.. code-block:: text
+
+    ecephys_123456_2022-12-12_05-06-07/
+        ecephys/<ephys data>
+        behavior/<video data>
+        data_description.json
+        subject.json
+        procedures.json
```
saskiad, line 52: would raw data have processing.json? Maybe for compression?

saskiad (commented):
I think we want to add something about data exists along different stages of processing and QC evaluation. Not all data might be "good enough" (probably have to phrase that more concretely) to use.

doc_template/source/opendata.rst
dyf: yeah, Jon generated one for ephys compression

saskiad (approved):

PR #167, opened 2023-01-26
revamp procedure type

### LINKED ISSUE #165: Update `Procedure.type`
Right now `Procedure.type` is not being set in any subclasses.  We should either remove it or populate it everywhere.

saskiad:
resolves #165


dyf (changes requested):
bump procedures version, otherwise looks good

dyf (approved):

PR #170, opened 2023-02-01
specimen

saskiad:
work in progress


src/aind_data_schema/procedures.py
```diff
@@ -28,6 +28,40 @@ class CurrentUnit(Enum):
...
+
+class Reagent(AindModel):
+    """Description of reagents used in procedure"""
+    name: str = Field(..., title="Name")
+    RRID: Optional[str] = Field(None, "Research Resource ID")
+    lot_number: str = Field(..., title="Lot number")
+    expiration_date: Optional[date] = Field(None, "Lot expiration date")
+
+
+class SpecimenProcedure(AindModel):
```
dyf, line 49: I think we can have a base `Procedure` class that has the shared fields between `SubjectProcedure` and `SpecimenProcedure`

src/aind_data_schema/procedures.py
```diff
@@ -28,6 +28,40 @@ class CurrentUnit(Enum):
     UA = "microamps"
 
 
+class TissuePrepName(Enum):
```
dyf, line 31: `SpecimenProcedureName`?

src/aind_data_schema/procedures.py
```diff
@@ -336,6 +353,16 @@ class WaterRestriction(AindModel):
...
+    """Description of a perfusion procedure that creates a specimen"""
+
+    specimen_id: List[str] = Field(
+        ..., 
+        title="Specimen ID",
+        description="IDs of specimens resulting from this procedure.",
+        unique_items=True,
+        )
+
 class Procedures(AindCoreModel):
```
dyf, line 366: How does this affect `Procedures`? Would we have separate lists for `SubjectProcedure`s and `SpecimenProcedure`s? Is there a separate `Specimens` model?

src/aind_data_schema/procedures.py
```diff
@@ -28,6 +28,40 @@ class CurrentUnit(Enum):
...
+    end_date: date = Field(..., title="End date")
+    experimenter_full_name: str = Field(
+        ...,
+        description="First and last name of the experimenter.",
+        title="Experimenter full name",
+    )
+    protocol_id: str = Field(..., title="Protocol ID", description="DOI for protocols.io")
+    reagents: Optional[List[Reagent]] = Field(None, "Reagents")
+    notes: Optional[str] = Field(None, title="Notes")
+
```
dyf, line 63: do we want a general `output_specimen` here? Or just add it to the procedures that yield new specimens?

src/aind_data_schema/procedures.py
saskiad: I have a separate issue where I want Procedures to be a Union. But yes, we might need to have separate lists of Subject and Specimen Procedures. This was the piece I haven't yet tackled.

saskiad: I'm not sure what a Specimens model needs to be. I had assumed we needed one, but I'm not sure what information it needs so I haven't made it yet.

src/aind_data_schema/procedures.py
saskiad: hmm... we can. I feel like there's not really enough in common that it makes things extra confusing.

src/aind_data_schema/procedures.py
saskiad: only for specific ones.

saskiad:
@dyf take a look at these updates from yesterday. I might be ready to clean up examples to pull this

dyf (approved):

Issue #180, opened 2023-02-09
move mouseplatform to device


saskiad:
used by both ophys/ephys rigs

saskiad:
also monitor

Issue #179, opened 2023-02-09
Add mesoscope to ophys session


saskiad:

saskiad:
closed with PR# 181

PR #181, opened 2023-02-10
adding mesoscope to ophys

saskiad:
changes to ophys rig, ophys session, device, ephys rig

### DIFF
--- a/src/aind_data_schema/ophys/ophys_session.py
+++ b/src/aind_data_schema/ophys/ophys_session.py
@@ -78,7 +78,7 @@ class FiberPhotometrySession(OphysSession):
     coupling_array: List[Coupling] = Field(..., title="Coupling array", unique_items=True)
 
 
-class FieldOfView(AindCoreModel):
+class FieldOfView(AindModel):
     """Description of an imaging field of view"""
 
     index: int = Field(..., title="Index")
@@ -87,20 +87,20 @@ class FieldOfView(AindCoreModel):
     targeted_structure: str = Field(..., title="Targeted structure")
     fov_coordinate_ml: float = Field(..., title="FOV coodinate ML")
     fov_coordinate_ap: float = Field(..., title="FOV coordinate AP")
+    fov_coordinate_unit: SizeUnit = Field(SizeUnit.um, title="FOV coordinate unit")
     fov_reference: str = Field(..., title="FOV reference", description="Reference for ML/AP coordinates")
     fov_width: int = Field(..., title="FOV width (pixels)")
     fov_height: int = Field(..., title="FOV height (pixels)")
     fov_size_unit: SizeUnit = Field(SizeUnit.PX, title="FOV size unit")
-    fov_scale_factor: float = Field(..., title="FOV scale factor (pixels/um)")
-    fov_scale_factor_unit: str = Field("um/pixels", title="FOV scale factor unit")
+    fov_scale_factor: float = Field(..., title="FOV scale factor (um/pixel)")
+    fov_scale_factor_unit: str = Field("um/pixel", title="FOV scale factor unit")
     frame_rate: float = Field(..., title="Frame rate (Hz)")
     frame_rate_unit: FrequencyUnit = Field(FrequencyUnit.Hz, title="Frame rate unit")
 
 
-class TwoPhotonSession(OphysSession):
+class TwoPhotonOphysSession(OphysSession):
     """Description of a two photon session"""
 
-    number_of_planes: int = Field(..., title="Number of planes")
     fovs: List[FieldOfView] = Field(..., title="Fields of view", unique_items=True)
 
 
--- a/src/aind_data_schema/ophys/ophys_session.py
+++ b/src/aind_data_schema/ophys/ophys_session.py
@@ -9,17 +9,17 @@ from typing import List, Optional
 from pydantic import Field
 
 from ..base import AindCoreModel, AindModel
-from ..device import PowerUnit, SizeUnit
+from ..device import PowerUnit, SizeUnit, FrequencyUnit
 
 
 class FiberName(Enum):
     """Fiber name"""
 
-    LASER_A = "Laser A"
-    LASER_B = "Laser B"
-    LASER_C = "Laser C"
-    LASER_D = "Laser D"
-    LASER_E = "Laser E"
+    FIBER_A = "Fiber A"
+    FIBER_B = "Fiber B"
+    FIBER_C = "Fiber C"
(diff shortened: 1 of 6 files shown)


saskiad:
issues #179 and #180

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -71,12 +71,44 @@ class Patch(AindModel):
...
+
+class FieldOfView(AindCoreModel):
+    """Description of an imaging field of view"""
+
+    index: int = Field(..., title="Index")
+    imaging_depth: int = Field(..., title="Imaging depth (um)")
+    imaging_depth_unit: SizeUnit = Field(SizeUnit.UM, title="Imaging depth unit")
+    targeted_structure: str = Field(..., title="Targeted structure")
+    fov_coordinate_ml: float = Field(..., title="FOV coodinate ML")
+    fov_coordinate_ap: float = Field(..., title="FOV coordinate AP")
```
dyf, line 89: need `fov_coordinate_unit`?

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -71,12 +71,44 @@ class Patch(AindModel):
...
+    fov_reference: str = Field(..., title="FOV reference", description="Reference for ML/AP coordinates")
+    fov_width: int = Field(..., title="FOV width (pixels)")
+    fov_height: int = Field(..., title="FOV height (pixels)")
+    fov_size_unit: SizeUnit = Field(SizeUnit.PX, title="FOV size unit")
+    fov_scale_factor: float = Field(..., title="FOV scale factor (pixels/um)")
```
dyf, line 94: the title includes the units but doesn't match the value in the next line

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -71,12 +71,44 @@ class Patch(AindModel):
...
+    fov_scale_factor_unit: str = Field("um/pixels", title="FOV scale factor unit")
+    frame_rate: float = Field(..., title="Frame rate (Hz)")
+    frame_rate_unit: FrequencyUnit = Field(FrequencyUnit.Hz, title="Frame rate unit")
+
+
+class TwoPhotonSession(OphysSession):
```
dyf, line 100: This makes me squeamish because tissuecyte was 2-p.  I can't think of a better name.

dyf: Maybe `TwoPhotonOphysSession`

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -71,12 +71,44 @@ class Patch(AindModel):
...
+    """Description of a two photon session"""
+
+    number_of_planes: int = Field(..., title="Number of planes")
```
dyf, line 103: why do we need `number_of_planes`?

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -71,12 +71,44 @@ class Patch(AindModel):
...
     output_power_unit: PowerUnit = Field(PowerUnit.UW, title="Output power unit")
 
+class FiberPhotometrySession(OphysSession):
+    """Description of a fiber photometry session"""
+
+    patch_cords: List[Patch] = Field(..., title="Patch cords", unique_items=True)
+    coupling_array: List[Coupling] = Field(..., title="Coupling array", unique_items=True)
+
+
+class FieldOfView(AindCoreModel):
```
dyf, line 81: this can be `FieldOfView(AindModel)` - it doesn't have a version,etc.

src/aind_data_schema/ophys/ophys_session.py
dyf, line 103: also, should it be `number_of_fovs`?

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -71,12 +71,44 @@ class Patch(AindModel):
...
+    """Description of an imaging field of view"""
+
+    index: int = Field(..., title="Index")
+    imaging_depth: int = Field(..., title="Imaging depth (um)")
+    imaging_depth_unit: SizeUnit = Field(SizeUnit.UM, title="Imaging depth unit")
+    targeted_structure: str = Field(..., title="Targeted structure")
+    fov_coordinate_ml: float = Field(..., title="FOV coodinate ML")
+    fov_coordinate_ap: float = Field(..., title="FOV coordinate AP")
+    fov_reference: str = Field(..., title="FOV reference", description="Reference for ML/AP coordinates")
```
dyf, line 90: What is an example `fov_reference`?

src/aind_data_schema/ophys/ophys_session.py
saskiad: yes. One open quesiton I have is what the reference for coordinate are - I think they use a reticle in the center on the cranial window. Which makes these coordinates not super useful... Still worth tracking, but less meaningful

src/aind_data_schema/ophys/ophys_session.py
saskiad: but what does ophys mean?

src/aind_data_schema/ophys/ophys_session.py
saskiad: lol... inverted

src/aind_data_schema/ophys/ophys_session.py
saskiad: yeah, I'm not sure we do.

src/aind_data_schema/ophys/ophys_session.py
saskiad: the reticle they use to calibrate the rigs

src/aind_data_schema/ophys/ophys_session.py
saskiad: but sure

src/aind_data_schema/ophys/ophys_session.py
saskiad: one thing I'm not sure about is whether frame rate should be per FOV or should be a total thing, in which case being specific about number of planes can be a useful parameter to have explicitly defined. eg 30 Hz across 6 planes

saskiad: oh, you want consistency?

dyf: yeah that's real. just wondering if maybe that gets computed as part of a view.

dyf (approved):

PR #183, opened 2023-02-10
Updates to photometry ophys

saskiad:
based on working with Smrithi
adding example


dyf (approved):
generated the example, also ran the linters

PR #185, opened 2023-02-14
fixing units

### LINKED ISSUE #184: Laser power unit type
Is this meant to be a string? Somehow the default string passes validation, but a different string does not.

https://github.com/AllenNeuralDynamics/aind-data-schema/blob/9439ac3080f812127201948aaac3c944f63859f7/src/aind_data_schema/imaging/tile.py#L18

dyf:
resolves #184


miketaormina (approved):
as long as the ExaSPIM doesn't get into MegaWatts!

saskiad (approved):

PR #186, opened 2023-02-14
SmartSPIM Acquisition Example

miketaormina:
Script parses metadata files to generate acquisition JSON post hoc

A few questions about SmartSPIM acquisition schemas:
1. SmartSPIM saves data as individual files for each z (scanning axis) position, but my understanding is that a `AcquisitionTile` should refer to an image stack. So the `filename` field doesn't quite work here, since the directory path carries no information about file names, extensions, etc. So I'm not sure what to put here.
2. In this example, I used a local path to `aind.vast01.corp.alleninstitute.org`, so it won't run without editing that line. Again, I'm not sure how we want to handle that for the purpose of creating an example (I needed to make one to share with other teams within AIBS, but it doesn't need to be part of this repo.

### DIFF
--- a/examples/smartspim_acquisition.json
+++ b/examples/smartspim_acquisition.json
@@ -1,0 +1,2727 @@
+{
+   "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/imaging/acquisition.py",
+   "schema_version": "0.4.1",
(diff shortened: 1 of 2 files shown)


dyf:
@miketaormina this script probably makes more sense to add to the `aind-data-transfer` repo. I pinged @camilolaiton there.  Let's leave this open as a draft PR for now.

dyf:
closing this, as work is done elsewhere.

PR #189, opened 2023-02-15
SmartSPIM specimen procedures

miketaormina:
I wanted to try generating one of these but needed to add a few things:
- `src.procedures.py` -- new `TissuePrepName` entries.
-- "Active delipidation" for using the SmartBatch+ clearing box
-- "Soak" as a general specimen in a reagent for an amount of time step. I'm not sure what you want to call this, but there are several instances where we leave a specimen in a reagent for 1 or more days, so I think it makes sense to have it.
-- "Embedding" in this case for agarose embedding for imaging, but could be used for others. Not sure if you want to specifically track this procedure or not, but it will have an SOP
- Fixed several instances of the `Field` class needing a `title` keyword

I'm not sure what the current thinking is for specimen vs. subject functionality here. In my example, the subject_id and the specimen_id are identical, but you may disallow this for clarity.

### DIFF
--- a/examples/aibs_smartspim_procedures.json
+++ b/examples/aibs_smartspim_procedures.json
@@ -1,18 +1,12 @@
 {
    "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/procedures.py",
-   "schema_version": "0.5.4",
+   "schema_version": "0.6.0",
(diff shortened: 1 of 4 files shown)


dyf:
@miketaormina @saskiad I'm going to try to keep this branch up-to-date as Saskia continues to work on the specimen model.

PR #190, opened 2023-02-15
adding background subtraction and cell segmentation processes

camilolaiton:
Hello, I've noticed that we don't have declared data processes for background subtraction nor cell segmentation. Please, let me know if these names are appropriate or not.


dyf (changes requested):
can you put these in alphabetical order?

dyf (approved):

Issue #171, opened 2023-02-01
surgical coordinates


saskiad:
In procedures, add the lambda-bregma distance, and add something that makes it clear that coordinates are relative to Bregma.

saskiad:
I thought this was going to be easy, but ... in the SWR the coordinates are "relative to Bregma unless noted" and several of the examples I see have "rostral to lambda" in the notes.
So perhaps we need a "coordinate reference" with options of Bregma or Lambda? What do you think @mekhlakapoor @dyf 
Ugh, can't we just use a single reference point people?

mekhlakapoor:
I agree that we should prob have a coordinate reference in the schema, and in my api i can have it default to bregma and write a parsing method to check notes for "lambda"

PR #194, opened 2023-02-17
procedure coordinate reference

saskiad:
addresses issue #171


src/aind_data_schema/procedures.py
```diff
@@ -131,6 +131,13 @@ class CraniotomyType(Enum):
     OTHER = "Other"
 
 
+class CoordinateReferenceType(Enum):
```
dyf, line 134: I think this is less a "type" than a location.

Maybe: `CranialReferenceLocation`?

dyf (changes requested):

dyf (approved):

Issue #172, opened 2023-02-04
Cage ID


saskiad:
Add Cage ID to subject schema to indicate social housing

dyf:
Call it `housing_group`? 

https://ontobee.org/ontology/NCIT?iri=http://purl.obolibrary.org/obo/NCIT_C90394

saskiad:
not sure... I'd prefer to be precise. But we need to see what this looks like in Labtacks. Can the same cage ID be used for different mice at different times. Eg. is the cage number a rack position?

saskiad:
one approach is to have a housing class that contains an identifier (cage number), list of devices (eg. wheel/enrichments), and list of co-housed subjects

PR #203, opened 2023-03-06
Adding housing to subject

saskiad:
addresses issue #172


saskiad:
I'd like to make the enrichments a device, but that might be something for when we have a way of actually tracking enrichments

dyf (changes requested):

dyf (approved):

Issue #201, opened 2023-03-06
Incorporate rorid for institution


dyf:
We have our RORID (https://ror.org/04szwah67)
There is also one for AIBS. Can we start having these added to data description files when data is uploaded?

AIND: [04szwah67](https://ror.org/04szwah67)
AIBS: [00dcv1019](https://ror.org/00dcv1019)

saskiad:
can this be automated during data ingestion?

saskiad:
ah, I thought we already had added the field without having the values figured out! Fixing that now.

Issue #175, opened 2023-02-06
add summary to data descriptor?


saskiad:
Something that consolidates key experimental parameters and/or processing results? Eg. X units passed default QC in these regions using session type Y

PR #212, opened 2023-03-10
updates to data description

saskiad:
added rorid and summary to data description. Closes Issue #201 and closes Issue #175. Optional for now but will add tools to make required later.

### DIFF
--- a/src/aind_data_schema/data_description.py
+++ b/src/aind_data_schema/data_description.py
@@ -103,7 +103,7 @@ class Funding(AindModel):
 class DataDescription(AindCoreModel):
     """Description of a logical collection of data files"""
 
-    schema_version: str = Field("0.3.0", title="Schema Version", const=True)
+    schema_version: str = Field("0.3.1", title="Schema Version", const=True)
     license: str = Field("CC-BY-4.0", title="License", const=True)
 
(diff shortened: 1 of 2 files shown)


src/aind_data_schema/data_description.py
```diff
@@ -167,6 +172,12 @@ class DataDescription(AindCoreModel):
         ...,
         regex=DataRegex.NO_UNDERSCORES.value,
         description="Unique identifier for the subject of data acquisition",
+        title="Subject ID"
+    )
+    data_summary: Optional[str] = Field(
+        None,
+        title="Data summary",
+        description="Semantic summary of experimental goal"
```
dyf, line 180: nitpick. we want to use this to also describe random interesting things too, like "lots of great units in Thalamus" or whatever, right?  If so I might change this description.

dyf (changes requested):
nitpicking description

src/aind_data_schema/data_description.py
saskiad: I want to avoid this being a haphazard free for all. Yes, I want to add things about yield and regions, but not so much random things.

dyf (approved):

PR #213, opened 2023-03-10
adding list of processing steps to be performed to the acquisition file

saskiad:


src/aind_data_schema/imaging/acquisition.py
```diff
@@ -76,11 +76,26 @@ class Immersion(AindModel):
     medium: str = Field(..., title="Immersion medium")
     refractive_index: float = Field(..., title="Index of refraction")
 
+class ProcessName(Enum):
```
dyf, line 79: I don't like duplicating these.  Instead you could use `processing_steps: List[Literal[MyEnum.A, MyEnum.B]]`

dyf (changes requested):
I think this needs to happen at the channel level. The goal here is to communicate what channels are meant for which processing steps (which channel is the CCF registration channel, which channel is for cell counting, etc).

saskiad:
we will need a validator to make sure the channel name matches what is in tile

dyf (approved):

PR #216, opened 2023-03-13
adding transcriptomics to procedures.

saskiad:


saskiad:
@dyf if you want to take a look at this - I'm still thinking about parts but some feedback on current state could be helpful.

src/aind_data_schema/procedures.py
```diff
@@ -68,6 +74,51 @@ class SpecimenProcedure(AindModel):
...
+    oligo_probes: List[OligoProbes] = Field(..., title="Oligo probes")
+    probe_concentration: float = Field(..., title="Probe concentration")
+    probe_concentration_unit: str = Field(..., title="Probe concentration unit")
+    intrument_id: str = Field(..., title="Instrument ID")
+
+
+class HybridizationChainReaction(AindModel):
+    """Description of HCR for mFISH"""
+
+    codebook_name: str = Field(..., title="Codebook name")
```
dyf, line 116: Do we need a `codebook_name`? I think `oligo_pools` is basically the codebook, if I understand this correctly.

src/aind_data_schema/procedures.py
```diff
@@ -68,6 +74,51 @@ class SpecimenProcedure(AindModel):
...
+    ALEXA_647 = "Alexa Fluor 647"
+
+
+class OligoProbes(Reagent):
+    """Description of oligo probes used for HCR"""
+
+    channel_index: int = Field(..., title="Channel index")
+    species: Species = Field(..., title="Species")
+    fluorophore: Fluorophore = Field(..., title="Fluorophore")
+    excitation_wavelength: int = Field(..., title="Excitation wavelength")
```
dyf, line 97: Should `excitation_wavelength` be tracked instead with the acquisition? I think `channel_index` is the key that will link a probe to a channel-round, right?

src/aind_data_schema/procedures.py
```diff
@@ -68,6 +74,51 @@ class SpecimenProcedure(AindModel):
...
+    ALEXA_546 = "Alexa Fluor 546"
+    ALEXA_594 = "Alexa Fluor 594"
+    ALEXA_647 = "Alexa Fluor 647"
+
+
+class OligoProbes(Reagent):
+    """Description of oligo probes used for HCR"""
+
+    channel_index: int = Field(..., title="Channel index")
+    species: Species = Field(..., title="Species")
```
dyf, line 95: Surprised to see `species` here...I guess it makes sense for a probe to have a species tracked with it, rather than having to dig around in `subject`. I guess I'm wondering if there would ever be a case we would use a probe from a different species.

src/aind_data_schema/procedures.py
saskiad: I think that it will be useful to have a name for the array of probes that get used. Eg. the same set will get used for certain experiments, and I think being able to identify experiments that all used that set can be useful? But let's ask some scientists

src/aind_data_schema/procedures.py
saskiad: this is the excitation wavelength of the fluorophore. It's going to be redundant with the name of the fluorophore, so mostly useless. I'm okay removing it.

src/aind_data_schema/procedures.py
saskiad: So this was in their database thing. I don't think it refers to the species of the subject - I think it's the species that the probe was designed for/with. I know that's a thing for antibodies - I didn't expect it to be a thing for RNA probes, but again, let's ask.

miketaormina: This is correct. It probably just migrated from the purchasing page for the probe, since species is specified there. A few things to note:
1. In HCR experiments, we don't typically design our own probes. We give the vendor a species, gene accession number, and sequence, they then design and make the probes.
2. There have been instances were people have purposefully used probes designed in one species on a sample from a different species.

src/aind_data_schema/procedures.py
miketaormina: I agree that it's okay if it is removed. It should be recoverable from the name, but be aware that it can be different from what happens in an acquisition. For example, Alexa 546 is a fluorophore with a peak excitation wavelength of 546 nm, but is excited in the microscope by a 561 nm laser.

src/aind_data_schema/procedures.py
miketaormina: I also think a name is useful for investigators to use for organization. Although it might make a version number necessary

src/aind_data_schema/procedures.py
```diff
@@ -68,6 +74,51 @@ class SpecimenProcedure(AindModel):
...
+    fluorophore: Fluorophore = Field(..., title="Fluorophore")
+    excitation_wavelength: int = Field(..., title="Excitation wavelength")
+    excitation_wavelength_unit = SizeUnit = Field(SizeUnit.NM, title="Excitation wavelength unit")
+    stain_type: StainType = Field(..., title="Stain type")
+    readout: Optional[str] = Field(None, title="Readout")
+
```
miketaormina, line 101: I think OligoProbe is missing the gene target. There are basically three things that make a HCR reaction unique:
1. A target sequence
2. An initiator/amplifier sequence
3. A fluorophore

We purchase probes as (1 & 2) pairs and we purchase readouts as (2 & 3) pairs. This way, all of the genes that you probe with Alexa 594 can be ordered with the same amplifier and you can use the same readout stock.

I think that the way the Oligo Pools page displays is hiding the fact that a Readout is the combination of an amplifier (B1-B5, etc.) and a fluorophore.

src/aind_data_schema/procedures.py
```diff
@@ -68,6 +74,51 @@ class SpecimenProcedure(AindModel):
     notes: Optional[str] = Field(None, title="Notes")
 
 
+class StainType(Enum):
+    """Stain Types for HCR probes"""
+
+    RNA = "RNA"
+
```
miketaormina, line 81: Stain type exists in the old database because we used DAPI once upon a time.

miketaormina: But something that came after that database (and therefore isn't captured properly by it) is that the Rn28s probe that people use for cell segmentation and registration isn't an HCR probe. It is a RNA FISH probe, it just isn't amplified so it has no initiator sequence or corresponding hairpins.

src/aind_data_schema/procedures.py
saskiad: @miketaormina should we include more of the probe information? gene accession number and sequence? Sequence might be too much... But I imagine that might be useful information to have here.

src/aind_data_schema/procedures.py
saskiad: right - this is what I meant above- we should include more of this. So if I understand, we want to document both probes and readouts?

src/aind_data_schema/procedures.py
miketaormina: The accession number is a good idea to track. Right now, it's just being noted in local documents. The actual probe sequence would be difficult to track here since it's not a single sequence but up to 20 for long enough targets. Molecular Instruments doesn't actually tell you what it is when you order, but we would always specifically ask for it since it seems necessary for any published result in the future.

src/aind_data_schema/procedures.py
miketaormina: Yes. I think that a probe should consist of a gene name and initiator name (more details than name can be tracked if needed), e.g., Gad2-B1. Separately, a readout is completely described by an initiator name and a fluorophore name, e.g., B1-Alexa488.

src/aind_data_schema/procedures.py
saskiad: you just convinced me that we need to include the sequence(s)

src/aind_data_schema/procedures.py
saskiad: what things other than name can be tracked for the initiator?

saskiad: do you think we need a separate class for probes and readout? So that in the HCR round we have two lists, one of probes and one of readouts, and the are linked by the initiator? Or is there a way we can describe both the probe and readout as a single concept (but still defining each of those parts)? So there'd be one list of <entities> that have probe gene, initiator, fluorophore? (does that make sense?)

src/aind_data_schema/procedures.py
saskiad: I'm adding a class of "other_probe" for this - are there things we should document?(beyond standard reagent properties)

src/aind_data_schema/procedures.py
saskiad: I think I'm going to keep it. I don't think the redundancy hurts us that much. 
Also, to note, that the actual wavelength used would show up in the acquisition.json rather than here, this here describes a property of the probe.

src/aind_data_schema/procedures.py
miketaormina: The initiators and hairpins have sequences, but I don't know if the vendor supplies those, since they are proprietary.

I suppose it would be more accurate to track the probes and readouts separately. I'm trying to think of what you gain by doing so... if you wanted to keep track of concentrations directly (right now it is indirect via the sop) or lot numbers, maybe.

Tracking these three as a single concept is what [Oligo Pools](http://aidc-mfishlightsheet/record-keeper/oligo-pools) was doing. It's an odd name for it, but it was a gene-initiator-fluorophore combination.

src/aind_data_schema/procedures.py
miketaormina: I think `OtherProbes` looks good.

src/aind_data_schema/procedures.py
saskiad: I just pushed a version with separate reagent classes for probes and readouts, that are then tracked together in "OligoPools" (though I'm open to a different name, I feel a bit lost in jargon). Let me know if this makes sense

dyf (commented):
Some more stuff to discuss

src/aind_data_schema/procedures.py
```diff
@@ -74,6 +81,86 @@ class SpecimenProcedure(AindModel):
...
+
+class OligoPools(AindModel):
+    """Description of probes and readouts used for HCR"""
+
+    channel_index: int = Field(..., title="Channel index")
+    oligo_probe: OligoProbe = Field(..., title="Oligo probe")
+    readout: Readout = Field(..., title="Readout")
+
+class OtherProbe(Reagent):
```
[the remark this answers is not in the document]
saskiad: oligo probes and readouts pair together - so I don't think it makes sense for oligo probe to be a subclass of other probe...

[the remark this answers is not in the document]
saskiad: @miketaormina I defer to you on this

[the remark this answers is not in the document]
saskiad: oh wait - I was confusing other things - this might work

src/aind_data_schema/procedures.py
```diff
@@ -412,6 +499,13 @@ class Procedures(AindCoreModel):
                 SubjectProcedure,
             ]
         ]
-    ] = Field([], title="Subject Procedures", unique_items=True)
-    specimen_procedures: Optional[List[SpecimenProcedure]] = Field([], title="Specimen Procedures", unique_items=True)
+    ] = Field(None, title="Subject Procedures", unique_items=True)
+    specimen_procedures: Optional[
+        List[
+            Union[
+                HCRSeries,
```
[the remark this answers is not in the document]
saskiad: yeah, let me switch these. I like HCR being the specimen procedure so we get the time stamps, but I'll just have to add those to HCR

src/aind_data_schema/procedures.py
```diff
@@ -74,6 +81,86 @@ class SpecimenProcedure(AindModel):
...
+    fluorophore: Fluorophore = Field(..., title="Fluorophore")
+    excitation_wavelength: int = Field(..., title="Excitation wavelength (nm)")
+    excitation_wavelength_unit = SizeUnit = Field(SizeUnit.NM, title="Excitation wavelength unit")
+
+
+class HybridizationChainReaction(SpecimenProcedure):
+    """Description of an HCR round""" 
+
+    round_index: int = Field(..., title="Round index")
+    oligo_probes: List[OligoProbe] = Field(..., title="Oligo probes")
```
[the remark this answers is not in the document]
saskiad: maybe, but not sure there's another option...

src/aind_data_schema/procedures.py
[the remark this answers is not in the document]
miketaormina: Because `OtherProbe` as currently written is a FISH probe that is designed for a specific gene, it should have a species.

There is a case to be made for allowing a type of probe that isn't an RNA hybridization probe (e.g. DAPI), which wouldn't need a species and wouldn't fit the current `OtherProbe` model, but I don't think there are any plans to use such a probe in the HCR experiments.

src/aind_data_schema/procedures.py
miketaormina: @saskiad This all looked like it made sense to me. I was on the fence to make the following suggestion, because it would kind of shake everything up, and I'm not sure if it's worth that. A lot of this is, as you say, related to confusing jargon:

Two classes, each with a child class:
1. `OligoProbe` with child `HCRProbe`. "Oligo" refers to a synthetically produced DNA or RNA molecule (an oligonucleotide), which both the Rn28s and HCR probes are. This would have attributes:
    - `target`
        - `gene_name`, `sequence`, `accession`, etc. kind of belong to a target
    - `sequences` -- list of actual probe sequence(s)
    - `initiator` -- unique to the `HCRProbe` class. This is an initiator sequence, but does not specify fluorophore. E.g., it can be `B1`
    - `readout` -- one of the two following.
2. `Readout` with child `HCRReadout` (synonym is `Hairpin`).
    - `fluorophore`
    - `initiator` -- unique to the HCR class. This is also, e.g. `B1`

In this model, your `OligoPool` isn't needed, since it is equivalent to choosing a probe/readout pair. Getting the opinion of a biochemist might make more sense than me fumbling at this. If you want a more expert opinion, you might ask Kevin Cao (AIND Jayaram's team?) or Christian Bonatto (Pete's team)

saskiad: @miketaormina the part that confuses me here is that this suggests that  Readout becomes a property of the OligoProbe (or just the HCRprobe?). But shouldn't the Readout be a separate entity? A readout can be used with different HCRprobes. So don't we want an entity that indicates what's used (eg. the OligoPool we have now - though happy to call this something different)

src/aind_data_schema/procedures.py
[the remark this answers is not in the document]
saskiad: @dyf  I think this contradicts the tread above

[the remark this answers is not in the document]
saskiad: If there aren't plans to include non RNA hybridzation probes right now, I'm somewhat inclined to leave out a class that covers such probes for now - even though I imagine it would be good to have something for them in the future...

src/aind_data_schema/procedures.py
saskiad: The other thing in the OligPool that we need is the channel index, which doesn't make sense being a property of the probe.

dyf (approved):

src/aind_data_schema/procedures.py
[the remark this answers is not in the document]
saskiad: I'm leaving the reagents there - I think there could be other reagents that need to be tracked here

PR #219, opened 2023-03-14
adds new modalities

### LINKED ISSUE #218: add new modalities
We need to add:
- `HSFP`  - hyperspectral fiber photometry
- `FIP` - frame-injected independent-fiber photometry
- `MRI` - magnetic resonance imaging

dyf:
resolves #218 

also runs the linters

### DIFF
--- a/src/aind_data_schema/data_description.py
+++ b/src/aind_data_schema/data_description.py
@@ -44,11 +44,11 @@ class DataLevel(Enum):
 class Institution(Enum):
     """Institution name"""
 
-    AIND = "AIND"
     AIBS = "AIBS"
+    AIND = "AIND"
+    CU = "Columbia University"
     HUST = "HUST"
     NYU = "NYU"
-    CU = "Columbia University"
 
 
 class Group(Enum):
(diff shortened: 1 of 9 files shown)


src/aind_data_schema/data_description.py
```diff
@@ -44,31 +44,34 @@ class DataLevel(Enum):
...
+    OPHYS = "ophys"
 
 
 class Modality(Enum):
     """Data collection modality name"""
 
     ECEPHYS = "ecephys"
     EXASPIM = "ExASPIM"
-    SMARTSPIM = "SmartSPIM"
+    FIP = "FIP"
```
saskiad, line 68: Can you define was FIP is?

src/aind_data_schema/data_description.py
```diff
@@ -44,31 +44,34 @@ class DataLevel(Enum):
...
+    FMOST = "fMOST"
+    HSFP = "HSFP"
```
saskiad, line 70: same here - it would be good to have some of these acronyms unpacked a bit more.

saskiad (changes requested):
I want to clean up some of our acronyms when they aren't obvious. Can you define these modalities?

dyf:
@saskiad can we add these now and handle adding long names in a separate PR.  This will impact our data entry form, I think, so will take some care to do properly.

saskiad (approved):

PR #225, opened 2023-03-17
adding timedelta, unit

### LINKED ISSUE #223: Duration fields should use `timedelta`
@saskiad @mekhlakapoor The duration fields in the pydantic models should probably be `timedelta` types, which is python's equivalent of Java's Duration class. The `time` class is supposed to represent a time of day, which has hard limits like hour < 24 etc. We can also avoid writing workarounds like the ones in the utils package here.

_Originally posted by @jtyoung84 in https://github.com/AllenNeuralDynamics/aind-metadata-service/pull/82#pullrequestreview-1346288332_

dyf:
resolves #223


mekhlakapoor (approved):
looks good

jtyoung84:
We'll probably need to update the metadata-entry app to handle the time-delta format.

rjsf_playground

jtyoung84 (approved):

jtyoung84:
Or we can just assume that everything is in total number of seconds?

saskiad (approved):

PR #226, opened 2023-03-18
name/abbrev, experiment type 

dyf:
This PR:

1) Creates a `BaseName` concept that is a long name + abbreviation and uses it for Modality and Institution
2) Renamed `InstrumentType` -> `ExperimentType` and moves it to `data_description.py`
3) Removes the `from_name` classmethods from `DataDescription`.  They are not very useful.
4) Removes `InstrumentType` from `Instrument`. <--- nevermind, not doing (4)
5) Makes `DataDescription.modality` a list.

### DIFF
--- a/src/aind_data_schema/data_description.py
+++ b/src/aind_data_schema/data_description.py
@@ -224,13 +224,6 @@ class DataDescription(AindCoreModel):
             creation_time=creation_time,
         )
 
-    @classmethod
-    def from_name(cls, name, **kwargs):
-        """construct a DataDescription from a name string"""
-        d = cls.parse_name(name)
-
-        return cls(**d, **kwargs)
-
 
 class DerivedDataDescription(DataDescription):
     """A logical collection of data files derived via processing"""
--- a/src/aind_data_schema/data_description.py
+++ b/src/aind_data_schema/data_description.py
@@ -7,7 +7,7 @@ from datetime import date, datetime, time
 from enum import Enum
 from typing import List, Optional
 
-from pydantic import BaseModel, Field
+from pydantic import Field
 
 from .base import AindCoreModel, AindModel, BaseName
 from .device import InstrumentType
--- a/src/aind_data_schema/data_description.py
+++ b/src/aind_data_schema/data_description.py
@@ -10,7 +10,6 @@ from typing import List, Optional
 from pydantic import Field
 
 from .base import AindCoreModel, AindModel, BaseName
-from .device import InstrumentType
 
 
 class RegexParts(Enum):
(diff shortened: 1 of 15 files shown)


src/aind_data_schema/data_description.py
```diff
@@ -44,11 +45,11 @@ class DataLevel(Enum):
...
     """Institution name"""
 
-    AIBS = "AIBS"
-    AIND = "AIND"
-    CU = "Columbia University"
-    HUST = "HUST"
-    NYU = "NYU"
+    AIBS = BaseName(name="Allen Institute for Brain Science", abbreviation="AIBS")
+    AIND = BaseName(name="Allen Institute for Neural Dynamics", abbreviation="AIND")
+    CU = BaseName(name="Columbia University", abbreviation="CU")
```
saskiad, line 50: Do we need abbreviations for every institution? eg. I think abbreviating Columbia is a bad idea.

src/aind_data_schema/data_description.py
```diff
@@ -63,15 +64,13 @@ class Group(Enum):
...
-    ECEPHYS = "ecephys"
-    EXASPIM = "ExASPIM"
-    FIP = "FIP"
-    FMOST = "fMOST"
-    HSFP = "HSFP"
-    MESOSPIM = "mesoSPIM"
-    MRI = "MRI"
-    OPHYS = "ophys"
-    SMARTSPIM = "SmartSPIM"
+    ECEPHYS = BaseName(name="extracellular electrophysiology", abbreviation="ecephys")
```
saskiad, line 67: Sentence case on names?

src/aind_data_schema/data_description.py
dyf: Because `CU` is so ambiguous? That's fine. Is "Columbia" okay? I think having an abbreviation w/o spaces would be handy.

src/aind_data_schema/data_description.py
```diff
@@ -172,13 +185,17 @@ class DataDescription(AindCoreModel):
...
-        regex=DataRegex.NO_UNDERSCORES.value,
-        description="A short name for the specific manner, characteristic, pattern of application, or the employment of"
-        " any technology or formal procedure to generate data for a study",
+        description="A short name for the specific manner, characteristic, pattern of application, or the employment"
+        "of any technology or formal procedure to generate data for a study",
         title="Modality",
     )
+    experiment_type: ExperimentType = Field(
+        ...,
+        description="A short name for the experiment motivating the collection of this data",
```
saskiad, line 196: This description sounds like how I would define session_type or project_name rather than experiment_type...

saskiad (approved):

Issue #227, opened 2023-03-22
add opto class to ophys


saskiad:
based on kenta's protocol

saskiad:
closed with PR #229

PR #229, opened 2023-03-22
Adding opto stim to ophys session schema

saskiad:


saskiad:
closes issue #227

saskiad:
My hunch is opto stim is more of a protocol than metadata, but I don't think it will work as a protocols.io protocol, so might as well add it here.

mekhlakapoor (approved):
looks good, thanks Saskia!

saskiad:
I'm going to make this something that can be used in ephys too

saskiad:
@dyf I'm going to move this into a place where we can use it for both ophys and ephys - should I make a stim class? or something else?

dyf:
> @dyf I'm going to move this into a place where we can use it for both ophys and ephys - should I make a stim class? or something else?

Yes, I think a generic stimulus class makes sense. Would presumably be a top-level thing. Is there an existing stimulus ontology we can crib from? NWB maybe has more now than it did back in the day.

dyf:
https://github.com/SciCrunch/NIF-Ontology/blob/estim/ttl/stimulation.ttl is what NWB's site points to.  They refer to a stimulus ontology working group: https://www.incf.org/sig/incf-working-group-electrophysiology-stimulation-ontology.

dyf (changes requested):
see other comments re: stimulus model, and TimeUnit

saskiad:
Let's make a new issue about how to incorporate NIF ontology into this - also for the behavior schema. I think that's a bigger question - so let's start with this and make it better.

src/aind_data_schema/stimulus.py
```diff
@@ -0,0 +1,58 @@
...
+class Stimulus(AindCoreModel):
+    """Description of stimulus used in a recording session"""
+
+    schema_version: str = Field(
+        "0.0.1",
+        description="schema version",
+        title="Schema Version",
+        const=True,
+    )
+    stimulus_name: str = Field(..., title="Stimulus name")
```
dyf, line 57: Should this rather be a base class and `OptoStim` a subclass?

dyf (changes requested):

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -84,6 +85,22 @@ class Camera(AindModel):
     name: str = Field(..., title="Camera name (must match rig JSON)")
 
 
+class Stim(AindModel):
+    """Description of stimulus used during session"""
+
+    stimulus: OptoStim = Field(..., title="Stimulus")
```
dyf, line 91: might as well turn this into a union (even if it has just one option), since there will be other stims coming

src/aind_data_schema/ephys/ephys_session.py
```diff
@@ -93,12 +110,13 @@ class Stream(AindModel):
     laser_modules: Optional[List[LaserModule]] = Field(None, title="Laser modules", unique_items=True)
     daqs: Optional[List[DAQDevice]] = Field(None, title="DAQ devices", unique_items=True)
     cameras: Optional[List[Camera]] = Field(None, title="Cameras", unique_items=True)
+    stimulus: Optional[List[Stim]] = Field(None, title="Stimulus")
```
dyf, line 113: Does this need to be per stream? I had assumed it would be per session.

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -100,11 +117,11 @@ class OphysSession(AindCoreModel):
     session_end_time: Optional[datetime] = Field(None, title="Session end time")
     subject_id: int = Field(..., title="Subject ID")
     session_type: str = Field(..., title="Session type")
-    stimulus_protocol_id: Optional[str] = Field(None, title="Stimulus protocol ID")
     iacuc_protocol: Optional[str] = Field(None, title="IACUC protocol")
     rig_id: str = Field(..., title="Rig ID")
     lasers: List[Laser] = Field(..., title="Lasers", unique_items=True)
     detectors: Optional[List[Detector]] = Field(None, title="Detectors", unique_items=True)
+    stimulus: Optional[List[Stim]] = Field(None, title="Stimulus")
```
dyf, line 124: I think NWB calls this `stimulus_presentations`

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -82,11 +84,26 @@ class Patch(AindModel):
     output_power_unit: PowerUnit = Field(PowerUnit.UW, title="Output power unit")
 
 
+class Stim(AindModel):
+    """Description of stimulus used during session"""
+
+    stimulus: OptoStim = Field(..., title="Stimulus")
```
dyf, line 90: make this a Union?

dyf (changes requested):

saskiad:
there's going to be a conflict in the procedures file - we don't want to lose the transcriptomics stuff.

saskiad:
I don't understand how the main got merged into feat_opto

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -82,11 +84,26 @@ class Patch(AindModel):
     output_power_unit: PowerUnit = Field(PowerUnit.UW, title="Output power unit")
 
 
+class Stim(AindModel):
```
dyf, line 87: Sorry, had one other thought.  Since this is duplicated in both of the sessions, how about we call this `StimulusPresentation` and move it into `../stimulus.py`?

dyf (changes requested):

src/aind_data_schema/ophys/ophys_session.py
saskiad: why be logical?!?

dyf (approved):

PR #264, opened 2023-04-20
Small ophys updates

### LINKED ISSUE #263: Small ophys items
- [x] make patch_cords Optional
- [x] Add camera to ophys session
- [x] Camera description/purpose
- [ ] Microscope class in device (what all does this need beyond base device class?)
- [x] magnification in FOV

saskiad:
closes #263

### DIFF
--- a/src/aind_data_schema/device.py
+++ b/src/aind_data_schema/device.py
@@ -245,10 +245,9 @@ class CameraAssembly(AindModel):
     """Named assembly of a camera and lens (and optionally a filter)"""
 
     # required fields
-    camera_assembly_name: str = Field(..., title="Name of this camera assembly")
+    camera_assembly_name: str = Field(..., title="Camera assembly name (describe purpose e.g. Body Camera)",)
     camera: Camera = Field(..., title="Camera")
     lens: Lens = Field(..., title="Lens")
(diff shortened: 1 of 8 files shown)


saskiad:
camera assembly name already being used to specify the purpose - eg. Eye Camera vs Body Camera. BUT, I think it would be helpful to have a specific field with an ENUM that can be better used to identify cameras

saskiad:
Until I have other field to add for microscope, I'm not adding it as a specific class and it will need to be an additional device. If there are other fields, then we can add it.

src/aind_data_schema/device.py
```diff
@@ -240,12 +240,21 @@ class Filter(Device):
         None, title="Description", description="More details about filter properties and where/how it is being used"
     )
 
+class CameraTarget(Enum):
```
dyf, line 243: I recently polled for behavior videos, if you want to update the Enum:
image

dyf (changes requested):

src/aind_data_schema/device.py
dyf: I think the only additions would be side vs body and tongue.

saskiad: Thanks. This allows confirms the need to have this field as relying on names won't yield useful searches!

dyf (approved):

Issue #262, opened 2023-04-20
Validator for units of optional values


saskiad:
**Is your feature request related to a problem? Please describe.**
When an optional value is provided, the unit must be reported. But if the unit is optional, this is unlikely to happen

**Describe the solution you'd like**
Write a validator to make sure units are provided for optional vlaues

**Describe alternatives you've considered**
Make units required (with default value) even when the value itself is optional

**Additional context**

saskiad:
Make a NumberWithUnit class that can be optional/required together

saskiad:
@dyf  Is this going to require separate FloatWithUnit and IntWithUnit classes? Or is there a way to do that generically?

dyf:
looking at this. if we go maximum generic, we probably also want to constrain the units beyond just `str`.

saskiad:
yes, most of the units we are using enums already

saskiad:
@Sun-flow let's discuss this on Monday, but this would be both a small item and a slightly bigger lift because it touches a lot of things. We have a strategy for how to do this that I can show you

PR #266, opened 2023-04-21
Feat 2p rig schema update

arielleleon:
@saskiad @dyf  @DowntonCrabby 

Added a few components to the device module and created an assembly module to group device components. The assembly classes have not been added to the rig b/c we would like your feedback before getting too far.


src/aind_data_schema/device.py
```diff
@@ -283,6 +255,11 @@ class DAQChannel(AindModel):
         False, title="Set to true if DAQ channel is sampled at irregular intervals"
     )
 
+class Computer(AindModel):
```
dyf, line 258: I'm not sure we need this class. Perhaps just `DAQDevice.computer_name = str` would suffice.

dyf (commented):
Some preliminary thoughts.  @saskiad should review further.

src/aind_data_schema/assembly.py
```diff
@@ -0,0 +1,85 @@
...
+
+    BODY = "Body"
+    BOTTOM = "Bottom"
+    EYE = "Eye"
+    FACE = "Face"
+    SIDE = "Side"
+    TONGUE = "Tongue"
+    OTHER = "Other"
+    
+class RelativePosition(AindModel):
```
saskiad, line 27: why not import this from device.py? we try to avoid redefining things in multiple places to help keep things aligned

src/aind_data_schema/assembly.py
```diff
@@ -0,0 +1,85 @@
...
+
+class WaterDelivery(AindModel):
+    """Description of water delivery system"""
+
+    # required fields
+    spout_diameter: str = Field(..., title="Spout diameter", units="mm")
+    spout_placement: RelativePosition = Field(..., title="Spout stage placement")
+    stage_type: str = Field(..., title="Stage build type")
+    calibration: dict = Field(..., title="Water calibration values")
+    solenoid_part: str = Field(..., title="Solenoid Part number")
```
saskiad, line 52: should solenoid be a device?

src/aind_data_schema/assembly.py
```diff
@@ -0,0 +1,85 @@
...
+    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
+
+
+class WaterDelivery(AindModel):
+    """Description of water delivery system"""
+
+    # required fields
+    spout_diameter: str = Field(..., title="Spout diameter", units="mm")
+    spout_placement: RelativePosition = Field(..., title="Spout stage placement")
+    stage_type: str = Field(..., title="Stage build type")
```
saskiad, line 50: what is stage_type? (I'm just unclear what this is)

src/aind_data_schema/assembly.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # optional fields
+    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
+
+
+class WaterDelivery(AindModel):
+    """Description of water delivery system"""
+
+    # required fields
+    spout_diameter: str = Field(..., title="Spout diameter", units="mm")
```
saskiad, line 48: need a spout_diameter_unit as a separate field using SizeUnit. Add (mm) to the title. See some of the other unit fields around as examples (or ask me and I can help)

src/aind_data_schema/assembly.py
```diff
@@ -0,0 +1,85 @@
...
+    spout_placement: RelativePosition = Field(..., title="Spout stage placement")
+    stage_type: str = Field(..., title="Stage build type")
+    calibration: dict = Field(..., title="Water calibration values")
```
saskiad, line 51: water_calibration_values: Dict[str, Any] = Field(..., title="Water calibration values")
-> add Dict to the typing imports. Also Any

We try to have the field name and the title close to the same

src/aind_data_schema/assembly.py
```diff
@@ -0,0 +1,85 @@
...
+    solenoid_part: str = Field(..., title="Solenoid Part number")
+
+class BehaviorPlatform(AindModel):
+    """Behavior platform for a mouse during a session"""
+    track_wheel: Union[Tube, Treadmill, Disc] = Field(..., title="Track wheel type")
```
saskiad, line 56: track_wheel isn't a good name here. None of these are wheels. Maybe we can call a disc a wheel... sure. But Tubes don't even allow running - by design.

src/aind_data_schema/assembly.py
```diff
@@ -0,0 +1,85 @@
...
+    """Description of water delivery system"""
+
+    # required fields
+    spout_diameter: str = Field(..., title="Spout diameter", units="mm")
+    spout_placement: RelativePosition = Field(..., title="Spout stage placement")
+    stage_type: str = Field(..., title="Stage build type")
+    calibration: dict = Field(..., title="Water calibration values")
+    solenoid_part: str = Field(..., title="Solenoid Part number")
+
+class BehaviorPlatform(AindModel):
```
saskiad, line 54: I share David's thought here.
I'm also unclear why these are grouped in this way. These are all the behavior equipment for your experiments? Do they need to be grouped this way? I might prefer to have these all as devices... But let me consider this some more ...

src/aind_data_schema/assembly.py
```diff
@@ -0,0 +1,85 @@
...
+    """Behavior platform for a mouse during a session"""
+    track_wheel: Union[Tube, Treadmill, Disc] = Field(..., title="Track wheel type")
+    
+    # optional fields
+    stage_software: Optional[Software] = Field(None, title="Stage software")
```
saskiad, line 59: this is the software that controls the stage for the water delivery? Is that correct? (stage is an common term, so we might need to add a modifier) If so, can we add this to the water delivery class directly?

src/aind_data_schema/assembly.py
```diff
@@ -0,0 +1,85 @@
...
+    camera_assembly_name: str = Field(..., title="Name of this camera assembly")
+    camera: Camera = Field(..., title="Camera")
+    lens: Lens = Field(..., title="Lens")
+    camera_target: CameraTarget = Field(..., title="Camera target")
+
+    # optional fields
+    filter: Optional[Filter] = Field(None, title="Filter")
+    position: Optional[RelativePosition] = Field(None, title="Relative position of this assembly")
+
+class VisualStimulusDisplayAssembly(AindModel):
```
saskiad, line 76: I'm unclear why this is here - it inherits all of these fields from Monitor.

src/aind_data_schema/device.py
```diff
@@ -44,6 +44,12 @@ class PowerUnit(Enum):
     UW = "microwatt"
     MW = "milliwatt"
 
+class Driver(Enum):
```
saskiad, line 47: what kind of Drivers are these? graphics? Add to the name

src/aind_data_schema/device.py
```diff
@@ -144,21 +150,13 @@ class DaqChannelType(Enum):
     DO = "Digital Output"
 
 
-class RelativePosition(AindModel):
```
saskiad, line 147: moving this to assembly.py outright assumes that all devices are part of assemblies, which is not true. This needs to be in device.py
The question I have is do we need a separate assembly.py file? I'm not sure we do

src/aind_data_schema/device.py
```diff
@@ -144,21 +150,13 @@ class DaqChannelType(Enum):
...
-    x: Optional[float] = Field(None, title="Position X (mm)", units="mm")
-    y: Optional[float] = Field(None, title="Position Y (mm)", units="mm")
-    z: Optional[float] = Field(None, title="Position Z (mm)", units="mm")
-    position_unit: SizeUnit = Field(SizeUnit.MM, title="Position unit")
-
-    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
+class Software(AindModel):
+    """ Generic rig software"""
 
+    software_name: str = Field(..., title="Software on computer")
```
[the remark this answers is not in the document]
saskiad: agreed

Though ... there might be code vs software? Eg. using ScanImage might not permit pointing to a code_url.

src/aind_data_schema/device.py
```diff
@@ -144,21 +150,13 @@ class DaqChannelType(Enum):
...
+    software_version: str = Field(..., title="Software version")
+    configuration: Optional[dict] = Field(..., title="Software configuration")
```
saskiad, line 158: see the syntax for parameters in the snippet David posted above to get the Dict right. And maybe sure Dict is imported.
Can we call configuration parameters?

src/aind_data_schema/device.py
```diff
@@ -395,8 +378,6 @@ class Monitor(Device):
     width: int = Field(..., title="Width (pixels)", units="pixels")
     height: int = Field(..., title="Height (pixels)", units="pixels")
     size_unit: SizeUnit = Field(SizeUnit.PX, title="Size unit")
-    viewing_distance: float = Field(..., title="Viewing distance (cm)", units="cm")
```
saskiad, line 398: why remove these from the device?

saskiad (changes requested):
I'm generally not fully sure having a separate assembly.py file makes sense. I'm happy to discuss to think it through together, but my instinct right now (also, my brain is a little fried from a long day) is to keep these in device.py

src/aind_data_schema/assembly.py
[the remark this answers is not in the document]
arielleleon: The MousePlatform is a base class for all surfaces that a mouse runs on (Disc, Tube, Treadmill). We were thinking that one could represent the entire behavior assembly (track wheel, water deliver, cameras, etc) as it's own system. After discussing this with Clark, we are going to remove the concept of a BehaviorPlatform to keep assemblies from becoming too nested.

src/aind_data_schema/device.py
```diff
@@ -144,21 +150,13 @@ class DaqChannelType(Enum):
...
-    position_unit: SizeUnit = Field(SizeUnit.MM, title="Position unit")
-
-    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
+class Software(AindModel):
+    """ Generic rig software"""
 
+    software_name: str = Field(..., title="Software on computer")
+    software_version: str = Field(..., title="Software version")
+    configuration: Optional[dict] = Field(..., title="Software configuration")
+    description: str = Field(..., title="Description of where and how software are used")
```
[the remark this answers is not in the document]
arielleleon: We could make this optional. Out of curiosity, why do you think this is unnecessary?

src/aind_data_schema/device.py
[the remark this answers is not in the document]
arielleleon: I agree with @saskiad latest comment. The software I am referring to is software used to manipulate hardware/acquire data and is typically versioned and in either PyPi, aibsdevpi (Engineering hosted server), or other 3rd party SW. What do you think about describing this better in the docstring associated with the class so folks can delineate processing vs hardware control software?

src/aind_data_schema/device.py
```diff
@@ -368,6 +345,11 @@ class Disc(MousePlatform):
     platform_type: str = Field("Disc", title="Platform type", const=True)
     radius: float = Field(..., title="Radius (cm)", units="cm", ge=0)
     radius_unit: SizeUnit = Field(SizeUnit.CM, title="radius unit")
+    analog: Optional[bool] = Field(None,description="analog or digital electronics")
```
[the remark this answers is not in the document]
arielleleon: Yes

src/aind_data_schema/device.py
arielleleon: I agree that we can remove it. It is only used in DAQDevice now. If a computer_name is used in multiple locations further down the line it would be nice to have a computer base class. That can be added later.

src/aind_data_schema/device.py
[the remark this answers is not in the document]
saskiad: I think I want a software class that has similar things to what we're tracking for the processing code:
software, version, parameters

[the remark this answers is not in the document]
arielleleon: @saskiad  Ok, we left the software class in device and renamed the keys to 'name', 'version' and 'configuration'. I removed the 'description' per earlier advice.

src/aind_data_schema/assembly.py
arielleleon: Agreed and moving.

src/aind_data_schema/assembly.py
arielleleon: We can do that and there is a built in unit option in PyDantic I thought was useful. Out of curiosity, why is that not used?

src/aind_data_schema/assembly.py
arielleleon: Got it and done

src/aind_data_schema/device.py
arielleleon: These are device drivers or firmware. Changed name to DeviceDriver

src/aind_data_schema/device.py
arielleleon: Correct, we do not need an assembly.py file and it's being removed.

src/aind_data_schema/device.py
arielleleon: I use the term configuration because you configure software with a set of parameters. Happy to change it though

src/aind_data_schema/device.py
arielleleon: We removed this because we added it to an assembly instance. It has been added back to Monitor

src/aind_data_schema/assembly.py
saskiad: because it doesn't show up in the json. Originally that's how we were doing the units, but they just live in the schema code and don't show up in the json. So that's why we switched the the cumbersome parameter_value and parameter_value_unit system we have now. 
We actually have an open issue (#262 ) to streamline this a little bit - I'll give you a heads up when we implement that

src/aind_data_schema/device.py
```diff
@@ -413,4 +393,77 @@ class Monitor(Device):
...
+    # optional fields
+    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
+
+
+class WaterDelivery(AindModel):
+    """Description of water delivery system"""
+
+    # required fields
+    spout_diameter: str = Field(..., title="Spout diameter")
+    spout_diameter_units_mm: SizeUnit = Field(..., title="")
```
saskiad, line 432: spout_diameter_unit: SizeUnit = Field(SizeUnit.MM, title="Spout diameter unit")

src/aind_data_schema/device.py
```diff
@@ -413,4 +393,77 @@ class Monitor(Device):
...
+
+    # optional fields
+    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
+
+
+class WaterDelivery(AindModel):
+    """Description of water delivery system"""
+
+    # required fields
+    spout_diameter: str = Field(..., title="Spout diameter")
```
saskiad, line 431: Add (mm) to the title for Spout diameter.

src/aind_data_schema/device.py
```diff
@@ -413,4 +393,77 @@ class Monitor(Device):
...
+    spout_diameter_units_mm: SizeUnit = Field(..., title="")
+    spout_placement: RelativePosition = Field(..., title="Spout stage placement")
+    stage_type: str = Field(..., title="Stage build type")
```
saskiad, line 434: what is stage_type? Should this be an ENUM of possible types?

src/aind_data_schema/device.py
```diff
@@ -413,4 +393,77 @@ class Monitor(Device):
...
+    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
+
+
+class WaterDelivery(AindModel):
+    """Description of water delivery system"""
+
+    # required fields
+    spout_diameter: str = Field(..., title="Spout diameter")
+    spout_diameter_units_mm: SizeUnit = Field(..., title="")
+    spout_placement: RelativePosition = Field(..., title="Spout stage placement")
```
saskiad, line 433: can we call this spout_position? (and adjust title too)

src/aind_data_schema/device.py
```diff
@@ -413,4 +393,77 @@ class Monitor(Device):
...
+    stage_type: str = Field(..., title="Stage build type")
+    water_calibration_values: Dict[str, Any] = Field(..., title="Water calibration values")
+    solenoid_part: str = Field(..., title="Solenoid Part number")
+
+class BehaviorPlatform(AindModel):
```
saskiad, line 438: I'm unclear why this is BehaviorPlatform instead of the MousePlatform we currently have. Especially since Tube is decidedly not a behavior platform.

src/aind_data_schema/device.py
```diff
@@ -413,4 +393,77 @@ class Monitor(Device):
...
+    """Behavior platform for a mouse during a session"""
+    track_wheel: Union[Tube, Treadmill, Disc] = Field(..., title="Track wheel type")
```
saskiad, line 440: likewise, this really should not be track wheel when only one is (arguably) a wheel

src/aind_data_schema/device.py
```diff
@@ -413,4 +393,77 @@ class Monitor(Device):
...
+    
+    # optional fields
+    stage_software: Optional[Software] = Field(None, title="Stage software")
```
saskiad, line 443: I'm confused what the stage software is - this controls the water spout?

src/aind_data_schema/device.py
```diff
@@ -170,6 +167,10 @@ class Device(AindModel):
     notes: Optional[str] = Field(None, title="Notes")
 
 
+class Solenoid(Device):
+    material: Optional[str] = Field(None, title="Material type")
```
saskiad, line 171: Should this be an ENUM? What are the material types this could be?

src/aind_data_schema/device.py
```diff
@@ -170,6 +167,10 @@ class Device(AindModel):
...
+    replacement_date: Optional[datetime] = Field(None, title="In use since")
```
saskiad, line 172: can we make the title be Replacement date? I assume this is a part that degrades with time and it's function changes this with time?

src/aind_data_schema/device.py
```diff
@@ -413,4 +393,77 @@ class Monitor(Device):
...
+class WaterDelivery(AindModel):
+    """Description of water delivery system"""
+
+    # required fields
+    spout_diameter: str = Field(..., title="Spout diameter")
+    spout_diameter_units_mm: SizeUnit = Field(..., title="")
+    spout_placement: RelativePosition = Field(..., title="Spout stage placement")
+    stage_type: str = Field(..., title="Stage build type")
+    water_calibration_values: Dict[str, Any] = Field(..., title="Water calibration values")
+    solenoid_part: str = Field(..., title="Solenoid Part number")
```
saskiad, line 436: This will be in the solenoid device class, we don't need it here as well, right?

src/aind_data_schema/device.py
arielleleon: Done

src/aind_data_schema/device.py
arielleleon: Done

src/aind_data_schema/device.py
arielleleon: Done - changed to MousePlatform

src/aind_data_schema/device.py
arielleleon: @saskiad  Is the tube more analogous to the head frame than it is to the treadmill/disc? If that's the case I would argue that the tube be separated from two. What do you think?

src/aind_data_schema/device.py
arielleleon: This is the firmware associated with the stage. Is it necessary? If so, I should rename it to firmware.

src/aind_data_schema/device.py
arielleleon: @saskiad  I'm thinking of removing this since it's most likely going to be aluminum. Also the material type is listed in the part number. I do think the replacement date should stay though. You OK if I remove it?

src/aind_data_schema/device.py
arielleleon: True. Removing

src/aind_data_schema/device.py
arielleleon: I think so but now I am second guessing myself and I am checking with systems engineers on Colin's team. I may remove this device if found stable, otherwise, I will change the title according to your suggestion.

src/aind_data_schema/device.py
saskiad: no, it's a literal tube that holds the mouse, so it's more like the treadmill or disc - except it largely prevents the mouse from running rather than enables it.

saskiad: Screen Shot 2023-05-11 at 4 11 11 PM

src/aind_data_schema/device.py
saskiad: sounds good

src/aind_data_schema/device.py
saskiad: my confusion is more about "stage" - I think of stage as being a platform and being controlled by the data acquisition software, so I'm more confused what this software controls. I think I just don't know what the stage is.

src/aind_data_schema/device.py
arielleleon: After checking with systems engineers they did say it's the fittings that need to be replaced not the solenoid itself. I will remove this.

src/aind_data_schema/device.py
arielleleon: @saskiad What about mobility_aid?

src/aind_data_schema/device.py
DowntonCrabby: Arielle and I just talked about this a bit- we were thinking the stage is basically the motorized platform that moves the thing the mouse is on in to place. 

Perhaps we can change the title, or make a more general device as I think there are multiple of these types of stages, for example on the scientifica its the microscope that's on a stage that moves in to place, on the mesoscope both the microscope and the animal are on stages that move. The lickspout also has a motorized stage that moves it in to place.

src/aind_data_schema/device.py
DowntonCrabby: mouse ambulation context ?

src/aind_data_schema/device.py
DowntonCrabby: See the comment I left above about what we were thinking a stage is. In that context I think there's a lot of things on stages that move around in the behavior enclosure like the lick spout, the actual thing that the wheel/disc etc sits on that moves in to place, etc.  Sometimes that acquisition software controls it and sometimes not- each of these could have their own hardware with it's own firmware/software.

src/aind_data_schema/device.py
saskiad: thanks - this helps! I think I understand better! I think motorized stage might be the the thing to use - I think we have that in the imaging instruments (the SPIMs) so perhaps we can leverage that.

But then I would expect this to be outside of Behavior Platform class - since sometimes it's the microscope that's on the stage and sometimes its the mouse.

src/aind_data_schema/device.py
saskiad: Again, one context is specifically non-ambulatory or mobile. Is there a specific reason you don't want Mouse Platform?

DowntonCrabby:
@saskiad we went through and tried to fix all the linting errors associated with our work. It looks like there are still some failures but those are related to items that we didn't touch so we weren't sure how to proceed.

saskiad:
Changes to the DAQ device are going to impact a bunch of ephys and imaging things, I'll work on that.

dyf (changes requested):

dyf (approved):

saskiad (approved):

Issue #249, opened 2023-04-17
move `EphysModule` fields to `EphysSession`


dyf:
- arc angle
- module angle
- rotation angle
- transform

also: 
- fix the "data" typo in the transform description
- add a lens to `StickMicroscope`

jsiegle:
Rig JSON:
- Switch "port index" description to use one-based indexing
- Remove radius units for disc platform
- Add Neuropixels probe headstage models, and optional serial number field for headstages
- Add "Notes" field to `DomeModule` class

Session JSON:
- For each probe, add manipulator coordinate for brain surface location (Z coordinate only)

saskiad:
I'm not sure we should remove the radius units - its important for being able to determine the distance the mouse runs on the disc.

saskiad:
I don't think we can make serial numbers optional for some devices and not others. Can you just enter N/A for headstages?

jsiegle:
> I'm not sure we should remove the radius units - its important for being able to determine the distance the mouse runs on the disc.

The issue is that the unit is specified both in the `radius` field and in the `radius_unit` field, and it's currently possible for these to conflict: https://github.com/AllenNeuralDynamics/aind-data-schema/blob/2d18f1657f4faf747af26e9da71fdcbc45c35b4f/src/aind_data_schema/device.py#L334-L335

> I don't think we can make serial numbers optional for some devices and not others. Can you just enter N/A for headstages?

This is fine

saskiad:
ah, sorry I misread the radius unit comment.  
We need the radius unit. It will show up in json as "radius": 5. Even though the title has a unit in it, that title doesn't show up in the json. This is how the units work throughout the schema. It was a sad day adding those in.

saskiad:
closed by PR #277

PR #277, opened 2023-04-29
moving ephys module positions to ephys session

saskiad:
WIP to move the dome module to session (and deal with other items listed in issue #249


saskiad:
Take a look at this and lmk what you think - I haven't dealt with tests and validators yet, just trying to get the logic right first

jsiegle:
In `ephys_session.py`, `ManipulatorModule` needs to inherit from `DomeModule` (since every manipulator module also has a position on the dome). And there needs to be a separate `EphysModule` class (similar to `LaserModule`) that holds a `List` of `EphysProbes`, since there can be multiple probes linked to single manipulator.

saskiad:
@jsiegle just clarifying that there can be multiple ephys probes in a manipulator. So multiple ephys probes can have the same DomeModule angles, etc, as well as the same ManipulatorModule coordinates? So this can be solved by making the EphysProbe class contain a list of probes, yes?

jsiegle:
I think we still want to differentiate between `EphysProbe` (each of which generates a unique data stream) and `EphysModule` (which represents probes attached to the same manipulator). But yes, the probes on each module will share manipulator coordinates, target CCF coordinates, etc.

saskiad:
cool - I'll work on the test and examples and let you know when it's ready for review

saskiad:
@jsiegle ready for review now.

src/aind_data_schema/ephys/ephys_rig.py
```diff
@@ -107,37 +107,24 @@ class OpenEphysAcquisitionBoard(DAQDevice):
...
     camera: Camera = Field(..., title="Camera for this module")
+    lens: Lens = Field(..., title="Lens for this module")
 
 
-class LaserModule(DomeModule):
-    """Module for optogenetic stimulation"""
+class LaserAssembly(AindModel):
+    """Assembly for optogenetic stimulation"""
 
+    assembly_name: str = Field(..., title="Laser assembly name")
```
jsiegle, line 127: Change to `laser_assembly_name` for consistency

jsiegle (approved):
Just one minor change needed. Otherwise I think this looks good!

dyf:
@jsiegle is a "module" the same thing as an "assembly"? if so, can we just use the term "assembly"?

jsiegle:
We like using the term "module" to refer to a module on the dome insertion system, whereas "assembly" is more general (i.e., a `CameraAssembly` can be part of the behavior setup).

PR #283, opened 2023-05-06
Add first pass at behavior rig

saskiad:
added a simple behavior rig schema, and made a behavior folder


saskiad:
Something we need to think about is whether the behavior "rig" is something that gets inherited into ephys and ophys rigs? Or if it should just be a parallel for a behavior only context. We can discuss later - I think that's beyond the scope of this PR

saskiad:
closes #284

src/aind_data_schema/behavior/behavior_rig.py
```diff
@@ -0,0 +1,39 @@
...
+    HarpDevice,
+    Monitor,
+    Treadmill,
+    Tube,
+)
+
+class BehaviorRig(AindCoreModel):
+    """Description of an behavior rig"""
+
+    describedBy: str = Field(
```
dyf, line 24: you can remove `describedBy`.  It's automatically included by inheriting from `AindCoreModel`.

src/aind_data_schema/behavior/behavior_session.py
```diff
@@ -7,13 +7,19 @@
...
-from .base import AindCoreModel
-from .procedures import VolumeUnit, WeightUnit
+from ..base import AindCoreModel
+from ..procedures import VolumeUnit, WeightUnit
 
 
 class BehaviorSession(AindCoreModel):
     """Description of a behavior session"""
 
+    describedBy: str = Field(
```
dyf, line 17: ditto

dyf (changes requested):

src/aind_data_schema/behavior/behavior_rig.py
saskiad: Ah! excellent. Let's remove these from other schemas - I'll make an issue

dyf (approved):

PR #294, opened 2023-05-11
Adding PIDName class

Sun-flow:
Plan to solve:

    Introduce class similar to BaseName, which takes inputs for species name, PID, and database info was acquired from
    PIDName(name=str,ID=str,database=str)
    Will place in base, and utilize in subject.species

### DIFF
--- a/examples/subject.json
+++ b/examples/subject.json
@@ -4,7 +4,7 @@
    "species": {
       "name": "Mus musculus",
       "ID": "10090",
-      "Database": "NCBI"
+      "database": "NCBI"
    },
    "subject_id": "12345",
    "sex": "Male",
--- a/examples/subject.json
+++ b/examples/subject.json
@@ -3,9 +3,8 @@
    "schema_version": "0.3.1",
    "species": {
       "name": "Mus musculus",
-      "PID": "10090",
-      "database_name": "NCBI",
-      "database_url": "https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=10090&lvl=3&p=has_linkout&p=blast_url&p=genome_blast&lin=f&keep=1&srchmode=1&unlock"
+      "registry": "NCBI",
+      "registry_identifier": "10090"
    },
    "subject_id": "12345",
    "sex": "Male",
--- a/examples/subject.json
+++ b/examples/subject.json
@@ -1,7 +1,11 @@
 {
    "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/subject.py",
-   "schema_version": "0.3.0",
-   "species": "Mus musculus",
+   "schema_version": "0.3.1",
+   "species": {
+      "name": "Mus musculus",
+      "ID":"10090",
+      "Database":"NCBI"
+   },
    "subject_id": "12345",
    "sex": "Male",
    "date_of_birth": "2022-11-22",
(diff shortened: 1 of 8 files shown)


examples/subject.json
```diff
@@ -1,7 +1,11 @@
 {
    "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/subject.py",
-   "schema_version": "0.3.0",
-   "species": "Mus musculus",
+   "schema_version": "0.3.1",
+   "species": {
+      "name": "Mus musculus",
+      "ID": "10090",
+      "database": "NCBI"
```
saskiad, line 7: Let's have the database be the url for the taxonomy so there's absolutely no confusion

src/aind_data_schema/base.py
```diff
@@ -51,6 +51,12 @@ class BaseName(AindModel):
     name: str
     abbreviation: str
 
+class PIDName(AindModel):
+    """Model tracking a species name, PID, and database PID was acquired from"""
+    name: str
+    ID: str
```
saskiad, line 57: it looks like ID might be a specific thing in pydantic (hence the different color). Maybe call this "identifier"?

src/aind_data_schema/base.py
```diff
@@ -51,6 +51,12 @@ class BaseName(AindModel):
     name: str
     abbreviation: str
 
+class PIDName(AindModel):
+    """Model tracking a species name, PID, and database PID was acquired from"""
+    name: str
```
saskiad, line 56: ah... some syntax here:

name: str = Field(..., title="Name") 

yeah, David didn't do that for the BaseName... but generally we want to stick with this syntax. For database, add a description that instructs users something like "URL for the external database"

src/aind_data_schema/base.py
```diff
@@ -51,6 +51,12 @@ class BaseName(AindModel):
...
+    ID: str
+    database: str
```
saskiad, line 58: thought I had from below - I think maybe add database_name and database_url

saskiad (changes requested):
see comments

dyf (approved):
love it

src/aind_data_schema/base.py
```diff
@@ -47,10 +49,15 @@ def __call__(cls, value, *args, **kw):
 
 class BaseName(AindModel):
     """A simple model associating a name with an abbreviation"""
-
     name: str
     abbreviation: str
 
+class PIDName(AindModel):
+    """Model tracking a species name, PID, and database PID was acquired from"""
```
saskiad, line 56: "Model for associate a name with a persistent identifier (PID) and the registry for that PID"

Also, add a line after the description (just to match our formatting) :)

src/aind_data_schema/base.py
```diff
@@ -47,10 +49,15 @@ def __call__(cls, value, *args, **kw):
...
+    name: str = Field(..., title="Name")
+    registry: str = Field(..., title="Registry")
```
saskiad, line 58: I think for now this is good - I think soon we might make the registry an ENUM of the databases we use. But we can do that when we apply PIDName to other things!

saskiad (changes requested):
one minor edit. Otherwise it looks good!

src/aind_data_schema/base.py
saskiad: Noting that we've decided not to add the registry url

saskiad (approved):
Awesome!!

Sun-flow:
close #294

PR #296, opened 2023-05-11
variable intervals

saskiad:
accommodate situation where the intervals are variable


mekhlakapoor (commented):
in the case that an interval is not fixed, how will the variable pulse train intervals be tracked?

mekhlakapoor (approved):

saskiad:
This is something we need to figure out - right now they are generated randomly, so I think that it's not something we can track in the metadata and I would expect it to be represented in the data itself. Honestly, I'm not sure this level of information should be in the metadata and should be in the stimulus table in the NWB file.

Issue #305, opened 2023-05-13
add sensory stimulus class to Stimulus


saskiad:
Needed for the RF mapping ecephys datasets.

saskiad:
closed with #306

PR #306, opened 2023-05-13
Adding visual stim class

saskiad:


saskiad:
closes #305

src/aind_data_schema/stimulus.py
```diff
@@ -44,10 +44,53 @@ class OptoStim(AindModel):
...
+        ..., 
+        title="Number of conditions",
+        description="Number of unique stimulus conditions",
+    )
+    stimulus_parameters: Optional[Dict[str, Any]] = Field(
+        None, 
+        title="Stimulus parameters",
+        description="Define and list the parameter values used (e.g. all TF or orientation values)"
+    )
+    stimulus_template_name: Optional[List[str]] = Field(
```
dyf, line 61: is this supposed to be a list?

dyf (changes requested):
This looks like it's meant to cover multiple stimuli at once, but if you link to it from a `StimulusPresentation` (which I think is meant to describe which stimulus was used for a given trial), then there's no way to know which stimulus and which parameters were used.  So, whatever is linked to in `StimulusPresentation` should refer to a specific "stimulus condition," I think, rather than a class of stimuli.

dyf (approved):

PR #328, opened 2023-05-22
Feat 262 value with unit

### LINKED ISSUE #262: Validator for units of optional values

Sun-flow:
closes #262 

Adding Unit class with enumerations for measurement types, as well as generic BlankVal types that can be utilized to create a <Value, Unit> Pairing.

Docs need to be updated to reflect usage of this class


src/aind_data_schema/utils/units.py
```diff
@@ -0,0 +1,112 @@
...
+class Units:
+    """Class containing enumerations of relevant measurement types"""
+
+    class Size(Enum):
+        """Enumeration of Length Measurements"""
+
+        MM = "millimetre"
+        CM = "centimetre"
+        M = "metre"
+        KM = "kilometre"
```
saskiad, line 18: I might leave off kilometer since we don't have any sizes that large.
Also, we need to add inches and pixels as size measurements. (sadly there are devices that are measured in inches - mostly optics).
I'd prefer using the "meter" spelling, but can be outvoted if @dyf has a preference

src/aind_data_schema/utils/units.py
```diff
@@ -0,0 +1,112 @@
...
+        G = "gram"
+        MG = "milligram"
+
+    class Frequency(Enum):
+        """Enumeration of Frequency Measurements"""
+
+        mHZ = "millihertz"
+        HZ = "hertz"
+        KHZ = "kilohertz"
+        MHZ = "megahertz"
```
saskiad, line 33: likewise, I don't think we use GHZ, THZ. Not even sure about MHZ ... 

We need a consistent milli vs mega prefix. Here milli is "m" while for mass it is "M" I might recommend using M and using "ME" for mega because I worry about casing getting lost.

src/aind_data_schema/utils/units.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class GenericValues:
+    """Constructs UnitWithValue class for each relevant measurement type"""
+
+    class SizeValue(GenericModel, Generic[ScalarType, Units.SizeType]):
+        """Generic for Size Measurements"""
+
+        value: ScalarType
+        unit: Units.SizeType = Units.Size.M
```
saskiad, line 74: let's default to MM here

src/aind_data_schema/utils/units.py
```diff
@@ -0,0 +1,112 @@
...
+        """Generic for Mass Measurements"""
+
+        value: ScalarType
+        unit: Units.MassType = Units.Mass.G
+
+    class VolumeValue(GenericModel, Generic[ScalarType, Units.VolumeType]):
+        """Generic for Volume Measurements"""
+
+        value: ScalarType
+        unit: Units.VolumeType = Units.Volume.L
```
saskiad, line 86: Let's default to NL

src/aind_data_schema/utils/units.py
```diff
@@ -0,0 +1,112 @@
...
+        KHZ = "kilohertz"
+        MHZ = "megahertz"
+        GHZ = "gigahertz"
+        THZ = "terahertz"
+
+    class Volume(Enum):
+        """Enumeration of Volume Measurements"""
+
+        L = "litre"
+        ML = "millilitre"
```
saskiad, line 41: Also add UL and NL (micro and nano) - I think these will be our more common units.

src/aind_data_schema/utils/units.py
```diff
@@ -0,0 +1,112 @@
...
+        """Generic for Frequency Measurements"""
+
+        value: ScalarType
+        unit: Units.FrequencyType = Units.Frequency.HZ
+
+    class AngleValue(GenericModel, Generic[ScalarType, Units.AngleType]):
+        """Generic for Angle Measurements"""
+
+        value: ScalarType
+        unit: Units.AngleType = Units.Angle.RAD
```
saskiad, line 98: default to DEG

saskiad (changes requested):
small changes - otherwise I think this looks good!

src/aind_data_schema/utils/units.py
```diff
@@ -0,0 +1,110 @@
...
+
+from pydantic.generics import GenericModel
+
+
+class Units:
+    """Class containing enumerations of relevant measurement types"""
+
+    class Size(Enum):
+        """Enumeration of Length Measurements"""
+
```
saskiad, line 14: Let's use the "meter" spelling since that's what the standards we want to feed into use.

src/aind_data_schema/utils/units.py
```diff
@@ -0,0 +1,110 @@
...
+    class Frequency(Enum):
+        """Enumeration of Frequency Measurements"""
+
+        mHZ = "millihertz"
+        HZ = "hertz"
+        KHZ = "kilohertz"
+
+    class Volume(Enum):
+        """Enumeration of Volume Measurements"""
+
```
saskiad, line 35: Likewise here, let's use "liter"

saskiad (changes requested):
After consulting our related standards, we're going with American spellings. Sorry Canadian friends :)

src/aind_data_schema/utils/units.py
dyf: for the record, we should use SI names, which means "er" not "re".

Sun-flow: Fixed now!

Sun-flow:
@dyf @saskiad requested changes have been addressed!

saskiad (approved):
Looks good!

PR #349, opened 2023-05-26
Adding stack class

### LINKED ISSUE #345: Create stack class in ophys session

saskiad:
closes #345


src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -157,3 +157,27 @@ class TwoPhotonOphysSession(OphysSession):
...
+    number_of_plane_repeats_per_volume: int = Field(..., title="Number of repeats per volume")
+    number_of_volume_repeats: int = Field(..., title="Number of volume repeats")
+    fov_coordinate_ml: float = Field(..., title="FOV coodinate ML")
+    fov_coordinate_ap: float = Field(..., title="FOV coordinate AP")
+    fov_coordinate_unit: SizeUnit = Field(SizeUnit.UM, title="FOV coordinate unit")
+    fov_reference: str = Field(..., title="FOV reference", description="Reference for ML/AP coordinates")
+    fov_width: int = Field(..., title="FOV width (pixels)")
+    fov_height: int = Field(..., title="FOV height (pixels)")
+    fov_size_unit: SizeUnit = Field(SizeUnit.PX, title="FOV size unit")
+    magnification: str = Field(..., title="Magnification")
```
KasparP, line 178: magnification is a bit unclear and archaic for laser scanning. Is pixel size not sufficient?

KasparP (commented):
Maybe add 
dimension order, number of channels 
as fields?

src/aind_data_schema/ophys/ophys_session.py
```diff
@@ -157,3 +157,27 @@ class TwoPhotonOphysSession(OphysSession):
...
     fovs: List[FieldOfView] = Field(..., title="Fields of view", unique_items=True)
+
+class Stack(OphysSession):
+    """Description of a two photon stack"""
+
+    start_depth: int = Field(..., title="Starting depth (um)")
+    end_depth: int = Field(..., title="Ending depth (um)")
+    depth_unit: SizeUnit = Field(SizeUnit.UM, title="Depth unit")
+    number_of_planes: int = Field(..., title="Number of planes")
+    step_size: float = Field(SizeUnit.UM, title="Step size")
```
dyf, line 168: Should probably be `Field(..., title="Step size")`

dyf: assuming that `depth_unit` applies to this field.

dyf (changes requested):

src/aind_data_schema/ophys/ophys_session.py
saskiad: anticipating valuewithunit coming soon, I'm keeping these separate since one is an int and one is a float

src/aind_data_schema/ophys/ophys_session.py
saskiad: I'm going to keep this and make it optional - I think the mesoscope team uses it

saskiad:
I've added a channel concept to both rig and the stack session. I kept detectors and filters in the rig outside of the channel concept to be safe, but we can probably remove the detector list in the rig schema. 
I've only added the channel to the stack session type - we might want to add it to the TwoPhotonOphysSession or FOV classes as well...

KasparP (approved):
thanks for addressing my comments

src/aind_data_schema/ophys/ophys_rig.py
```diff
@@ -60,11 +60,21 @@ class Patch(Device):
     photobleaching_date: Optional[date] = Field(None, title="Photobleaching date")
 
 
+class Channel(AindModel):
```
dyf, line 63: Should we consider combining this with `imaging.tile.Channel`?

dyf (changes requested):

saskiad:
closes #430

src/aind_data_schema/ophys/ophys_rig.py
```diff
@@ -75,9 +75,10 @@ class OphysRig(AindCoreModel):
     vibration_control: Optional[bool] = Field(None, title="Vibration control")
     patch_cords: Optional[List[Patch]] = Field(..., title="Patch cords", unique_items=True)
     light_sources: List[Union[Laser, LightEmittingDiode]] = Field(..., title="Light sources", unique_items=True)
-    detectors: Optional[List[Detector]] = Field(None, title="Detectors", unique_items=True)
+    channels: Optional[List[Channel]] = Field(None, title="Channels", unique_items=True)
```
dyf, line 78: I think we decided channels are in session only.

dyf (changes requested):
I am wondering if the right way to model this is instead to say that there is an OphysSession, and an OphysSession can have:
- planar fields of view
- stacks (which are fields of view with depth and step size and repeats)
- patch coords and coupling arrays

saskiad:
> I am wondering if the right way to model this is instead to say that there is an OphysSession, and an OphysSession can have:
> 
> * planar fields of view
> * stacks (which are fields of view with depth and step size and repeats)
> * patch coords and coupling arrays

I think I want to hold off doing this for when we merge the sessions across everything - I worry this will disrupt current ingest things, and I'd prefer to do that as few times as possible.

saskiad:
replaced by #479

PR #355, opened 2023-05-30
new ValueWithUnit implementation

### LINKED ISSUE #353: refactor UnitWithValue to apply new method
@dyf offers a new method for UnitWithValue https://codeocean.allenneuraldynamics.org/capsule/5515684/tree

consider refactoring to this method and avoiding generics

Sun-flow:
closes #353 

Updating ValueWithUnit method to utilize a simpler implementation. Usage remains the same, user must import Measure and MeasureValue from Units, then instantiate a value in the form:

value = MeasureValue(value=, unit=Measure.unit

### DIFF
--- a/src/aind_data_schema/utils/units.py
+++ b/src/aind_data_schema/utils/units.py
@@ -5,7 +5,6 @@ from enum import Enum
 from pydantic import create_model
 
 
-
 class Size(Enum):
     """Enumeration of Length Measurements"""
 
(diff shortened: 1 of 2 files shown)


saskiad:
I'm inclined to keep the enums to the units we actually use. Eg. g but not kg. deg but not rad. Partly for keeping things clean and partly for preventing confusion. But also willing to go with it with more units than we use if others prefer that

dyf (approved):

PR #371, opened 2023-06-13
Feat 363 analysis step

Sun-flow:


src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+from aind_data_schema.processing import ProcessName
+
+
+class TouchType(Enum):
+    ANALYSIS = "Analysis"
+    PROCESSING = "Processing"
+
+    ## These could be subtypes of analysis/processing
+
+class ProcessType(Enum):
```
saskiad, line 22: how is this different from the ProcessType in processing.py?

saskiad: sorry, ProcessName

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+    CURATION = "Curation"
+    SORTING = "Sorting"
+
+class AnalysisType(Enum):
+    uhhh = "some analysis"
+
+
+
+
+class DataTouchInfo:
```
saskiad, line 32: need doc strings for this class

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
+""" schema for processing """
```
saskiad, line 1: details, but this is a schema for all data manipulations, right? We need to be consistently clear about the distinction between processing and analysis.

src/aind_data_schema/datamanipulation.py
saskiad, line 32: need doc strings for this class.
Also can this be "DataTouch" just for brevity sake?

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+    SORTING = "Sorting"
+
+class AnalysisType(Enum):
+    uhhh = "some analysis"
+
+
+
+
+class DataTouchInfo:
+    touch_type: TouchType = Field(..., title="Manipulation type")
```
saskiad, line 33: Is there a reason the title isn't the same as the name itself? Or vice versa?

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+    start_date_time: datetime = Field(..., title="Start date time")
+    end_date_time: datetime = Field(..., title="End date time")
+    input_location: str = Field(..., description="Path to data inputs", title="Input location")
+    output_location: str = Field(..., description="Path to data outputs", title="Output location")
+    notes: Optional[str] = None
+
+
+class CodeManipulation:
```
saskiad, line 41: need docstring

src/aind_data_schema/datamanipulation.py
saskiad, line 32: Does this need the code manipulation class? (or code information?)

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+    outputs: Optional[Dict[str, Any]] = Field(None, description="Output parameters", title="Outputs")
+
+
+class AnalysisStep(AindModel):
+    """Description of a single processing step"""
+
+    analysis_name: str = Field(..., description="Name of the analysis method used", title="Analysis name")
+    analysis_type: AnalysisType = Field(..., description="Type of analysis performed on dataset", title="Analysis type")
+    touch_info: DataTouchInfo = Field(..., description="General information regarding the data manipulation performed")
+    analysis_code: CodeManipulation = Field(..., description="Info regarding code used to manipulate data")
```
saskiad, line 55: yeah, I think I prefer this to be in the data touch class.

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class DataProcess(AindModel):
+    """Description of a single processing step"""
+
+    process_name: ProcessName = Field(..., title="Process name")
+    processing_type: ProcessType = Field(..., description="Type of processing performed on dataset", title="Processing type")
```
saskiad, line 62: ah, I haven't yet indoctrinated you to my aversion to the use of "dataset". Dataset is used super ambiguously to refer either to a single data experiment or a conglomeration of experiments. I prefer to use it for the later, but try to avoid it because of the ambiguity. So I prefer data asset to refer to the data from a single experiment. Data is also acceptable by itself.

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+class CodeManipulation:
+    code_url: str = Field(..., description="Path to code respository", title="Code URL")
+    code_version: Optional[str] = Field(None, description="Version of the code", title="Code version")
+    input_parameters: Dict[str, Any] = Field(..., title="Input parameters")
+    output_parameters: Optional[Dict[str, Any]] = Field(..., title="Output parameters")
+    outputs: Optional[Dict[str, Any]] = Field(None, description="Output parameters", title="Outputs")
+
+
+class AnalysisStep(AindModel):
+    """Description of a single processing step"""
```
saskiad, line 50: single analysis step?

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+
+    schema_version: str = Field(
+        "0.2.1",
+        description="Schema version",
+        title="Schema version",
+        const=True,
+    )
+    pipeline_version: Optional[str] = Field(None, description="Version of the pipeline", title="Pipeline version")
+    pipeline_url: Optional[str] = Field(None, description="URL to the pipeline code", title="Pipeline URL")
+    data_processes: List[AnalysisStep] = Field(..., title="Data processing", unique_items=True)
```
saskiad, line 78: this should be a list of ProcessStep, right?

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class DataProcess(AindModel):
+    """Description of a single processing step"""
+
+    process_name: ProcessName = Field(..., title="Process name")
+    processing_type: ProcessType = Field(..., description="Type of processing performed on dataset", title="Processing type")
+    touch_info: DataTouchInfo = Field(..., description="General information regarding the data manipulation performed")
+    analysis_code: CodeManipulation = Field(..., description="Info regarding code used to manipulate data")
+
```
saskiad, line 65: do we need an Analysis model that creates a list of analysis steps?

src/aind_data_schema/processing.py
```diff
@@ -10,6 +10,7 @@
 
 from aind_data_schema.base import AindCoreModel, AindModel
 from aind_data_schema.imaging.tile import Tile
+from aind_data_schema.datamanipulation import DataTouchInfo
```
saskiad, line 13: I'm a little confused - are we keeping this schema? How is this different from the processing in the datamanipulation file?

Sun-flow: relic of an earlier edit, ideally we could drop the processing.py file after this update, though we could also have two separate files for analysis + processing if we wanted

src/aind_data_schema/datamanipulation.py
Sun-flow: DataTouch is imported by AnalysisStep and ProcessStep. There are two structure options here: 

DataTouch (overarching "Step" class):
- class for basic info 
- either process/analysis tag
- class for relevent step (analysis/processing)
- class for codemanipulation info

but it seems you prefer the separation of Analysis/processing, so alternative looks like:

Process:
- basic info about process (name/desc)
- class for DataTouchInfo
- class for codemanipulation info

Analysis:
- basic info about analysis (name/desc)
- class for DataTouchInfo
- class for codemanipulation info

And some file tracking DataManipulations (we should maybe just change to this name,,,, since DataTouch is so weird...) would be a series of nodes (Process + Analysis Steps) tracking the history.

src/aind_data_schema/datamanipulation.py
Sun-flow: im not sure I am following this comment

src/aind_data_schema/datamanipulation.py
Sun-flow: ive gone ahead and changed this -- it would be nice to go over what is being tracked in this, and discuss if analysis needs an analog for it

src/aind_data_schema/datamanipulation.py
Sun-flow: ah! youve addressed my prior question, lol. It can easily, i would like to discuss briefly what it is that this model is doing first.

src/aind_data_schema/datamanipulation.py
```diff
@@ -0,0 +1,78 @@
...
+    input_parameters: Dict[str, Any] = Field(..., title="Input parameters")
+    output_parameters: Optional[Dict[str, Any]] = Field(..., title="Output parameters")
+    outputs: Optional[Dict[str, Any]] = Field(None, description="Output parameters", title="Outputs")
+
+
+class AnalysisStep(AindModel):
+    """Description of a single processing step"""
+
+    analysis_name: str = Field(..., description="Name of the analysis method used", title="Analysis name")
+    analysis_type: AnalysisType = Field(..., description="Type of analysis performed on dataset", title="Analysis type")
```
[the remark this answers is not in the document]
saskiad: but we need to have overarching classes of analysis, and I'm not sure we have them...

src/aind_data_schema/datamanipulation.py
saskiad: sorry - I think I confused myself. Let me take another look.

src/aind_data_schema/datamanipulation.py
saskiad: the analysis_code - I think this needs to live in the DataTouch class because it's part of how the data was used.

src/aind_data_schema/datamanipulation.py
saskiad: Sorry, I realize I got confused - too many different things swimming in my head at once. We want processing and analysis together - but I think then I missed something. Because the DataTouch is still the step, and we need something that has the list of steps, for both processing and analysis. 
(and apologies if you jsut removed it in response to this)

src/aind_data_schema/datamanipulation.py
Sun-flow: Sometimes all of these classes are going to be necessary, but I think dividing them up based on topic is useful for organization, and lets us see exactly what we are using. Additionally if more fields need to be added, there's some clear organization for where they go based on what they are related to.

src/aind_data_schema/datamanipulation.py
Sun-flow: let me make a figma doc detailing this organization, and I can show it to you later in the week

src/aind_data_schema/datamanipulation.py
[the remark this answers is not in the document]
Sun-flow: If that's the case then that's fine, we won't need an enum. It will be easy to remove in favor of a string.

saskiad:
@Sun-flow can we delete this PR? We already closed this issue, correct? (I might be confusing things so correct me if I'm wrong)

saskiad:
@Sun-flow can you confirm that this PR should be closed and deleted?

PR #373, opened 2023-06-14
Doc 291 full manufacturer names

### LINKED ISSUE #291: Full names for manufacturers
There are several manufacturers listed only by their abbreviation. I'd like us to know what who these are. @dyf does it make sense to use the BaseName class here? My worry with that is that we don't have abbreviations for all manufacturers.

Specific examples:
OEPS = Open Ephys Production Site
IMEC
FLIR
LG
MPI

I think for some of these we want to use the abbreviation in most instances (eg. IMEC) but I also think we want to know what these all are.

Sun-flow:
closes #291 

Names expanded, some RORIDs input

### DIFF
--- a/src/aind_data_schema/device.py
+++ b/src/aind_data_schema/device.py
@@ -68,11 +68,31 @@ class Manufacturer(Enum):
         registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
         registry_identifier=""
     )
-    CHROMA = "Chroma"
-    COHERENT_SCIENTIFIC = "Coherent Scientific"
-    CUSTOM = "Custom"
-    DORIC = "Doric"
-    EALING = "Ealing"
+    CHROMA = PIDName(
+        name=BaseName(name="Chroma"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
+    )
+    COHERENT_SCIENTIFIC = PIDName(
+        name=BaseName(name="Coherent Scientific"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
+    )
+    CUSTOM = PIDName(
+        name=BaseName(name="Custom"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
+    )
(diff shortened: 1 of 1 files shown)


src/aind_data_schema/device.py
```diff
@@ -56,40 +57,156 @@ class DeviceDriver(Enum):
...
+        name=BaseName(name="Doric"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
+    )
+    EALING = PIDName(
+        name=BaseName(name="Ealing"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
+    )
     EDMUND_OPTICS = "Edmund Optics"
```
saskiad, line 105: So it looks like some manufacturers aren't being turned into PIDName and others are? Is that right? If so, why?
(and also if so, can we keep "Custom" as not being a PIDName)

src/aind_data_schema/device.py
```diff
@@ -56,40 +57,156 @@ class DeviceDriver(Enum):
...
+    )
+    LEICA = PIDName(
+        name=BaseName(name="Leica"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
+    )
+    LG = PIDName(
+        name=BaseName(name="LG"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
```
saskiad, line 130: We have the LG ror id in the spreadsheet

saskiad (changes requested):
There are some ROR ids that we have in the spreadsheet that aren't on here. Can we add those in?

src/aind_data_schema/device.py
saskiad: also IMEC.

src/aind_data_schema/device.py
Sun-flow: Basically, this PR was for expanding the names. I have a separate PR for the full implementation, bc I didn't originally have full names paired with ROR IDs. 

Here is the link to the PR that your comments here are asking for: https://github.com/AllenNeuralDynamics/aind-data-schema/pull/377

PR #374, opened 2023-06-14
Fix 372 write file path

### LINKED ISSUE #372: `write_standard_file()` does not handle paths including directories correctly
Example:

`processing_instance.write_standard_file("/results/")` saves `/results/_processing.json`

Expected behavior:

`processing_instance.write_standard_file("/results/")` saves `/results/processing.json`

Louise-X10:
close #372

### DIFF
--- a/src/aind_data_schema/base.py
+++ b/src/aind_data_schema/base.py
@@ -6,6 +6,7 @@ import os
 import re
 import urllib.parse
 from enum import EnumMeta
+from pathlib import Path
 from typing import Optional
 
 from pydantic import BaseModel, Extra, Field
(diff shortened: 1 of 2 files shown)


src/aind_data_schema/base.py
```diff
@@ -122,6 +122,8 @@ def write_standard_file(self, prefix=None):
         """
         if prefix is None:
             filename = self.default_filename()
+        elif str(prefix)[-1] == "/" or str(prefix)[-1] == "\\":
```
jtyoung84, line 125: I think it'll be better to add an Optional output directory such that the function signature looks like:
`def write_standard_file(self, output_directory: Optional[Path] = None, prefix=None)`

jtyoung84 (changes requested):
See comments

src/aind_data_schema/base.py
Louise-X10: Do we want to continue using Path or usual strings for directories? The ticket seems to show that they should be strings, but the unit test uses Path. The issue with Path is that they strip off the trailing slashes, so you would have to add them back manually. It may be easier if we could assume the directory argument is a string, and convert Path to strings in the unit tests.

jtyoung84: Assume the input directory is a Path object. The "prefix" is supposed to a string

jtyoung84 (commented):
Looks good to me. Double-check with @dyf that this is the desired behavior.

jtyoung84 (approved):

PR #377, opened 2023-06-20
Doc 343 manufacturer ror i ds

### LINKED ISSUE #343: Add RORIDs to Manufacturers
- [ ] get list of RORIDs available for our current list of Manufacturers
- [ ] Add RORID to device schema with Manufacturer (similar to Institution)

Sun-flow:
closes #343 

Expanding all Manufacturer names to implement BaseName/PIDName, and adding ROR IDs that we know of.


src/aind_data_schema/device.py
```diff
@@ -48,40 +48,173 @@ class PowerUnit(Enum):
...
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
+    )
+    MKS_NEWPORT = PIDName(
+        name=BaseName(name="MKS Newport"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier="00k17f049"
+    )
+    MPI = PIDName(
+        name=BaseName(name="Materials Processing, Inc", abbreviation="MPI"),
```
saskiad, line 143: I'm not actually sure that's the name, let's leave it as MPI

src/aind_data_schema/device.py
```diff
@@ -48,40 +48,173 @@ class PowerUnit(Enum):
...
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier="01j1gwp17"
```
saskiad, line 145: this is the ID for FLIR.

src/aind_data_schema/device.py
```diff
@@ -48,40 +48,173 @@ class PowerUnit(Enum):
...
+        name=BaseName(name="Chroma"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
+    )
+    COHERENT_SCIENTIFIC = PIDName(
+        name=BaseName(name="Coherent Scientific"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier="031tysd23"
+    )
+    CUSTOM = PIDName(
```
saskiad, line 81: can we have CUSTOM not be a PIDName? such as how we do OTHER? Because Custom is not a company, it means the device was custom built.

src/aind_data_schema/device.py
```diff
@@ -48,40 +48,173 @@ class PowerUnit(Enum):
...
+    )
+    SEMROCK = PIDName(
+        name=BaseName(name="Semrock"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
+    )
+    THORLABS = PIDName(
+        name=BaseName(name="Thorlabs"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier=""
```
saskiad, line 201: Thorlabs rorid = 04gsnvb07

src/aind_data_schema/device.py
```diff
@@ -56,40 +58,169 @@ class DeviceDriver(Enum):
...
+    OEPS = PIDName(
+        name=BaseName(name="Open Ephys Production Site", abbreviation="OEPS"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier="007rkz355"
+    )
+    OLYMPUS = PIDName(
+        name=BaseName(name="Olympus"),
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier="02vcdte90
+"
```
saskiad, line 177: Fix this end quote - looks like it got bumped to the next line. I think this is causing the test/lint errors

saskiad (changes requested):
fix the end quote for Olympus id

src/aind_data_schema/base.py
```diff
@@ -65,7 +65,7 @@ class PIDName(BaseName):
     Model for associate a name with a persistent identifier (PID),
     the registry for that PID, and abbreviation for that registry
     """
-
+    name: str = Field(..., title="name")
```
jtyoung84, line 68: This field get's inherited from BaseName

src/aind_data_schema/device.py
```diff
@@ -56,41 +56,173 @@ class DeviceDriver(Enum):
...
+        name="Vieworks",
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier="",
+    )
+    VORTRAN = PIDName(
+        name="Vortran",
+        registry=BaseName(name="Research Organization Registry", abbreviation="ROR"),
+        registry_identifier="",
+    )
+    OTHER = "OTHER"
```
jtyoung84, line 225: OTHER and CUSTOM also need to be a PIDName

Sun-flow: @saskiad requested that this not be a PIDName, as it will never have an RORID. I don't mind to change it, either way is fine for me.

saskiad (approved):

Issue #272, opened 2023-04-27
imaging-calibration


miketaormina:
**Issue**
Certain aspects of data acquisition do not neatly fall into the `instrument` or `acquisition` categories, but are needed to apply to one or more acquisition instances and are likely to change frequently. Two examples on the SmartSPIM are laser power and flat-field correction images, the first needing a measured correspondence between "percent on" and "milli Watts at objective" and the second needing one or more 2D images be acquired to represent the DC response of the instrument.

These two examples are the ones that will immediately be solved by this feature.

**Proposed Solution**
I would like to suggest a new metadata schema to track either this information directly (in the first example's case) or a way to locate/compute this information (in the second example's case, a file path or the amplitude/phase of a cosine approximation for flat field). I envision other experimental modalities requiring similar parameters be tracked for periodic instrument calibration.

**Describe alternatives you've considered**
It may be possible to shoe-horn these into an acquisition or instrument instance, but a separate schema makes sense for a few reasons:
1. Calibration data can be generated too frequently to warrant inclusion in `instrument.json`.
2. Even if done once per day, a single instrument calibration may apply to several acquisitions.
3. Acquisition metadata can be generated from instrument output, while calibration data is inherently user-generated.

saskiad:
I'm a bit reluctant to add a new schema because I think it doesn't quite solve the issue but just punts it. But I'm a little bit confused. Can you unpack a bit how often you expect the laser power to be calibrated and how often the flat image is acquired. 
Something that is calibrated once a day can be referenced in the acquisitions schema. And multiple acquisitions can reference the same calibration data.
Something that is calibrated less frequently should probably live in the instrument - and the instrument can be updated with each calibration (incrementing the version) - even if that's the only thing changing.
There's probably some threshold of update frequency at which we put it in the instrument vs the acquisition, but I think we can figure something out for that.

Is 100% of the acquisition metadata is instrument generated? I think we can work on a system to make sure that calibration data can be added to that the json.

saskiad:
That sounds good! Thanks Mike!

miketaormina:
Finally got around to roughing this out in `acquisition.py.`
# Summary
I assumed a linear relationship between a laser set point and the power measured at the sample, but I guess this accommodates multiple entries, so is generalizable if people get crazy with it.

Added classes:
1. `PowerUnit(Enum)`
[...]

PR #379, opened 2023-06-21
Feat imaging calibration

miketaormina:
# Purpose
To accommodate a calibration of the type: `(excitation power set point, excitation power measured at objective)`.
Set point allows `percent` as a unit for the set point, which is what the SmartSPIM logs in it's instrument-created metadata.
There is a bit more discussion in [Issue 272](https://github.com/AllenNeuralDynamics/aind-data-schema/issues/272), but generally speaking, the power set point on a laser is very different from the power that makes it to a sample, and can vary across instruments and time.

## Scope
Added to `acquisition.py` schema, added example of a stand-alone calibration measurement. Schema additions are `Optional`, so should not create conflicts in main.

Edit -- Unit tests failing for examples, even though this field is designated `Optional`. Am I misunderstanding unit tests?


src/aind_data_schema/imaging/acquisition.py
```diff
@@ -102,6 +103,35 @@ class ProcessingSteps(AindModel):
...
+
+
+class LightSourcePowerCalibration(AindModel):
+    """Calibration for laser set point/power at sample relationshipe"""
+
+    date_of_calibration: datetime = Field(
+        ...,
+        title="Date and time of calibration"
+    )
+    light_source: Lightsource = Field(..., title="Light Source")
```
saskiad, line 121: can light_source be light_source_name? That way don't have to include all the specs of the device that are in the rig but have the name to link between them?

saskiad (changes requested):
Rather than include the full light_source device, it would be cleaner to use the name of the light_source to link between the device and the calibration

src/aind_data_schema/imaging/acquisition.py
dyf: +1

src/aind_data_schema/imaging/acquisition.py
```diff
@@ -102,6 +103,35 @@ class ProcessingSteps(AindModel):
     ]
 
 
+class PowerUnit(Enum):
```
dyf, line 106: please move this to `utils/units.py`

dyf (changes requested):

src/aind_data_schema/imaging/acquisition.py
miketaormina: I agree as well. I think this was my initial idea, but the light sources didn't have names. Looking at the instrument configurations, this is only because `name` is an optional field that wasn't filled in, so I will need to also alter the instrument JSON files for this to work. I think I will use the convention `Ex_<Lightsource.wavelength>`, since that is how the instrument names acquisition files.

camilolaiton:
Hello @miketaormina, I have updated your branch to use a generic calibration class I included in this [branch](https://github.com/AllenNeuralDynamics/aind-data-schema/blob/feat-imaging-generic-calibration/src/aind_data_schema/device.py#L620).

I noticed that the [LightSourcePowerCalibration](https://github.com/AllenNeuralDynamics/aind-data-schema/blob/feat-imaging-calibration/src/aind_data_schema/imaging/acquisition.py#L111) class is in the `acquisition.py` script. Should we move that one as well to `device.py`? If so, there are circular dependencies that we will have to fix @dyf.

saskiad:
Ideally LightSourcePowerCalibration will use the generic calibration class. I imagine this should stay in the acquisition file - @miketaormina this is likely something you want to run before each imaging session? If so, acquisition file is the place that makes most sense.

miketaormina:
That's right, @saskiad . My thinking was that it would happen at most once per day, but maybe even once per week. In either case, one calibration would be associated with several acquisitions. We may need to do this once per imaging day until we know how stable the values are on that time scale.

How I imagined this working would be generating calibration files in a standard location on the acquisition machine and then ingesting the most recent one when the data sets are prepared for transfer. This is similar to how I have been grabbing the instrument json, but I think it belongs in the acquisition file, since it would be used to establish the actual laser power in milliwatts for that particular imaging run (since the device currently only reports this value as a percentage of maximum).

I just need a day or so to sort out what's needed on the updated branch.

src/aind_data_schema/imaging/acquisition.py
miketaormina: added PowerUnit and PowerValue to `utils/units.py`

dyf (approved):
Looks fine.

src/aind_data_schema/device.py
```diff
@@ -596,3 +604,30 @@ class VisualStimulusDisplayAssembly(AindModel):
 
     # optional fields
     position: Optional[RelativePosition] = Field(None, title="Relative position of the monitor")
+
+
+class CalibrationType(Enum):
+    """Calibration types"""
+
+    LASER = "laser"
```
saskiad, line 612: please use sentence case: LASER = "Laser

src/aind_data_schema/device.py
```diff
@@ -596,3 +604,30 @@ class VisualStimulusDisplayAssembly(AindModel):
...
+    MONITOR = "monitor"
+    SPEAKER = "speaker"
+    WATER_V = "water valve"
+    CHAMBER = "chamber"
+    OTHER = "other"
+
+
+class Calibration(AindModel):
```
saskiad, line 620: This class need the device name that the calibration is for. This is a key field

examples/smartspim_power_calibrations.json
```diff
@@ -0,0 +1,44 @@
+[
+    {
+       "date_of_calibration": "2023-06-20T14:04:05.770680",
+       "light_source": {
+          "name": null,
```
saskiad, line 5: please add names to your devices so that we can actually link to them from the acquisition file

src/aind_data_schema/imaging/acquisition.py
```diff
@@ -98,10 +99,22 @@ class ProcessingSteps(AindModel):
     ]
 
 
+class LightSourcePowerCalibration(Calibration):
```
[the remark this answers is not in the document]
miketaormina: I guess it's only because there wasn't a generic class when I started. I don't think there is a reason that it couldn't just be an instance of `Calibration`,  are the dictionaries on this unrestricted (the example json would be the only place where it is defined)?

examples/smartspim_power_calibrations.json
miketaormina: I guess I forgot about this file. I've given both aibs and aind light source devices name in the instrument json and will add one here. I can also make a script that generates this example file, to match the rest of the examples.

saskiad:
sorry to add a monkey wrench at this point, but I think that cleaning a chamber isn't a calibration event so much as a maintenance event.

miketaormina:
>I feel like we aren't aligned on how we want to use the calibration class

@saskiad I agree and think it takes some more thought. Although a calibration should be associated with an instrument and one or more acquisitions, it doesn't seem to fit neatly into one or the other schemas. It often will be generated separately from an acquisition, so if it goes into `acquisition.json`, it needs to be ingested, which I think is a pattern we are trying to avoid in this workflow (correct me if I'm wrong). If it goes into `instrument.json`, that file will be updated much more often than I think we'd like. Does this mean we need an `instrument_maintenance.json` with calibrations and cleanings? I suspect that we don't like that either.

Can we:
Make `class CalibrationSet(AindCoreModel)` be a list of `Calibration` of the same type and use a convention of including the `CalibrationType` in the file name? 
E.g., 
```
class CalibrationSet(AindCoreModel):
    calibration_type: ...
    instrument_id: ...
    calibrations: List(Calibration)

c = CalibrationSet(. . . , calibrations=[cal0, cal1, cal2])
c.write_standard_file(prefix=f"{c.calibration_type}") 

```
outputs `laser_calibration.json` or `monitor_calibration.json`, which consist of lists of calibrations of the same type (and performed at the same time)? Then raw data gets packaged with the most recent version of locally available calibrations. This gets around the acquisition vs. device/instrument but spreads the information over more files.

Am I overcomplicating this?

saskiad:
I think you are overcomplicating it. I imaging in the acquisition schema there is a list of calibrations. The list uses a class called Calibration which is being added to the device schema in this PR. That class needs a field for the name of the device that was calibrated. Date of calibration. A dictionary of input variables (e.g. laser power settings) and a dictionary of output variables (e.g. measured power). 
I think making separate classes for each device calibration is going to get very bloated very fast. Already for the behavior schema we have about 5 devices that need regular calibration. 
So then in the acquisition schema you create a list of all the calibrations for the instrument that you need.

Separate from this I suggest we create a class called Maintenance which is for things like cleaning the chamber where it isn't a calibration per se and doesn't have the input/output dictionaries.

camilolaiton:
Hello! I just wanted to follow up on this @dyf @saskiad @miketaormina, I would like to have these changes in the main branch to update the SmartSPIM data transfer scripts. Thanks!

saskiad:
do you want me to take a pass at how I would structure this? I can probably do that tonight. That okay @miketaormina ?

miketaormina:
> do you want me to take a pass at how I would structure this? I can probably do that tonight. That okay @miketaormina ?

Okay with me, for sure!

src/aind_data_schema/device.py
```diff
@@ -260,6 +261,28 @@ class Software(AindModel):
...
+
+
+class Maintenance(AindModel):
+    """Generic maintenance class"""
+
+    date_of_maintenance: datetime = Field(..., title="Date and time of maintenance")
+    device_name: str = Field(..., title="Device name", description="Must match a device name in rig/instrument")
+    description: str = Field(..., title="Description", description="Description on maintenance procedure")
+    protocol_id: Optional[str] = Field(None, title="Protocol ID")
+    reagents: Optional[List[Reagent]] = Field(None, title="Reagents")
```
dyf, line 282: Does `Maintenance` need reagents?

dyf (commented):
Quick question re: Maintenance.

src/aind_data_schema/device.py
saskiad: Does it require reagents? I expect not - there can be maintenance that doesn't have reagents. But it sounds like there are some that do - eg. cleaning chambers.

examples/exaspim_acquisition.py
```diff
@@ -13,14 +13,23 @@
...
         Calibration(
             date_of_calibration=t,
-            calibration_type=CalibrationType.CHAMBER,
-            reagents=[
-                Reagent(name="reagent1", source="xxx", rrid="xxx", lot_number="xxx", expiration_date=t),
-            ],
-            notes="Some notes",  # Notes are optional
+            device_name="Laser_1",
+            description="Laser power calibration",
+            input={"power setting": 1},
```
miketaormina, line 31: Does the Calibration object here need to use `PowerValue`, as in `tests/test_imaging.py` below? Or is the schema built such that any valid dictionary is okay? Asking becuase `PowerValue` would require units on these numbers.

Not asking about schema validation, but rather if the example should indicate what is expected (not necessarily required).

miketaormina (commented):
This looks good to me, if you need me to update one of the smartspim examples to include calibrations and/or maintenance, let me know.

examples/exaspim_acquisition.py
saskiad: dictionary can take anything, but agree that example should demonstrate how we want to use it and I really like having units. I can update this

saskiad (approved):

PR #388, opened 2023-06-29
makes manufacturer basenameenummeta

### LINKED ISSUE #387: Manufacturer should be BaseNameEnumMeta
**Describe the bug**
Manufacturer on the metadata entry app is rendering as [object, object] because it does not have enumNames. This can be easily fixed by adding metaclass=BaseNameEnumMeta to the Manufacturer definition

**Expected behavior**
It should render the list of manufacturer names from PIDName

**Screenshots**
Screenshot 2023-06-29 at 13 40 11

**Smartphone (please complete the following information):**

**Additional context**

mekhlakapoor:
closes #387 
quick fix to make enumnames for manufacturer

### DIFF
--- a/src/aind_data_schema/device.py
+++ b/src/aind_data_schema/device.py
@@ -11,7 +11,7 @@ except ImportError:  # pragma: no cover
 
 from pydantic import Field
 
-from .base import AindModel, BaseName, PIDName, BaseNameEnumMeta
+from aind_data_schema.base import AindModel, BaseName, PIDName, BaseNameEnumMeta
 
 
 class SizeUnit(Enum):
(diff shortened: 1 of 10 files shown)


src/aind_data_schema/device.py
```diff
@@ -11,7 +11,7 @@
 
 from pydantic import Field
 
-from .base import AindModel, BaseName, PIDName
+from .base import AindModel, BaseName, PIDName, BaseNameEnumMeta
```
jtyoung84, line 14: There was a PR to remove relative imports a while back. It looks like this one got missed. It'd be cool if you can update the import in this pr since you're already making a change

jtyoung84 (approved):

PR #395, opened 2023-07-01
Bug 389 optional fields

### LINKED ISSUE #389: Ephys rig camera assembly require filter which is optional
UPDATE: 
The form is rendered directly from schemas, and is therefore requiring nested fields, even though the main field is optional. This is because even though the main field may be optional, the nested fields are defined as required. 

**Describe the bug**
The filter is optional in the schema but the entry form requires it.

**Additional context**
Relaying info from Anna Lakunina

mekhlakapoor:
closes AllenNeuralDynamics/aind-metadata-entry-js#93 

Converts some of those optional fields with required sub-fields to Lists, that way users can add it if they'd like. If a user decides to add (ex) Filter, only then will they be required to submit the fields required for it. This ensures that we still get complete metadata if optional fields are selected. This is an example of how it looks: 

Screen Shot 2023-06-30 at 7 45 22 PM
Screen Shot 2023-06-30 at 7 45 30 PM

### DIFF
--- a/examples/ephys_rig.json
+++ b/examples/ephys_rig.json
@@ -1,6 +1,6 @@
 {
    "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/ephys/ephys_rig.py",
-   "schema_version": "0.7.1",
+   "schema_version": "0.7.2",
    "rig_id": "323_EPHYS1",
    "ephys_assemblies": [
       {
@@ -85,7 +85,7 @@
             "chroma": "Color",
             "sensor_format": "1/2.9\"",
             "format_unit": null,
-            "recording_software": null,
+            "recording_software": [],
             "driver": null,
             "driver_version": null
          },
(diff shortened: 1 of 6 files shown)


saskiad (commented):
I think I'd like to add a max_items to these particular lists because there shouldn't be multiple items in them and I'm worried having them a list could lead to confusion... Do you know if that would work with the gui?

mekhlakapoor:
> I think I'd like to add a max_items to these particular lists because there shouldn't be multiple items in them and I'm worried having them a list could lead to confusion... Do you know if that would work with the gui?

There probably is. I'm going to pivot this back to an issue on aind-metadata-entry app though. looks like there was an update last week to rjsf that could help https://github.com/rjsf-team/react-jsonschema-form/issues/3617

mekhlakapoor:
closing this because we are handling this issue on aind-metadata-entry-js instead https://github.com/AllenNeuralDynamics/aind-metadata-entry-js/issues/93

PR #400, opened 2023-07-06
Feat: added from_*() constructors for DerivedDataDescription

### LINKED ISSUE #399: Add utils and tests for schema operations
There are no utility functions for schema-related operations, such as constructing `DerivedDataDescription` from `DataDescription`

Add some utility functions and tests for schema-related operations, such as constructing `DerivedDataDescription` from `DataDescription`

Adding such functions elsewhere, but this repo seems more appropriate.

**Additional context**
See comment here: https://github.com/AllenNeuralDynamics/aind-capsule-ephys-spikesort-kilosort25-full/pull/5#discussion_r1251125381

alejoe91:
fixes #399


pyproject.toml
```diff
@@ -16,6 +16,7 @@ dynamic = ["version"]
 dependencies = [
     'pydantic<2.0',
     'jsonschema',
+    'packaging',
```
jtyoung84, line 19: We can probably avoid adding another package to the dependencies list. I think you can parse version numbers using the built-in:
```
>>> from pkg_resources import parse_version
>>> parse_version('0.13.0') < parse_version('1.0.0')
True
>>> parse_version('1.13.0') < parse_version('1.0.0')
False
```

tests/resources/ephys_data_description/data_description_0.6.2.json
```diff
@@ -0,0 +1,41 @@
...
+    "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/data_description.py",
+    "schema_version": "0.6.2",
+    "license": "CC-BY-4.0",
+    "funding_source": [
+        {
+            "funder": "Allen Insitute for Neural Dynamics"
+        }
+    ],
+    "investigators": [
+        "Galen Lynch",
```
jtyoung84, line 11: We've been making an effort to scrub out names and using stuff "John Doe" or "Mary Smith"

src/aind_data_schema/utils/aind_utils.py
```diff
@@ -34,3 +59,110 @@ def aind_core_models() -> Iterator[Type[AindCoreModel]]:
     """
     for model in AindCoreModel.__subclasses__():
         yield model
+
+
+def create_derived_data_description(
```
jtyoung84, line 64: It'd be better if this was moved in the DerivedDataDescription class as an alternate class constructor instead of a method in the aind_utils package

src/aind_data_schema/utils/aind_utils.py
```diff
@@ -34,3 +59,110 @@ def aind_core_models() -> Iterator[Type[AindCoreModel]]:
...
+    process_name: str,
+    existing_data_description: str | Path | DataDescription | None = None,
+    input_data_name: str | None = None,
```
jtyoung84, line 67: You can import Optional and List from the typing module. This then becomes `input_data_name: Optional[str] = None`

src/aind_data_schema/utils/aind_utils.py
```diff
@@ -34,3 +59,110 @@ def aind_core_models() -> Iterator[Type[AindCoreModel]]:
...
+    subject_id: str = None,
```
jtyoung84, line 68: This should be `Optional[str]`

src/aind_data_schema/utils/aind_utils.py
```diff
@@ -34,3 +59,110 @@ def aind_core_models() -> Iterator[Type[AindCoreModel]]:
...
+    institution: Institution = Institution.AIND,
+    funding_source: list[Funding] = [Funding(funder="AIND")],
+    investigators: list = [],
```
jtyoung84, line 71: It isn't recommended to use mutable data types as defaults. One way to circumvent the warnings is to do something like:
```
investigators: Optional[List] = None
```
and then in the code:
```
investigators = [] if investigators is None else investigators
```

src/aind_data_schema/utils/aind_utils.py
```diff
@@ -34,3 +59,110 @@ def aind_core_models() -> Iterator[Type[AindCoreModel]]:
     """
     for model in AindCoreModel.__subclasses__():
         yield model
+
+
+def create_derived_data_description(
+    process_name: str,
+    existing_data_description: str | Path | DataDescription | None = None,
```
jtyoung84, line 66: I'm not a huge fan of mixing data types like this in function args. You may want to consider creating different class constructors on the DerivedDataDescription class: from_data_description and from_data_description_path

tests/test_utils.py
```diff
@@ -23,6 +25,46 @@ def test_get_classes(self):
         """Tests get classes method"""
         self.assertEqual(list(get_classes()), list(get_classes(__name__)))
 
+    def test_create_derived_data_description(self):
+        """Tests create_derived_data_description method"""
+        from aind_data_schema.data_description import DataDescription, DerivedDataDescription, ExperimentType, Modality
```
jtyoung84, line 30: It should be fine to add these to the imports at the top

jtyoung84 (changes requested):
Left a few notes. I'd recommend moving the code into the DerivedDataDescription class instead of putting it as a utils method.

tests/test_utils.py
```diff
@@ -23,6 +25,46 @@ def test_get_classes(self):
...
+            derived_data_description_from_file = create_derived_data_description(
+                process_name="test_process",
+                existing_data_description=data_description_arg,
+                experiment_type=experiment_type,
+            )
+            with open(data_description_arg, "r") as f:
+                data_description_dict = json.load(f)
+            if "experiment_type" not in data_description_dict:
+                data_description_dict["experiment_type"] = ExperimentType.ECEPHYS
+            data_description = DataDescription.construct(**data_description_dict)
```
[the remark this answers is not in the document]
alejoe91: Parse obj will fail for old schema versions! So this is on purpose and mismatches are handled later. That's why I added old files to test with.

src/aind_data_schema/utils/aind_utils.py
alejoe91: done

src/aind_data_schema/utils/aind_utils.py
alejoe91: I made two separate args ;)

pyproject.toml
alejoe91: done

alejoe91:
@jtyoung84 ready to review ;)

src/aind_data_schema/utils/aind_utils.py
alejoe91: done, split in `from_data_description`, `rom_data_description_file`, `from_scratch`

src/aind_data_schema/data_description.py
```diff
@@ -326,6 +340,189 @@ def parse_name(cls, name):
             input_data_name=m.group("input"),
         )
 
+    @staticmethod
```
jtyoung84, line 343: I think it might be better to make this a classmethod instead of a staticmethod. The return would be `cls(name=..., subject_id=..., etc.)`

src/aind_data_schema/data_description.py
```diff
@@ -326,6 +340,189 @@ def parse_name(cls, name):
...
+            "ror_id",
+            "creation_time",
+            "creation_date",
+        ]
+        for key in skip_existing_data_description_keys:
+            if key in existing_data_description_dict:
+                del existing_data_description_dict[key]
+        derived_data_description_dict.update(existing_data_description_dict)
+        derived_data_description_dict["input_data_name"] = existing_data_description_dict["name"]
+        derived_data_description = DerivedDataDescription(process_name=process_name, **derived_data_description_dict)
```
jtyoung84, line 422: I don't think process_name is a field or property of DerivedDataDescription

src/aind_data_schema/data_description.py
```diff
@@ -326,6 +340,189 @@ def parse_name(cls, name):
...
+            data_description=data_description,
+            process_name=process_name,
+            experiment_type=experiment_type,
+            institution=institution,
+            funding_source=funding_source,
+            investigators=investigators,
+        )
+
+    @staticmethod
+    def from_scratch(
```
jtyoung84, line 477: I'm not sure how this is different than the standard class constructor

src/aind_data_schema/data_description.py
```diff
@@ -326,6 +340,189 @@ def parse_name(cls, name):
...
+
+        Returns
+        -------
+        DerivedDataDescription
+            The derived data description
+        """
+        derived_data_description_dict = create_base_data_description_dict(
+            investigators=investigators, institution=institution, funding_source=funding_source
+        )
+        existing_data_description_dict = data_description.dict()
```
jtyoung84, line 384: It might be less complicated to use the DataDescription attributes directly instead of converting everything to a dictionary

src/aind_data_schema/data_description.py
```diff
@@ -301,6 +305,16 @@ class DerivedDataDescription(DataDescription):
         title="Data Level",
         const=True,
     )
+    # Utilities
+    skip_existing_data_description_keys = [
```
jtyoung84, line 309: This is duplicated on line 408

src/aind_data_schema/data_description.py
```diff
@@ -326,6 +340,189 @@ def parse_name(cls, name):
...
+            if key in existing_data_description_dict:
+                del existing_data_description_dict[key]
+        derived_data_description_dict.update(existing_data_description_dict)
+        derived_data_description_dict["input_data_name"] = existing_data_description_dict["name"]
+        derived_data_description = DerivedDataDescription(process_name=process_name, **derived_data_description_dict)
+
+        return derived_data_description
+
+    @staticmethod
+    def from_data_description_file(
```
jtyoung84, line 427: I think it should be simple enough for a user to open a file and parse it into a model. I'm not sure if we want to add a method in this repo that does that. If so, we should create a generic method in the base class.

src/aind_data_schema/data_description.py
```diff
@@ -366,3 +563,38 @@ def parse_name(cls, name):
...
+
+    Returns
+    -------
+    dict
+        The base data description dictionary
+    """
+    now = datetime.now()
+    # make base dictionary form scratch
+    base_data_description_dict = {}
+    base_data_description_dict["creation_time"] = now.time()
```
jtyoung84, line 593: I feel like we should just set defaults in the pydantic model rather than creating a method in the python module

src/aind_data_schema/data_description.py
alejoe91: It's needed by the constructor to make the label: https://github.com/AllenNeuralDynamics/aind-data-schema/blob/main/src/aind_data_schema/data_description.py#L305-L308

src/aind_data_schema/data_description.py
alejoe91: good point

src/aind_data_schema/data_description.py
alejoe91: not sure how one would set the `datetime.now()` as default in the pydantic model...

src/aind_data_schema/data_description.py
alejoe91: We need the dict later to skip some keys. Not sure how to do it with the DataDescription object

src/aind_data_schema/data_description.py
alejoe91: removed

src/aind_data_schema/data_description.py
alejoe91: removed

jtyoung84:
@dyf I set the `investigators` field to default to an empty list. Is that cool with you?
@alejoe91 This tries to parse info from a DataDescription model. I didn't want to override the defaults with AIND specific stuff though. It's possible to explicitly set fields that you don't want to pull though.

alejoe91:
@jtyoung84 looking great! Thanks for working on it and extending the tests :)

dyf (changes requested):
Can we define these functions elsewhere? Say, some sort of legacy.py module? We can still import them to be validators here, but they make the schema hard to read.

jtyoung84 (approved):

PR #401, opened 2023-07-06
Bug 387 manufacturer

### LINKED ISSUE #387: Manufacturer should be BaseNameEnumMeta

mekhlakapoor:
closes #387 

Adding Manufacturer enumNames was not enough to fix manufacturer dropdowns. This PR removes the subsets for manufacturer in devices (defined as Literals) so that each device has the full list of manufacturers.


src/aind_data_schema/device.py
```diff
@@ -336,7 +323,7 @@ class Lens(Device):
     """Lens used to focus light onto a camera sensor"""
 
     # required fields
-    manufacturer: Literal[Manufacturer.EDMUND_OPTICS, Manufacturer.THORLABS, Manufacturer.OTHER]
+    manufacturer: Manufacturer = Field(..., title="Manufacturer")
```
jtyoung84, line 326: This won't restrict the choices to three that are listed. There's a couple of ways to do this. You can either create a new LensManufacturer enum with or there's probably a way to make this a Field and add some additional metadata in the Field object to help render it.

mekhlakapoor: I already talked to @Saskia about this -- we didn't want to make separate enums for each subset. Also, I tried exploring additional metadata to help render it for the subsets but it was proving difficult. In the end we decided for now to let it be the full Manufacturer list

saskiad: We want a single list of manufacturers, so having separate LensManufacturers and CameraManufacturers (etc) isn't going to work. We do want to find a solution to have subsets to make it easier for users, but so far haven't landed on somethign that works.

saskiad (approved):

src/aind_data_schema/device.py
mekhlakapoor: I am closing this PR. We decided to handle it through #407

PR #425, opened 2023-07-22
Feat behavior

### LINKED ISSUE #352: update behavior rig devices
info from Linda about components

saskiad:
closes #352


src/aind_data_schema/device.py
```diff
@@ -564,17 +571,20 @@ class Treadmill(MousePlatform):
...
+    """Description of visual display for visual stimuli"""
 
     # required fields
+    stimulus_type: str = Field("Visual monitor", title="Stimulus type")
     manufacturer: EnumSubset[Manufacturer.LG]
     refresh_rate: int = Field(..., title="Refresh rate (Hz)", units="Hz", ge=60)
     width: int = Field(..., title="Width (pixels)", units="pixels")
     height: int = Field(..., title="Height (pixels)", units="pixels")
     size_unit: SizeUnit = Field(SizeUnit.PX, title="Size unit")
     viewing_distance: Decimal = Field(..., title="Viewing distance (cm)", units="cm")
```
dyf, line 583: replace w unit+value?

src/aind_data_schema/device.py
```diff
@@ -590,37 +600,44 @@ class Monitor(Device):
...
-    track_wheel: Union[Tube, Treadmill, Disc] = Field(..., title="Track wheel type")
-
-    # optional fields
-    stage_software: Optional[Software] = Field(None, title="Stage software")
-    water_delivery: Optional[WaterDelivery] = Field(None, title="Water delivery")
+class WaterSpout(Device):
+    """Description of a water spout"""
 
+    side: SpoutSide = Field(..., title="Spout side")
+    spout_diameter: float = Field(..., title="Spout diameter (mm)")
```
dyf, line 614: unit+value

src/aind_data_schema/device.py
```diff
@@ -590,37 +600,44 @@ class Monitor(Device):
...
 
-    # required fields
-    monitor: Monitor = Field(..., title="Monitor")
-    viewing_distance: Decimal = Field(..., title="Viewing distance (cm)", units="cm")
-    viewing_distance_unit: SizeUnit = Field(SizeUnit.CM, title="Viewing distance unit")
+class WaterDelivery(AindModel):
+    """Description of water delivery system"""
 
-    # optional fields
+    stimulus_type: str = Field("Water delivery", title="Stimulus type")
```
dyf, line 624: should this be a constant?

src/aind_data_schema/device.py
```diff
@@ -590,37 +600,44 @@ class Monitor(Device):
...
+    water_spouts: List[WaterSpout] = Field(..., title="Water spouts")
+    lick_holder: Device = Field(..., title="Lick holder")
+
+
+class Speaker(Device):
+    """Description of a speaker for auditory stimuli"""
+
+    stimulus_type: str = Field("Speaker", title="Stimulus type")
+    manufacturer: EnumSubset[Manufacturer.TYMPHANY]
+    minimum_impedance: float = Field(..., title="Minimum impedance (ohms)", units="ohms")
```
dyf, line 635: unit+value for `minimum_impedance`, `rated_noise_power`, `sensitivity`, and `test_spectrum_bandwidth`

dyf (changes requested):
Some small things.  Does this affect Mekhla's FIP ophys metadata converter?

src/aind_data_schema/device.py
saskiad: until that's merged, I can't use it

src/aind_data_schema/device.py
saskiad: same

src/aind_data_schema/device.py
saskiad: yes. Since the stimulus in the session is a Union, we need something to indicate what the types of instances are.

src/aind_data_schema/device.py
saskiad: same as above

saskiad:
> Some small things. Does this affect Mekhla's FIP ophys metadata converter?

How would I know if it affects it or not?

mekhlakapoor:
> Some small things. Does this affect Mekhla's FIP ophys metadata converter?

It should be an easy fix-- i just need to change StimulusPresentation to StimulusEpoch.

dyf:
@saskiad add `const=True` to the `Field(...)` declarations for `stimulus_type` and this is gtg.

saskiad:
> @saskiad add `const=True` to the `Field(...)` declarations for `stimulus_type` and this is gtg.

I'll do that. I have more things to change based on feedback from behavior people so they'll be new commits before I finish the PR. I'll make it a draft.

src/aind_data_schema/device.py
saskiad: Oh, it's possible we want a different name for this. This is in the rig, saying what kind of stimulus device it is. So maybe we call it something like "stimulus device" instead of stimulus type because that will be overloaded.

src/aind_data_schema/device.py
saskiad: we've removed these fields

dyf (changes requested):
let's pivot back to paired `x` and `x_unit` fields for now, since unit with value is paused.

PR #427, opened 2023-07-24
adding new ophys modalities

### LINKED ISSUE #426: add ophys modalities and experiment types
- `single-plane-ophys`
- `multi-plane-ophys`

dyf:
resolves #426

### DIFF
--- a/src/aind_data_schema/data_description.py
+++ b/src/aind_data_schema/data_description.py
@@ -187,7 +187,7 @@ class RelatedData(AindModel):
 class DataDescription(AindCoreModel):
     """Description of a logical collection of data files"""
 
-    schema_version: str = Field("0.7.3", title="Schema Version", const=True)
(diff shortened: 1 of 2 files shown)


src/aind_data_schema/data_description.py
```diff
@@ -121,6 +121,8 @@ class Modality(Enum, metaclass=BaseNameEnumMeta):
     MRI = BaseName(name="Magnetic resonance imaging", abbreviation="MRI")
     OPHYS = BaseName(name="Optical physiology", abbreviation="ophys")
     POPHYS = BaseName(name="Planar optical physiology", abbreviation="pophys")
```
saskiad, line 123: You are keeping POPHYS in Modality and not in ExperimentType. What is that? I assume because we've had POPHYS data already? But why can we remove it from one and not the other?

saskiad (changes requested):
question more than a change

saskiad (approved):

PR #432, opened 2023-07-28
make injection dv list

saskiad:


saskiad:
closes #432

saskiad:
we'll need a validator to assure that the length of DV coordinates and the length of injection volume match

saskiad:
This will also impact how metadata service gets this info (e.g. the volume is one volume that is split between the two locations, so that math will need to be done)

saskiad:
I don't understand this linter error. @jtyoung84 do you know what this is?

dyf:
@saskiad fixed. @jtyoung84 looks like flake8 prefers to use `is` and `is not` for type comparisons now.

dyf (approved):

PR #436, opened 2023-08-04
Adding generic calibration class for acquisition

camilolaiton:
Adding a generic calibration class for acquisition. I am adding this since John needs to track other metadata related to the microscope maintenance.

I included:

- Date of calibration
- Description
- Notes

I believe @miketaormina can use inherit this class in the `LightSourcePowerCalibration` here: https://github.com/AllenNeuralDynamics/aind-data-schema/pull/379

### DIFF
--- a/examples/exaspim_acquisition.json
+++ b/examples/exaspim_acquisition.json
@@ -10,8 +10,10 @@
    "calibrations": [
       {
          "date_of_calibration": "2022-11-22T08:43:00",
-         "description": "Cleaning chamber",
-         "notes": "We used a special liquid XXX"
+         "calibration_type": "chamber",
+         "notes": "Some notes",
+         "input": {},
+         "output": {}
       }
    ],
    "session_start_time": "2022-11-22T08:43:00",
--- a/examples/exaspim_acquisition.json
+++ b/examples/exaspim_acquisition.json
@@ -1,12 +1,19 @@
 {
    "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/imaging/acquisition.py",
-   "schema_version": "0.4.6",
+   "schema_version": "0.4.7",
    "experimenter_full_name": [
       "###"
    ],
    "specimen_id": "###",
    "subject_id": "###",
    "instrument_id": "###",
+   "calibrations": [
+      {
+         "date_of_calibration": "2022-11-22T08:43:00",
+         "description": "Cleaning chamber",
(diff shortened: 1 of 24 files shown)


dyf:
@camilolaiton can you just update Mike's branch directly? https://github.com/AllenNeuralDynamics/aind-data-schema/tree/feat-imaging-calibration

saskiad:
I think for a generic class we should have input and output dictionaries. I know for the behavior calibrations and laser calibrations we will need this.
If we anticipate having solutions (e.g. cleaning agents as in the example here) then we should use the reagent class as an optional input

saskiad:
also given that we need this class for lots of different types of sessions, I would recommend putting this in core I think?

src/aind_data_schema/imaging/acquisition.py
```diff
@@ -98,10 +98,18 @@ class ProcessingSteps(AindModel):
     ]
 
 
+class Calibration(AindModel):
+    """Generic calibration class for acquisition"""
+
+    date_of_calibration: datetime = Field(..., title="Date and time of calibration")
+    description: str = Field(..., title="Calibration description")
```
saskiad, line 105: Instead of description, I would have "calibration_type" and make it an enum. Laser. Monitor. Speaker. Water valve. Chamber. Other

src/aind_data_schema/imaging/acquisition.py
```diff
@@ -98,10 +98,18 @@ class ProcessingSteps(AindModel):
...
+    notes: Optional[str] = Field(..., title="Notes")
```
saskiad, line 106: and input and output as dictionaries. Many calibrations involve testing different parameter settings and seeing the output.

examples/exaspim_acquisition.json
```diff
@@ -1,12 +1,19 @@
...
       "###"
    ],
    "specimen_id": "###",
    "subject_id": "###",
    "instrument_id": "###",
+   "calibrations": [
+      {
+         "date_of_calibration": "2022-11-22T08:43:00",
+         "description": "Cleaning chamber",
+         "notes": "We used a special liquid XXX"
```
saskiad, line 14: if we anticipate other calibrations to involve reagents we should add the reagent class as a optional field

saskiad (changes requested):
see comments

camilolaiton:
Closing this merge to update Mike's branch. https://github.com/AllenNeuralDynamics/aind-data-schema/tree/feat-imaging-calibration

Issue #454, opened 2023-08-16
injection concentration disappeared


saskiad:
add concentration back to injection materials

PR #455, opened 2023-08-17
added concentration back

saskiad:
The concentration field we added to injection materials was somehow disappeared. This brings it back.


saskiad:
closese #454

PR #476, opened 2023-09-08
Fix bug in `EphysRig` validator

### LINKED ISSUE #458: Cannot initialize a rig with DAQs that are connected to each other via a channel
When creating a rig object, if a list of DAQs is provided (e.g. daqs = [basestation, nidaq0, nidaq1]), if basestation is connected to nidaq0 via one of its channels (in the "channels" field), the program will fail when adding basestation to the rig, claiming that nidaq0 is mentioned in a channel but not part of the rig. In other words, even though the second DAQ device would be added to the rig in the next step, adding the first one fails because the second does not exist yet.

There is a workaround for this, which is adding the DAQ devices to the rig first with no channels, and then initializing rig.daqs[i].channels afterwards, but I feel that this was an oversight, as it should be possible to initialize all your devices and channels first and add them to the rig at the very end as the last step.

image

In this case, the code is failing when trying to add basestation to the list of DAQs, as basestation's channel mentions nipx, which is not a part of the rig (yet).

jsiegle:
The validator for the `EphysRig` class checks whether all devices connected to `DAQDevice` objects exist in the schema. But it did not previously allow connections *between* `DAQDevice` objects. This change fixes that.

Fixes #458

### DIFF
--- a/src/aind_data_schema/ephys/ephys_rig.py
+++ b/src/aind_data_schema/ephys/ephys_rig.py
@@ -144,7 +144,7 @@ class EphysRig(AindCoreModel):
     notes: Optional[str] = Field(None, title="Notes")
 
     @root_validator
-    def validate_device_names(cls, values): # noqa: C901
+    def validate_device_names(cls, values):  # noqa: C901
         """validate that all DAQ channels are connected to devices that
         actually exist
         """
--- a/src/aind_data_schema/ephys/ephys_rig.py
+++ b/src/aind_data_schema/ephys/ephys_rig.py
@@ -144,7 +144,7 @@ class EphysRig(AindCoreModel):
     notes: Optional[str] = Field(None, title="Notes")
 
     @root_validator
-    def validate_device_names(cls, values):
+    def validate_device_names(cls, values): # noqa: C901
         """validate that all DAQ channels are connected to devices that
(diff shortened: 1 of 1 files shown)


jsiegle:
@jtyoung84 Any recommendations on how to fix the linter issue? (`EphysRig.validate_device_names` is too complex)

jtyoung84:
> @jtyoung84 Any recommendations on how to fix the linter issue? (`EphysRig.validate_device_names` is too complex)

@jsiegle It's not something that should be done very often, but if it isn't possible to reduce the complexity in code, you can add a tag to tell flake8 to ignore the block. You can see an example here: https://github.com/AllenNeuralDynamics/aind-data-transfer/blob/15151e53de39aa11141391c8fc199e6bd8e0fe12/src/aind_data_transfer/jobs/openephys_job.py#L54

It should be possible to test flake8 locally. Also, for internal collaborators, we've leaned towards using feature branches instead of forks. It's not a big deal if you have a preference for opening prs through forks, but it makes it easier for maintainers to pull down the branch.

jsiegle:
Thanks, just fixed it.

I'll be sure to use feature branches for future PRs.

jtyoung84 (approved):

PR #481, opened 2023-09-13
Feat 480 update data level enum

### LINKED ISSUE #480: update DataLevel enum with new nomenclature
in order to change the tags, schema will need to be updated first. we are moving to utilize the terms [raw, derived] for both tags and custom metadata.

Sun-flow:
closes #480 

Updates the DataLevel enum to reflect the newly standardized tag/custom metadata terms


src/aind_data_schema/data_description.py
```diff
@@ -202,7 +202,7 @@ class RelatedData(AindModel):
 class DataDescription(AindCoreModel):
     """Description of a logical collection of data files"""
 
-    schema_version: str = Field("0.7.6", title="Schema Version", const=True)
+    schema_version: str = Field("0.7.7", title="Schema Version", const=True)
```
dyf, line 205: let's bump to `0.8.0` since this is a breaking change.

dyf (approved):

PR #482, opened 2023-09-14
feat-362-add-analysis-to-processing-schema

### LINKED ISSUE #362: Make an analysis schema
Our schema should be able to describe an analysis result.

Sun-flow:
closes #362 

Implements an analysis subtype of DataManipulation. Allows for tracking of processing separately from analysis in schema.


src/aind_data_schema/processing.py
```diff
@@ -55,21 +56,45 @@ class DataProcess(AindModel):
...
+
+    pass
 
 
 class Processing(AindCoreModel):
     """Description of all processes run on data"""
 
     schema_version: str = Field(
-        "0.2.4",
+        "0.2.5",
```
saskiad, line 86: let's make this 0.3.0

src/aind_data_schema/processing.py
```diff
@@ -55,21 +56,45 @@ class DataProcess(AindModel):
...
     parameters: Dict[str, Any] = Field(..., title="Parameters")
     outputs: Optional[Dict[str, Any]] = Field(None, description="Output parameters", title="Outputs")
-    notes: Optional[str] = None
+    notes: Optional[str] = Field(None, title="Notes")
+
+
+class DataManipulation(AindModel):
+    """Description of a set of Data Manipulations (processing/analysis steps)"""
+
+    person: str = Field(..., title="Person")
```
saskiad, line 65: person should be in the pipelineProcess class.
And lets say "Person full name"? We need to indicate full name. Not sure if person is the right term....

src/aind_data_schema/processing.py
```diff
@@ -55,21 +56,45 @@ class DataProcess(AindModel):
...
+class PipelineProcess(DataManipulation):
+    """Description of a Processing Pipeline"""
+
+    pipeline_version: Optional[str] = Field(None, description="Version of the pipeline", title="Pipeline version")
+    pipeline_url: Optional[str] = Field(None, description="URL to the pipeline code", title="Pipeline URL")
+
+
+class AnalysisProcess(DataManipulation):
+    """Description of an Analysis"""
+
```
saskiad, line 78: here we want the person full name as well

src/aind_data_schema/processing.py
```diff
@@ -55,21 +56,45 @@ class DataProcess(AindModel):
     code_version: Optional[str] = Field(None, description="Version of the code", title="Code version")
     parameters: Dict[str, Any] = Field(..., title="Parameters")
     outputs: Optional[Dict[str, Any]] = Field(None, description="Output parameters", title="Outputs")
-    notes: Optional[str] = None
+    notes: Optional[str] = Field(None, title="Notes")
+
+
+class DataManipulation(AindModel):
```
saskiad, line 62: Do we need DataManipulation? Can we just use DataProcess?

saskiad (changes requested):
A few changes - I'm not sure we need DataManipulation (or we can rename DataProcess to DataManipulation ... but it feels like it's just a wrapper of DataProcess

src/aind_data_schema/processing.py
Sun-flow: Data process is a specific part of the pipeline, datamanipulation contains a set of these processes. This class is inhereted by both of our new pipeline classes, and is performing a different function than the DataProcess class is.

src/aind_data_schema/processing.py
saskiad: Analyst Full name

src/aind_data_schema/processing.py
```diff
@@ -55,21 +56,52 @@ class DataProcess(AindModel):
...
+    """Description of a set of Data Manipulations (processing/analysis steps)"""
+
+    data_processes: List[DataProcess] = Field(..., title="Data processing", unique_items=True)
+
+
+class PipelineProcess(DataManipulation):
+    """Description of a Processing Pipeline"""
+
+    processing_person: str = Field(
```
saskiad, line 71: use "full name" instead of person to match how we do experimenter full name in procedures/sessions/etc

src/aind_data_schema/processing.py
saskiad: I still don't understand - DataManipulation is a list of DataProcess. Why not just have a list of DataProcess in PipelineProcess and in AnalysisProcess?

saskiad: in fact right now DataProcess doesn't actually get used anywhere - or am I confused?

Sun-flow: Since DataManipulation is a parent class of both AnalysisPipeline and ProcessPipeline, it's effectively the same as having a list in either of them. But it maintains clarity of the shared structure, and reduces copy-paste code.

Sun-flow: > in fact right now DataProcess doesn't actually get used anywhere - or am I confused?

DataProcess is used within DataManipulation, which is inherited by ProcessPipeline and AnalysisPipeline.

src/aind_data_schema/processing.py
```diff
@@ -55,21 +57,50 @@ class DataProcess(AindModel):
...
+    notes: Optional[str] = Field(None, title="Notes")
+
+
+class DataManipulation(AindModel):
+    """Description of a set of Data Manipulations (processing/analysis steps)"""
+
+    data_processes: List[DataProcess] = Field(..., title="Data processing", unique_items=True)
+
+
+class PipelineProcess(DataManipulation):
```
saskiad, line 69: oh I see. I think the problem is that the pipeline version applies to a list of processing steps. So having each step have pipeline version/url is not the right way. I think PipelineProcess(AindModel) can have the fields it has and then List[DataProcess].

src/aind_data_schema/processing.py
```diff
@@ -55,21 +57,50 @@ class DataProcess(AindModel):
...
+    """Description of a Processing Pipeline"""
+
+    processor_full_name: str = Field(
+        ..., title="Processor Full Name", description="Name of person responsible for processing pipeline"
+    )
+    pipeline_version: Optional[str] = Field(None, description="Version of the pipeline", title="Pipeline version")
+    pipeline_url: Optional[str] = Field(None, description="URL to the pipeline code", title="Pipeline URL")
+
+
+class AnalysisProcess(DataManipulation):
```
saskiad, line 79: and then this just is a DataProcess instance and we don't need to have the duplicate DataManipulation calss

saskiad (changes requested):
see new comments - I still think we only need DataProcess and not DataManipulation

src/aind_data_schema/processing.py
```diff
@@ -55,21 +57,46 @@ class DataProcess(AindModel):
...
     outputs: Optional[Dict[str, Any]] = Field(None, description="Output parameters", title="Outputs")
-    notes: Optional[str] = None
+    notes: Optional[str] = Field(None, title="Notes")
+
+
+class PipelineProcess(AindModel):
+    """Description of a Processing Pipeline"""
+
+    data_processes: List[DataProcess] = Field(..., title="Data processing", unique_items=True)
+
```
saskiad, line 67: remove this blank line

src/aind_data_schema/processing.py
```diff
@@ -55,21 +57,46 @@ class DataProcess(AindModel):
     code_version: Optional[str] = Field(None, description="Version of the code", title="Code version")
     parameters: Dict[str, Any] = Field(..., title="Parameters")
     outputs: Optional[Dict[str, Any]] = Field(None, description="Output parameters", title="Outputs")
-    notes: Optional[str] = None
+    notes: Optional[str] = Field(None, title="Notes")
+
```
saskiad, line 61: up in line 49 or 51 there's a version and then there's code_version. I think these are redundant. I'd keep the code_version and remove the version.

src/aind_data_schema/processing.py
```diff
@@ -55,21 +57,46 @@ class DataProcess(AindModel):
...
         title="Schema version",
         const=True,
     )
-    pipeline_version: Optional[str] = Field(None, description="Version of the pipeline", title="Pipeline version")
-    pipeline_url: Optional[str] = Field(None, description="URL to the pipeline code", title="Pipeline URL")
-    data_processes: List[DataProcess] = Field(..., title="Data processing", unique_items=True)
+    processing_pipeline: PipelineProcess = Field(
+        ..., description="Pipeline used to process data", title="Processing Pipeline"
+    )
+    analysis: Optional[List[AnalysisProcess]] = Field(
```
saskiad, line 96: let's make this analyses (since it will be a list and likely have multiple items)

src/aind_data_schema/processing.py
```diff
@@ -55,21 +57,46 @@ class DataProcess(AindModel):
...
+    """Description of a Processing Pipeline"""
+
+    data_processes: List[DataProcess] = Field(..., title="Data processing", unique_items=True)
+
+    processor_full_name: str = Field(
+        ..., title="Processor Full Name", description="Name of person responsible for processing pipeline"
+    )
+    pipeline_version: Optional[str] = Field(None, description="Version of the pipeline", title="Pipeline version")
+    pipeline_url: Optional[str] = Field(None, description="URL to the pipeline code", title="Pipeline URL")
+
```
saskiad, line 73: will you just add a notes field (Optional[str]) to this PiplineProcess class for me? Thanks!

saskiad (changes requested):
a couple of little things, but otherwise it's good

src/aind_data_schema/processing.py
Sun-flow: From what I gathered, version was like "the software used to run the script" and code version was like "the version of the script". That was just my assumption though? These terms were both already here, I did not add them afaik.

saskiad: yeah, let's change 'version' to 'software_version' so it's clear

saskiad (changes requested):
just change that one field name and it's gtg!

dyf (changes requested):
Small change - let's make sure that `AnalysisProcess.name` is always `ProcessName.ANALYSIS`.

dyf (approved):

saskiad (approved):

PR #486, opened 2023-09-15
Unified rig/session schema

saskiad:


src/aind_data_schema/session.py
```diff
@@ -0,0 +1,324 @@
...
+    trials_rewarded: int = Field(..., title="Rewarded trials")
+
+
+class Stream(AindModel):
+    """Stream of data with a start and stop time"""
+
+    stream_start_time: datetime = Field(..., title="Stream start time")
+    stream_end_time: datetime = Field(..., title="Stream stop time")
+    stream_modalities: List[Modality] = Field(..., title="Modalities")
+    behavior: Optional[Behavior] = Field(None, title="Behavior")
```
dyf, line 251: Is it one `Behavior` per stream or, is it per `Session`?

src/aind_data_schema/session.py
```diff
@@ -0,0 +1,324 @@
...
+    stimulus_epochs: Optional[List[StimulusEpoch]] = Field(None, title="Stimulus")
+    reward_delivery: Optional[RewardDelivery] = Field(None, title="Reward delivery")
+    daq_names: Optional[List[str]] = Field(None, title="DAQ devices", unique_items=True)
+    camera_names: Optional[List[str]] = Field(None, title="Cameras", unique_items=True)
```
dyf, line 255: Do we need these to indicate they are active? If so, maybe `activate_camera_names`?

src/aind_data_schema/session.py
```diff
@@ -0,0 +1,324 @@
...
+    trials_finished: int = Field(..., title="Finished trials")
+    trials_rewarded: int = Field(..., title="Rewarded trials")
+
+
+class Stream(AindModel):
+    """Stream of data with a start and stop time"""
+
+    stream_start_time: datetime = Field(..., title="Stream start time")
+    stream_end_time: datetime = Field(..., title="Stream stop time")
+    stream_modalities: List[Modality] = Field(..., title="Modalities")
```
dyf, line 250: My (potentially wrong) understanding is that a stream should only have one modality, since a session can have many streams.

src/aind_data_schema/session.py
dyf: and if so, would it be simpler just to have `active_device_names: Optional[List[str]]`? Just a thought, maybe useful to keep them separate.

src/aind_data_schema/rig.py
```diff
@@ -0,0 +1,67 @@
...
+    ephys_assemblies: Optional[List[EphysAssembly]] = Field(None, title="Ephys probes", unique_items=True)
+    stick_microscopes: Optional[List[StickMicroscopeAssembly]] = Field(None, title="Stick microscopes")
+    laser_assemblies: Optional[List[LaserAssembly]] = Field(None, title="Laser modules", unique_items=True)
+    patch_cords: Optional[List[Patch]] = Field(..., title="Patch cords", unique_items=True)
+    light_sources: List[Union[Laser, LightEmittingDiode]] = Field(..., title="Light sources", unique_items=True)
+    detectors: Optional[List[Detector]] = Field(None, title="Detectors", unique_items=True)
+    objectives: Optional[List[Objective]] = Field(None, title="Objectives", unique_items=True)
+    filters: Optional[List[Filter]] = Field(None, title="Filters", unique_items=True)
+    lenses: Optional[List[Lens]] = Field(None, title="Lenses", unique_items=True)
+    additional_devices: Optional[List[Device]] = Field(None, title="Additional devices", unique_items=True)
```
dyf, line 65: one more: do you just want this to be `devices: Optional[List[...everything...]]`?

dyf (commented):
one more comment

src/aind_data_schema/session.py
saskiad: no, a stream is everything recorded at one time. So multiple modalities. A session can have multiple streams, but can also be one stream. Think of it more as we positioned the electrodes, recorded activity in response to stimulus A, then moved the electrodes to different positions, recorded activity in response to stimulus B.

src/aind_data_schema/session.py
saskiad: Honestly, I can imagine use cases for both, but I think session is the one we want for now. thanks

src/aind_data_schema/session.py
saskiad: the fact that they are listed in the session means that they are active.

saskiad: I considered that - I was going to make it something that was device type and name, because people do a bad job of naming devices. Eg. If I trusted people to name things "Camera A" and "DAQ 1" I would be fine with just a list of names, but I think the schema will be much easier to understand if we know what device is what.

src/aind_data_schema/rig.py
saskiad: I think there's benefit in knowing what's what. But maybe that can go in the device classes? Device type?

saskiad:
The thing I'm stuck on is where stimulus epoch lives. It presently lives within the Behavior class, because that's where the stimulus side of the behavior gets specifed. But in passive experiments, the behavior class won't be instantiated, but there still needs to be a stimulus epoch field outside of behavior. And we could remove it from Behavior, but conceptually that is weird...

src/aind_data_schema/rig.py
saskiad: here's a question - we're going to need validators that say "if modality is ephys then ephys_assemblies and stick_microscopes can not be empty" etc. If we make the list of devices, my hunch is that validation will be harder. But I don't know the validators that well - what do you think?

jsiegle (changes requested):
Overall this looks good, it will definitely streamline things. I assume this will deprecate `ephys_session` and `ephys_rig` once it's merged?

src/aind_data_schema/session.py
```diff
@@ -0,0 +1,294 @@
...
+        title="Data streams",
+        description=(
+            "A data stream is a collection of devices that are recorded simultaneously. Each session can include"
+            " multiple streams (e.g., if the manipulators are moved to a new location)"
+        ),
+        unique_items=True,
+    )
+    stimulus_epochs: Optional[List[StimulusEpoch]] = Field(None, title="Stimulus")
+    reward_delivery: Optional[RewardDelivery] = Field(None, title="Reward delivery")
+    stick_microscopes: Optional[List[DomeModule]] = Field(
```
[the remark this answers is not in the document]
saskiad: But they aren't the same. The camera_names is the list of names of cameras - mostly the behavior cameras. Whereas stick_microscopes are DomeModules that have angle and position information. That will get lost if we remove this.

src/aind_data_schema/session.py
```diff
@@ -0,0 +1,294 @@
...
+
+    name: str = Field(..., title="Ephys probe name (must match rig JSON)")
+    other_targeted_structures: Optional[List[str]] = None
+
+
+class EphysModule(ManipulatorModule):
+    """Probe recorded in a Stream"""
+
+    ephys_probes: List[EphysProbe] = Field(..., title="Ephys probes used in this module")
+
```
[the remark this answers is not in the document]
saskiad: moving to #505

src/aind_data_schema/session.py
```diff
@@ -0,0 +1,294 @@
...
+class Stream(AindModel):
+    """Data streams with a start and stop time"""
+
+    stream_start_time: datetime = Field(..., title="Stream start time")
+    stream_end_time: datetime = Field(..., title="Stream stop time")
+    stream_modalities: List[Modality] = Field(..., title="Modalities")
+    daq_names: Optional[List[str]] = Field(None, title="DAQ devices", unique_items=True)
+    camera_names: Optional[List[str]] = Field(None, title="Cameras", unique_items=True)
+    light_sources: Optional[List[Union[Laser, LightEmittingDiode]]] = Field(..., title="Light source", unique_items=True)
+    ephys_modules: Optional[List[EphysModule]] = Field(None, title="Ephys modules", unique_items=True)
```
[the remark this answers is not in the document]
saskiad: Moving to #505

src/aind_data_schema/device.py
```diff
@@ -540,6 +369,158 @@ class LightEmittingDiode(Device):
...
+    headstage: Optional[Headstage] = Field(None, title="Headstage for this probe")
+
+
+class EphysAssembly(AindModel):
+    """Module for electrophysiological recording"""
+
+    ephys_assembly_name: str = Field(..., title="Ephys assembly name")
+    manipulator: Manipulator = Field(..., title="Manipulator")
+    probes: List[EphysProbe] = Field(..., title="Probes that are held by this module")
+
```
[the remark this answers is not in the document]
saskiad: moving to #505

src/aind_data_schema/rig.py
```diff
@@ -0,0 +1,71 @@
...
+    Treadmill,
+    Tube,
+    Wheel,
+)
+
+
+class Rig(AindCoreModel):
+    """Description of a rig"""
+
+    schema_version: str = Field("0.1.0", description="schema version", title="Version", const=True)
```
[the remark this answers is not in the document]
dyf: @jsiegle if we did this then `Rig` would be version-locked with the schema library as a whole, which means we wouldn't have a way to communicate how individual core models have changed.  Unless you are suggesting something else?

src/aind_data_schema/session.py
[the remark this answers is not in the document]
saskiad: I can move the 'stick_microscopes' to 'Stream' - I am torn on whether it belongs there or here.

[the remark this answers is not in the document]
jsiegle: My logic was that the stick microscope doesn't generate data on its own -- only the camera inside it does, and these cameras are associated with microscopes in `rig.json`. However, we do need to specify the arc and module angle of each microscope, so the `stick_microscopes` does need to go in `session.json`. But it's not technically a manipulator module, since it doesn't have a manipulator. I'm realizing that this is why we separated `DomeModule` and `ManipulatorModule` previously. We could either add back the `DomeModule` class, or make the `manipulator_coordinates` optional.

dyf (commented):
Some minor comments, but otherwise I think all that's left is to update the example to use `Rigs` and `Session`.

dyf (approved):

PR #504, opened 2023-09-27
moves data asset record to repo

### LINKED ISSUE #384: Move DataAssetRecord model to this repo
# User story
As a user, I want to store all the schema instances into a single record, so I can manage and update it more easily.

# Acceptance criteria
- [ ] The DataAssetRecord model should be moved to this repo (https://github.com/AllenNeuralDynamics/aind-data-access-api/blob/main/src/aind_data_access_api/models.py)
- [ ] An agreement should be made as to what the default filename should be (.aind.metadata.json)?
- [ ] An agreement should be made as to what the required fields should be (id, created, last_modified, location, curated, quality_controlled)?
- [ ] Add field for list of AindCoreModels?

# Sprint Ready Checklist 
- [ ] 1. Acceptance criteria defined 
- [ ] 2. Team understands acceptance criteria 
- [ ] 3. Team has defined solution / steps to satisfy acceptance criteria 
- [ ] 4. Acceptance criteria is verifiable / testable 
- [ ] 5. External / 3rd Party dependencies identified
- [ ] 6. Ticket is prioritized and sized

# Notes
*Add any helpful notes here.*

mekhlakapoor:
closes #384 

Some discussion points we need to agree upon: 
- What should the default filename be (.aind.metadata.json)?
- What should the required fields be (currently id, created, last_modified, location, metadata_status)? should we include curated, quality_controlled?
- Should extra have some validation rules (for example, expect AindCoreModel)? or can we keep it general

### DIFF
--- a/src/aind_data_schema/data_asset_record.py
+++ b/src/aind_data_schema/data_asset_record.py
@@ -1,0 +1,60 @@
+"""Module for models"""
+
+from datetime import datetime
+
+from pydantic import BaseModel, Extra, Field
+
+
+class DataAssetRecord(BaseModel):
+    """The records in the Data Asset Collection needs to contain certain fields
+    to easily query and index the data."""
+
+    code_ocean_id: str = Field(
+        ...,
+        alias="_id",
+        title="Data Asset ID",
+        description="The unique id of the data asset.",
+    )
+    name: str = Field(
+        ...,
+        alias="_name",
+        description="Name of the data asset.",
+        title="Data Asset Name",
+    )
+    created: datetime = Field(
+        ...,
+        alias="_created",
+        title="Created",
+        description="The data and time the data asset created.",
+    )
+    location: str = Field(
+        ...,
+        alias="_location",
+        title="Location",
+        description="Current location of the data asset.",
+    )
+
+    class Config:
+        """Need to allow for additional fields to append to base model"""
+
+        extra = Extra.allow
+
+    @property
+    def _id(self):
+        """Property for _id"""
+        return self.code_ocean_id
+
+    @property
+    def _name(self):
+        """Property for _name"""
+        return self.name
+
+    @property
+    def _location(self):
+        """Property for _location"""
+        return self.location
+
+    @property
+    def _created(self):
+        """Property for _created"""
+        return self.created
--- a/src/aind_data_schema/data_asset_record.py
+++ b/src/aind_data_schema/data_asset_record.py
@@ -1,15 +1,27 @@
 """Module for models"""
 
 from datetime import datetime
+from enum import Enum
 
-from pydantic import BaseModel, Extra, Field
+from pydantic import Extra, Field
+
+from aind_data_schema.base import AindCoreModel
 
 
-class DataAssetRecord(BaseModel):
+class MetadataStatus(Enum):
+    """Status of Metadata"""
+
+    VALID = "Valid"
+    INVALID = "Invalid"
+    MISSING = "Missing"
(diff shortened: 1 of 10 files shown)


mekhlakapoor:
David: 

(1) what do we use for _id? is it currently Code Ocean asset id? 
(2) do we need to explicitly version this? 

For (1) I don't have a strong feeling. I guess it might be nice to have it explicitly be  a code_ocean_id, in which case _id could be a plai

 

For (2) I think yes, and maybe it should just be the aind_data_schema library version.

src/aind_data_schema/data_asset_record.py
```diff
@@ -0,0 +1,60 @@
...
+from datetime import datetime
+
+from pydantic import BaseModel, Extra, Field
+
+
+class DataAssetRecord(BaseModel):
+    """The records in the Data Asset Collection needs to contain certain fields
+    to easily query and index the data."""
+
+    code_ocean_id: str = Field(
```
jtyoung84, line 12: This should just be id

src/aind_data_schema/data_asset_record.py
```diff
@@ -0,0 +1,60 @@
...
+        ...,
+        alias="_id",
+        title="Data Asset ID",
+        description="The unique id of the data asset.",
+    )
+    name: str = Field(
+        ...,
+        alias="_name",
```
jtyoung84, line 20: We can probably remove the alias on all the fields other than id

src/aind_data_schema/data_asset_record.py
```diff
@@ -0,0 +1,60 @@
...
+        description="Name of the data asset.",
+        title="Data Asset Name",
+    )
+    created: datetime = Field(
```
jtyoung84, line 24: We can also add a last_modified field

dyf:
> > David:
> > (1) what do we use for _id? is it currently Code Ocean asset id? (2) do we need to explicitly version this?
> > For (1) I don't have a strong feeling. I guess it might be nice to have it explicitly be a code_ocean_id, in which case _id could be a plai
> > For (2) I think yes, and maybe it should just be the aind_data_schema library version.
> 
> For 1, this is an index of our data assets that we maintain in our buckets. It will have nothing to do with Code Ocean directly. It will either be a uuid or be the same as the location field. We can add a field to link to Code Ocean if we want. For 2, it can have a schema_version the same as the other fields. We can use S3's versioning to track updates to the json file directly and maybe pass that along to a "record_version" field or something.

I'm good with all of this.

dyf:
Suggest we call this `DataAsset` and `data_asset.py`, unless others feel strongly.

mekhlakapoor:
We'll handle adding metadata fields (such as subject) in another ticket (once Saskia's refactor is complete)

saskiad:
> Suggest we call this `DataAsset` and `data_asset.py`, unless others feel strongly.

I do feel strongly. I think we should call it DataRecord. We use data asset already to refer to the data asset - it will be confusing to call this data asset too.

mekhlakapoor:
> > Suggest we call this `DataAsset` and `data_asset.py`, unless others feel strongly.
> 
> I do feel strongly. I think we should call it DataRecord. We use data asset already to refer to the data asset - it will be confusing to call this data asset too.

Just renamed it. Let me know if there's any other comments.

dyf:
Sorry, we just pow-wowed at decided:
- let's just call it `Metadata`
- it would be defined in `metadata.py`
- and it would generate `metadata.json`

dyf (approved):

PR #522, opened 2023-10-09
feat-169 adding validators for 'Other'

### LINKED ISSUE #169: Validator for "Other"
When people select Other from an Enum, they need to provide the information in the Notes. Can we have a validator that forces this?

Sun-flow:
closes #169 

Adds a validator that checks to ensure there is text in the `notes` field whenever an `enum.OTHER` instance is passed in.

Locations I have found that contain reference to `enum.OTHER` and contain a `notes` field: `procedures.SpecimenProcedure`, `processing.DataProcess`, `session.RewardDelivery`, `behavior_session.RewardDelivery`, `instrument.Instrument`, `mri_session.MRIScan`


src/aind_data_schema/behavior/behavior_session.py
```diff
@@ -40,6 +40,14 @@ class RewardDelivery(AindModel):
     reward_spouts: List[RewardSpout] = Field(..., title="Reward spouts", unique_items=True)
     notes: Optional[str] = Field(None, title="Notes")
 
+    @root_validator
+    def other_notes(cls, v):
```
dyf, line 44: can we call this "validate_other"?

src/aind_data_schema/behavior/behavior_session.py
```diff
@@ -40,6 +40,14 @@ class RewardDelivery(AindModel):
...
+        """Validator for other/notes"""
+
+        if v.get("reward_solution") == RewardSolution.OTHER and v.get("notes") is None:
+            raise ValueError("Other must be specified in notes")
```
dyf, line 48: This message is unclear - maybe, "If reward_solution is "Other", the notes field must be non-empty."

src/aind_data_schema/imaging/instrument.py
```diff
@@ -177,3 +177,15 @@ class Instrument(AindCoreModel):
     )
     com_ports: Optional[List[Com]] = Field(None, title="COM ports", unique_items=True)
     notes: Optional[str] = None
+
+    @root_validator
+    def other_notes(cls, v):
```
dyf, line 182: same as above, call it "validate_other"

src/aind_data_schema/behavior/behavior_session.py
dyf: Let's check for an empty string as well. I think `... and not v.get("notes")` will handle both cases, but please confirm.

src/aind_data_schema/imaging/instrument.py
```diff
@@ -177,3 +177,15 @@ class Instrument(AindCoreModel):
...
+        """Validator for other/notes"""
+
+        if v.get("instrument_type") == ImagingInstrumentType.OTHER and v.get("notes") is None:
+            raise ValueError("Instrument Type: Other must be specified in notes")
```
dyf, line 186: same comment from above applies to these two messages and string checks.

src/aind_data_schema/imaging/mri_session.py
```diff
@@ -81,6 +81,13 @@ class MRIScan(AindModel):
     additional_scan_parameters: Dict[str, Any] = Field(..., title="Parameters")
     notes: Optional[str] = Field(None, title="Notes")
 
+    @root_validator
+    def other_notes(cls, v):
```
dyf, line 85: ditto for this one

tests/test_imaging.py
```diff
@@ -74,6 +74,29 @@ def test_constructors(self):
 
         assert i is not None
 
+        with self.assertRaises(ValidationError):
```
dyf, line 77: add a test for `notes=""` for at least one of these

dyf (changes requested):
A few requests regarding on naming, and such. Please apply everywhere they apply (I stopped typing "ditto" after awhile).

src/aind_data_schema/behavior/behavior_session.py
saskiad: While we can only test that the notes is not empty, I think the instructions we want to give people is to provide what the Other is in the notes. But I do like having an error message that points people to where the problem is. So maybe something like "reward_solution of "Other" must be specified in the notes"

saskiad (commented):
Ditto David's comments

dyf (approved):

PR #525, opened 2023-10-10
fix: updates union types

### LINKED ISSUE #518: Consider adding support for `Union` discriminators

AFAIK, there is currently no way to, on deserialization, know the type that is included in a given `Union`. While `Union`s are flexible and necessary, the inability to ensure that types are preserved and to (de)serialize them while maintaining such type information is critical, since it affords polymorphic behavior, especially for applications that rely on strongly typing these objects (e.g. c# and bonsai).

Pydantic’s discriminator feature (which follows the proposal by OpenAPI, see below) allows the definition of data structures with multiple types, using a discriminator field to determine the actual object type.
Some references:
https://typethepipe.com/post/pydantic-discriminated-union/
https://docs.pydantic.dev/latest/concepts/fields/
https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/
https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#discriminatorObject

Using pydantic2 seems to improve this support even further by simply declaring
```python
Field(discriminator='type')])
```

**Additional context**
This would be a critical feature for a sustainable integration between python-schemas<->JSON-schemas<->bonsai.

jtyoung84:
Closes #518 

- Fixes issues with de-serializing fields with Union types
- Updates test_examples to output more informative error message


examples/procedures.json
mekhlakapoor, line 1: when I add a newline at the end of file, it fails coverage tests

src/aind_data_schema/session.py
```diff
@@ -213,9 +214,12 @@ class Stream(AindModel):
     stream_modalities: List[Modality] = Field(..., title="Modalities")
     daq_names: Optional[List[str]] = Field(None, title="DAQ devices", unique_items=True)
     camera_names: Optional[List[str]] = Field(None, title="Cameras", unique_items=True)
-    light_sources: Optional[List[Union[Laser, LightEmittingDiode]]] = Field(
-        None, title="Light source", unique_items=True
-    )
+    light_sources: Optional[
+        Annotated[
+            List[Union[Laser, LightEmittingDiode]],
```
dyf, line 219: It is confusing, but here `Laser` and `LightEmittingDiode` are not from `device.py`.  They are actually device configuration classes. We need to add `device_type` to `Laser` and `LightEmittingDiode` in this file.  

I'm going to create an issue to make this less confusing in a subsequent PR, but for the time being let's just add `device_type` to the models above.

dyf (changes requested):

src/aind_data_schema/session.py
jtyoung84: Added device_type to those classes

PR #535, opened 2023-10-12
Feat 500 add Analysis Description

### LINKED ISSUE #500: Add an `AnalysisDescription` subclass of `DataDescription`
- [ ] follow the pattern of `RawDataDescription` to make an `AnalysisDescription` subclass of `DataDescription` 
- [ ] the name field will have four components: `<project-abbreviation>_<analysis-name>_<date>_<time>`
- [ ] none of these can contain dashes

Sun-flow:
closes #500 

Implements a class similar to RawDataDescription, for the use of Analysis. 

Name format: `<project-abbreviation>_<analysis-name>_<date>_<time>`


Sun-flow:
@dyf do you think this is the right way to ensure no hyphens are in the name? 

I don't see any examples of the string being checked to exclude certain characters, unless it is hidden in the Regex in a way I don't notice. I see the `NO_UNDERSCORES` Regex, but it's not used for name checking only for field input.

src/aind_data_schema/data_description.py
```diff
@@ -31,7 +31,12 @@ class DataRegex(Enum):
         f"^(?P<input>.+?_{RegexParts.DATE.value}_{RegexParts.TIME.value})_(?P<process_name>.+?)_(?P<c_date>"
         f"{RegexParts.DATE.value})_(?P<c_time>{RegexParts.TIME.value})"
     )
+    ANALYZED = (
```
jtyoung84, line 34: We probably should use re.compile to precompile the regex patterns.

src/aind_data_schema/data_description.py
```diff
@@ -437,3 +443,47 @@ def parse_name(cls, name):
...
+
+
+class AnalysisDescription(DataDescription):
+    """A collection of data files as analyzed from an asset"""
+
+    data_level: DataLevel = Field(
+        DataLevel.DERIVED, description="Level of processing that data has undergone", title="Data Level", const=True
+    )
+
+    def __init__(self, analysis_name, **kwargs):
```
jtyoung84, line 455: It's probably tech-debt that needs to be cleaned up elsewhere also, but we should avoid writing custom constructors on pydantic models. You can just override the project_name field to make it mandatory instead of doing the check here. This information doesn't get serialized into the json_schema. I'll leave it up to @dyf as to whether we should fix it as a tech-debt ticket.

src/aind_data_schema/data_description.py
dyf: https://github.com/AllenNeuralDynamics/aind-data-schema/issues/545

src/aind_data_schema/data_description.py
dyf: Agree with @jtyoung84 . Go ahead and override `project_name` here and make it required instead of optional.

dyf (changes requested):

src/aind_data_schema/data_description.py
Sun-flow: The only reason I can see that it might still be useful is that we are constructing custom label fields, which differ between models. Should I just not have a custom label?

jtyoung84: You can make it a property like:
```
@property
def label(self):
  return f"{self.project_name}_{self.analysis_name}"
```
or if you want it to be a part of the schema, you'll need to use a `validator` or `root_validator`

dyf (approved):

dyf:
@Sun-flow I didn't mean to approve this before the tests were passing.  Go ahead and fix those and then we can merge.

src/aind_data_schema/schema_upgrade/data_description_upgrade.py
```diff
@@ -158,7 +158,7 @@ def upgrade_data_description(self, **kwargs) -> DataDescription:
             data_level=self._get_or_default(self.old_data_description_model, "data_level", kwargs),
             group=self._get_or_default(self.old_data_description_model, "group", kwargs),
             investigators=self._get_or_default(self.old_data_description_model, "investigators", kwargs),
-            project_name=self._get_or_default(self.old_data_description_model, "project_name", kwargs),
+            project_name=self._get_or_default(self.old_data_description_model, "projectname", kwargs),
```
dyf, line 161: I think this should still have an underscore in it?

dyf (changes requested):

src/aind_data_schema/schema_upgrade/data_description_upgrade.py
Sun-flow: The problem is that we have made `"_"` an invalid character for project name.

Sun-flow: oh this is for a getter, so maybe you're right

Sun-flow: no, leaving the underscore causes validation errors. I could use a hyphen, if you prefer

dyf (approved):

PR #537, opened 2023-10-16
using common devices

### LINKED ISSUE #511: Differences with similar devices between Imaging/EPhys
This might not be needed or worth the effort, but I recently noticed there are similar devices defined differently between Imaging and EPhys instruments/rigs.

1. EPhys instrument has a ```Lens``` whereas this is not present in the imaging schema. I am adding them in as additional imaging devices, but would prefer to inherit Lens definition from Ephys.
2. EPhys has a ```Camera``` whereas for Imaging we are using ```Detector```.  Again here does it make sense to update Imaging with the Camera definition from Ephys?
3. Same for ```Laser``` in Ephys and the current ```Lightsource``` in Imaging.

saskiad:
closes #511

### DIFF
--- a/examples/aibs_smartspim_instrument.json
+++ b/examples/aibs_smartspim_instrument.json
@@ -1,6 +1,6 @@
 {
    "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/imaging/instrument.py",
-   "schema_version": "0.8.2",
+   "schema_version": "0.9.0",
    "instrument_id": "SmartSPIM2-2",
    "modification_date": "2023-10-04",
    "instrument_type": "SmartSPIM",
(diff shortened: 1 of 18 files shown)


saskiad:
@adamkglaser Take a look at this and let me know if I missed anything. I'm still updating examples and tests but want to make sure all the things we discussed are here as I deal with all that.

thanks!

saskiad:
@adamkglaser can you also look at the list of manufacturers for the Laser, Lens classes and make sure we aren't missing any that you use there? Eg. Vortran wasn't in the list for Lasers - just added it based on one of the examples.

adamkglaser:
Everything looks great to me. For ```Lens``` I would add the following manufacturers

Nikon
Olympus
Zeiss
Leica
Mitutoyo
Hamamatsu
Schneider-Kreuznach

Thanks!
Adam

dyf (approved):

PR #558, opened 2023-10-24
Feat 509 rig and session validators

### LINKED ISSUE #509: Modality specific validators for rig and session schemas

Sun-flow:
closes #509 

Validators for ensuring the proper fields are completed for a given modality. 

Edit: Added change to schema regarding `Session`, `Stream`, and the field `stick_microscopes`, which has been moved from the former to the latter, and then validatred under `Stream`

Current implementation is as per @saskiad 's guidance:
  

For Rig:
  If ecephys modality, need ephys_assemblies, stick_microscope
  If FIB: need light_source, detector, patch_cords
  if pophys: need light_source, detector, objective
  If slap need light source, detector, objective (this will probably be a longer list soon)
  If behavior_videos: need cameras
  if trained_behavior: need stimulus_devices
  
  For session (some of these are in the Stream class):
  Ecephys: need stick_microscopes, ephys_modules <-- (There is no stick_microscopes field in Stream, and Session does not include a modality)
  FIB: need light_source, detector, fiber_photometry_assemblies
  POPHYS: need either ophys_fov OR stack_parameters ... not sure that's possible
  SLAP: not sure yet, hold off on this
  Behavior_videos: camera_names
  Trained_behavior: stimulus_device_names


saskiad (commented):
One solution for the stick_microscope is to have the list of modalities be a session property. I don't love that...

I assume this is a problem because the validator can only act on one specific class? Is it not possible to have a validator of the session that asks if any stream within the session has ecephys as a modality then the session must have stick_microscope? (I'm not super familiar with how the validators work so I'm asking rather than suggesting)

saskiad:
@Sun-flow what's the status of this?

examples/ophys_session.json
```diff
@@ -46,6 +46,18 @@
             }
          ],
          "ephys_modules": null,
+         "stick_microscopes": [
```
saskiad, line 49: I would leave this as null in the ophys_session as I'm not sure that they'll be used for ophys

saskiad: leave them null but moved to the stream as I believe you've done here, yes?

examples/ophys_session.py
```diff
@@ -36,6 +36,13 @@
                     excitation_power_unit="milliwatt",
                 ),
             ],
+            stick_microscopes=[
```
saskiad, line 39: see comment above - stick_microscopes should probably be null for the ophys_session example

saskiad (changes requested):
Just one small tweak - for the ophys_sesion example let's not include an actual stick_microscopes (let's have it be null) as I don't think stick microscopes will be used in ophys and we want our examples to be as real as possible.

saskiad (approved):

PR #568, opened 2023-10-27
fix: DataDescriptionUpgrade sets data_level to raw if not correct

### LINKED ISSUE #567: `DataDescriptionUpgrade` should handle wrong data level field values
**Describe the bug**
Some old datasets that have been re-processed recently have a wrong field in the data level:
`raw level` (instead of just `raw`).
This makes the `DataDescriptionUpgrade` fail at validation.

**To Reproduce**
Try to upgrade the `data_description.json` from session `ecephys_624643_2022-07-26_10-52-15`

**Expected behavior**
The `DataDescriptionUpgrade`  should take care of this issue and set the upgraded data level to RAW as default.

alejoe91:
fixes #567

### DIFF
--- a/src/aind_data_schema/data_description.py
+++ b/src/aind_data_schema/data_description.py
@@ -7,7 +7,7 @@ from datetime import datetime
 from enum import Enum, EnumMeta
 from typing import Any, List, Optional
 
-from pydantic import Field
+from pydantic import Field, validator
 
 from aind_data_schema.base import AindCoreModel, AindModel, BaseName, BaseNameEnumMeta, PIDName, Registry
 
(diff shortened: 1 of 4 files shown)


alejoe91:
@mekhlakapoor can we assume `data_level` will always be present on the old data description? In that case you can remove the second part of the added test that deletes that field

jtyoung84:
> @mekhlakapoor can we assume `data_level` will always be present on the old data description? In that case you can remove the second part of the added test that deletes that field

@alejoe91 In the case that the data_level field is missing in the old json file, a user will need to set the data_level explicitly in the upgrader instead of falling back on a default value.

jtyoung84 (approved):

PR #573, opened 2023-10-31
Feat 505 photometry

### LINKED ISSUE #505: Add fiber photometry items to new rig/session
For manipulator based fiber photometry.

saskiad:
closes #505


saskiad:
@dyf @jsiegle The thing I need help on is finding good names that can disambiguate implanted fiber optics and inserted fiber optics. I don't think I've got that down yet, so suggestions on that would be helpful

jsiegle:
It looks like these are the names that currently exist:

#### Rig or Procedure JSON
`FiberProbe` - an optical fiber that can be inserted acutely or chronically implanted
`FiberAssembly` - a combination of a `FiberProbe` and `Manipulator` used in acute experiments
`Patch` - a patch cord that delivers light to/from a `FiberProbe`

#### Session JSON

`FiberModule` - includes information about the position and targeted structure of a `FiberAssembly`; mapped to the rig JSON by name
`FiberPhotometryAssembly` - specifies a connection between a `Patch` and `FiberProbe` object; `Patch` is mapped to the rig JSON by name, while `FiberProbe` can exist in either the rig or procedure JSON.

The distinction between acute vs. chronic currently depends on whether the `FiberProbe` is tracked in the rig or procedure JSON. I don't think there needs to be any change to the name to reflect this.

It is a bit confusing that `FiberAssembly` and `FiberPhotometryAssembly` are different things. Maybe the latter can change to `FiberConnection`, because it's specifying a connection between a `FiberProbe` and `Patch`, which will be session specific?

src/aind_data_schema/session.py
```diff
@@ -191,6 +172,12 @@ class EphysModule(ManipulatorModule):
     ephys_probes: List[EphysProbe] = Field(..., title="Ephys probes used in this module")
 
 
+class FiberModule(ManipulatorModule):
+    """Inserted fiber photometry probe recorded in a stream"""
+
+    fibers: List[str] = Field(..., title="Fiber optics probe name (must match rig JSON)")
+
```
jsiegle, line 179: Add `patch_cords : List[Patch]` to specify which patch cords are used for this module

saskiad: ah, thanks - good catch!

saskiad: what do you think about using the FiberPhotometryAssembly here. The key difference is that it also have the output power of the patch cord - would that be helpful here? And it might reduce some schema clutter.

PR #599, opened 2023-11-09
Implement `BaseModelUpgrade` and `ProcessingUpgrade`

### LINKED ISSUE #552: How to update an existing `Processing` from an old schema version?
**Describe the bug**
The `Processing` schema has changed recently. In the ephys pipeline, we used to update the existing `data_processes` of an existing `Processing` object with the ephys-related steps:

```
existing_processing.data_processes.append(ephys_data_processes)
```

The new schema doesn't expose the `data_processes` directly at the `Processing` level, but they are instead part of the `PipelineProcess`.

The problem is that if a `processing.json` is generated with an old schema, the fields are incompatible. How to solve this?

**To Reproduce**

Try to append data processes to a `Processing` generated with schema version 0.2.5 with the ew 0.3.0 schema version. Note that currently our data transfer scripts are generating 0.2.5 processing files. 

**Expected behavior**
There is a clear guideline / API function to upgrade and merge old schemas. 
Ideally, we could have a `ProcessingUpgrade` class in the `schema_upgrade` module which takes care of this, similar to the [DataDescriptionUpgrade](https://github.com/AllenNeuralDynamics/aind-data-schema/blob/main/src/aind_data_schema/schema_upgrade/data_description_upgrade.py#L77).

alejoe91:
Fixes #552 by:

- Refactoring `schema_upgrader` with a `BaseModelUpgrade`
- `DataDescriptionUpgrade` and `ProcessingUpgrade` inherits from this base class


jtyoung84 (changes requested):
A couple of small notes:

- We're trying to be consistent with using absolute imports: 
```
.base_upgrade -> aind_data_schema.schema_upgrade.base_upgrade
```
- Not a huge deal, but you can use use abstract base classes `from abc import ABC` to make a class as abstract (`class BaseModelUpgrade(ABC)`) and use a decorator on the abstract methods that are expected to be implemented.

alejoe91:
> A couple of small notes:
> 
> * We're trying to be consistent with using absolute imports:
> 
> ```
> .base_upgrade -> aind_data_schema.schema_upgrade.base_upgrade
> ```
> 
> * Not a huge deal, but you can use use abstract base classes `from abc import ABC` to make a class as abstract (`class BaseModelUpgrade(ABC)`) and use a decorator on the abstract methods that are expected to be implemented.

done!

jtyoung84 (approved):

PR #600, opened 2023-11-10
Feat 563 remove overwritten inits

### LINKED ISSUE #563: remove overwritten __init__ functions in models
There are many `__init__` functions in the data_description class, which @jtyoung84 would like removed in favor of validators and properties. 

Most of these could be solved by establishing a `label` function as a property, except for the default `DataDescription`, which utilized the label field to construct the name field. @dyf @saskiad @jtyoung84 should advise on what the goal functionality should be for this model.

dyf:
closes #563 

changes:
- added a field `label` to `DataDescription` which would be used in constructing some names <-- (the logic of these name constructions should be reviewed, and changed to match the desired protocol)
- removed all `__init__` functions from `data_description.py`
- added a validator `build_name` to all models which fulfills the naming conventions that were contained within the respective `__init__` functions
- updated tests to reflect new changes
- EDIT: `build_name` validator will now also check to ensure that `creation_time` was set, and will throw an error if not


src/aind_data_schema/data_description.py
```diff
@@ -307,6 +295,13 @@ def parse_name(cls, name):
             label=m.group("label"),
             creation_time=creation_time,
         )
+    
+    @property  
+    def name(self):
```
dyf, line 300: @Sun-flow this is the issue.  `name` still needs to be a field, not a property. By removing it from the schema, the test data that attempts to construct a `DataDescription` that includes a `name` fails to validate because `name` is not part of the schema.

src/aind_data_schema/data_description.py
```diff
@@ -359,6 +355,14 @@ def parse_name(cls, name):
             creation_time=creation_time,
             input_data_name=m.group("input"),
         )
+    
+    @property
+    def label(self):
```
dyf, line 360: `label` is not a field in the schema - it shouldn't be a property as well. I think you want this to be a normal function `build_label(self)`.

src/aind_data_schema/data_description.py
```diff
@@ -431,20 +435,16 @@ class RawDataDescription(DataDescription):
         const=True,
     )
 
-    def __init__(self, platform, subject_id, **kwargs):
-        """Construct a raw data description"""
+    @property
+    def label(self):
```
dyf, line 439: see above

src/aind_data_schema/data_description.py
```diff
@@ -484,7 +484,7 @@ class AnalysisDescription(DataDescription):
     )
 
     @property
-    def label(self):
+    def make_label(self):
```
dyf, line 487: see above

tests/test_schema_upgrade.py
```diff
@@ -55,8 +55,9 @@ def test_upgrades_0_3_0(self):
             "errors=[{"
             "'loc': ('platform',), "
             "'msg': 'none is not an allowed value', "
-            "'type': 'type_error.none.not_allowed'"
-            "}])"
+            "'type': 'type_error.none.not_allowed'}, "
```
dyf, line 58: `name` should be a field, so `extra fields not permitted` should not be expected here

tests/test_schema_upgrade.py
```diff
@@ -90,7 +91,8 @@ def test_upgrades_0_3_0_wrong_field(self):
             "errors=[{"
             "'loc': ('platform',), "
             "'msg': 'none is not an allowed value', "
-            "'type': 'type_error.none.not_allowed'"
+            "'type': 'type_error.none.not_allowed'}, "
```
dyf, line 94: ditto

jtyoung84 (approved):

src/aind_data_schema/data_description.py
```diff
@@ -431,20 +445,17 @@ class RawDataDescription(DataDescription):
...
 
-        super().__init__(
-            label=f"{platform_abbreviation}_{subject_id}",
-            platform=platform,
-            subject_id=subject_id,
-            **kwargs,
-        )
+        platform_abbreviation = platform.value.abbreviation
+
+        v["name"] = f"{platform_abbreviation}_{v.get('subject_id')}"
```
dyf, line 456: I think we want to timestamp these.  So you probably want `v["name"] = build_data_name(...)` here.

src/aind_data_schema/data_description.py
```diff
@@ -483,11 +494,16 @@ class AnalysisDescription(DataDescription):
...
-    def label(self):
+    @root_validator
+    def build_name(cls, v):
         """returns the label of the file"""
 
-        return f"{self.project_name}_{self.analysis_name}"
+        project_name = v.get("project_name")
+        analysis_name = v.get("analysis_name")
+
+        v["name"] = f"{project_name}_{analysis_name}"
```
dyf, line 504: I think we should timestamp these, so you should use `build_data_name` here.

src/aind_data_schema/data_description.py
```diff
@@ -360,6 +361,19 @@ def parse_name(cls, name):
...
         )
 
+    @root_validator
+    def build_name(cls, v):
+        """returns the label of the file"""
+
+        process_name = v.get("process_name")
+
+        if process_name:
+            v["name"] = f"{v.get('input_data_name')}_{process_name}_{v.get('creation_time')}"
```
dyf, line 371: I think both of these need to use `build_data_name`?

dyf (commented):
@Sun-flow Some comments about `build_data_name`.

src/aind_data_schema/data_description.py
Sun-flow: I've added the timestamp, and also expanded the validator so that `build_name` will check to ensure that `creation_time` was included in construction

Issue #593, opened 2023-11-04
Add Spinnaker as manufacturer


saskiad:
- [ ] no ror ID found. Confirm
- [ ] add to camera list

Issue #592, opened 2023-11-04
add path for schematic/CAD diagrams for custom devices


saskiad:
in device class.

Might also want a similar thing for rig schema.

saskiad:
probably the most effective way to do this is to have a public database of cad drawings for custom parts. We should discuss with hardware people - maybe have some open hardware list or something that we can point to

Issue #590, opened 2023-11-04
Add mouse platform to stream class in session


saskiad:
also add a boolean for active platform (eg. when treadmill is being actively manipulated vs a passive platform)

Issue #589, opened 2023-11-04
Software


saskiad:
Make software in behavior stimulation class accept a list.
Consider a controlled vocabulary for software?

Sun-flow:
@saskiad do you have any further thoughts on this controlled vocab?

saskiad:
this ticket is already close

PR #603, opened 2023-11-14
Feat behavior items

### LINKED ISSUE #594: add enclosure to rig/instrument
include fields for:

- [x] material (enum?)
- [x] size
- [x] grounded = boolean

@jsiegle @adamkglaser What do you think of including this in the rig/instrument schemas? Are there other fields you would want?

saskiad:
closes #594 #593 #592 #591 #590 #589 #578


src/aind_data_schema/stimulus.py
```diff
@@ -92,12 +92,12 @@ class BehaviorStimulation(AindModel):
     )
     behavior_name: str = Field(..., title="Behavior name")
     session_number: int = Field(..., title="Session number")
-    behavior_software: str = Field(
+    behavior_software: List[str] = Field(
```
dyf, line 95: instead of having separate fields for software and version, should we use our existing `Sotware` model?

https://github.com/AllenNeuralDynamics/aind-data-schema/blob/e23f69ed1d18f9f148cfa02ae0c67cb7c9def400/src/aind_data_schema/device.py#L100

e.g. `behavior_software = List[Software]`

dyf (changes requested):

dyf (approved):

PR #607, opened 2023-11-15
Feat 588 daq validator

### LINKED ISSUE #588: Validator for device_name of DAQDevices in rig and instrument schema
When channels are specified on the DAQDevice, one field is the device_name which needs to point to the name of another existing device in the rig. We need a validator that checks and makes sure the names match.
Josh wrote something like this in the ephys_rig schema that we can probably replicate

Sun-flow:
closes #588 

Adds validators to rig and session schema for DAQ channels to ensure that if a DAQ is connected, the port it is connected to is a named and existing device in the schema.


src/aind_data_schema/rig.py
```diff
@@ -97,6 +95,63 @@ class Rig(AindCoreModel):
...
+        cameras = values.get("cameras")
+        ephys_assemblies = values.get("ephys_assemblies")
+        laser_assemblies = values.get("laser_assemblies")
+        mouse_platform = values.get("mouse_platform")
+        stimulus_devices = values.get("stimulus_devices")
+        stick_microscopes = values.get("stick_microscopes")
+        light_sources = values.get("light_sources")
+        patch_coords = values.get("patch_cords")
+        detectors = values.get("detectors")
+        objectives = values.get("objectives")
```
saskiad, line 114: we can probably remove objectives, filter, lens as I can't imagine a way that they can connect to the daq. Probably a few more too. 
At the same time, there's sometimes value in being overcomplete.

src/aind_data_schema/rig.py
```diff
@@ -97,6 +95,63 @@ class Rig(AindCoreModel):
...
+        digital_micromirror_devices = values.get("digital_micromirror_devices")
+        polygonal_scanners = values.get("polygonal_scanners")
+        pockels_cells = values.get("pockels_cells")
+        additional_devices = values.get("additional_devices")
+        daqs = values.get("daqs")
+
+        if daqs is None:
+            return values
+
+        for device_type in [daqs, stimulus_devices, light_sources, patch_coords,detectors, objectives, filters, lenses, digital_micromirror_devices, polygonal_scanners, pockels_cells, additional_devices]:
```
saskiad, line 126: is there a reason to build this around device_type? Can we not have this be 
`for device in [stimulus_devices, light sources]...`
maybe that won't iterate properly ... because we want each device within each of these classes of device_types... is that right?

src/aind_data_schema/rig.py
```diff
@@ -97,6 +95,63 @@ class Rig(AindCoreModel):
...
     @root_validator
+    def validate_device_names(cls, values):  # noqa: C901
+        """validate that all DAQ channels are connected to devices that
+        actually exist
+        """
+
+        device_names = []
+
+        cameras = values.get("cameras")
+        ephys_assemblies = values.get("ephys_assemblies")
```
saskiad, line 106: I think we want to remove the assemblies, assemblies consist of specific devices and it would be the device that is connected to the daq

src/aind_data_schema/rig.py
```diff
@@ -97,6 +95,63 @@ class Rig(AindCoreModel):
...
+        if daqs is None:
+            return values
+
+        for device_type in [daqs, stimulus_devices, light_sources, patch_coords,detectors, objectives, filters, lenses, digital_micromirror_devices, polygonal_scanners, pockels_cells, additional_devices]:
+            if device_type is not None:
+                device_names += [device.name for device in device_type]
+
+        for camera_device in [cameras, stick_microscopes]:
+            if camera_device is not None:
+                device_names += [camera.camera.name for camera in camera_device]
```
saskiad, line 132: this confuses me - why isn't it device.name?

saskiad: ditto below

src/aind_data_schema/rig.py
Sun-flow: what you are describing is what this loop does.

I grouped all fields that have a list of devices nested inside of them, so the first loop moves through each larger class, and the second loop (the list comprehension) iterates through each device in the nested structure.

src/aind_data_schema/rig.py
```diff
@@ -97,6 +95,63 @@ class Rig(AindCoreModel):
...
+
+        if ephys_assemblies is not None:
+            device_names += [probe.name for ephys_assembly in ephys_assemblies for probe in ephys_assembly.probes]
+
+        if laser_assemblies is not None:
+            device_names += [laser.name for laser_assembly in laser_assemblies for laser in laser_assembly.lasers]
+
+        if mouse_platform is not None:
```
saskiad, line 140: why isn't mouse_platform in line 126?

Sun-flow: Mouse platform does not require iteration, it is just a single element

saskiad (changes requested):
looks close - a few questions

src/aind_data_schema/imaging/instrument.py
```diff
@@ -150,6 +150,55 @@ class Instrument(AindCoreModel):
...
 
+    @root_validator
+    def validate_device_names(cls, values):  # noqa: C901
+        """validate that all DAQ channels are connected to devices that
+        actually exist
+        """
+
+        device_names = []
+
+        optical_tables = values.get("optical_tables")
```
saskiad, line 161: let's remove optical table, enclosure, lenses and fluorescence filters. None of them can be connected to a daq

tests/test_imaging.py
```diff
@@ -5,13 +5,14 @@
...
 
-from aind_data_schema.device import Calibration
+from aind_data_schema.device import Calibration, DAQChannel
 from aind_data_schema.imaging import acquisition as acq
 from aind_data_schema.imaging import instrument as inst
 from aind_data_schema.imaging import mri_session as ms
 from aind_data_schema.imaging import tile
 from aind_data_schema.manufacturers import Manufacturer
 from aind_data_schema.processing import Registration
+from aind_data_schema.rig import NeuropixelsBasestation
```
saskiad, line 15: it's a little weird to use the NeuropixelsBasestation in the imaging example since NeuropixelsBasestation cannot be used in an imaging context. Is it possible to use a different DAQ device? I'm willing to write this test if that would help. lmk.

saskiad: My recommendation would to be to use the DAQ from the exaspim example in this test - you don't need all the channels, but if you want to just copy the whole thing you can do that too.

PR #608, opened 2023-11-17
Feat 498 pyd v2

### LINKED ISSUE #498: Update to pydantic 2
There are a few things that would benefit us if we updated to pydantic v2.

- timezone-aware datetime validation: #484 
- simplify BaseName/PIDName metaclasses
- better handling of Literal lists

An early exploration into how hard this would be:
https://github.com/AllenNeuralDynamics/aind-data-schema/pull/397

Notes:
- planning to do this in November

jtyoung84:
Closes #498

### Updating the validators
- v2 removes `validator` and `root_validator` and changes them to `field_validator` and `model_validator` with different interfaces

### Resolving serialization warnings
- There seems to be an issues doing round trip serialization/de-serialization of nested models. I'm looking into this issue further.

### Changes in the example json files:
- Decimal class now gets rendered as strings instead of floats. (This change is unavoidable)
- I changed stuff like `Optional[List[str]] = None` to `List[str] = []` and changed the corresponding null values to empty lists. I can change them back, but this makes things slightly easier to maintain since we don't need to do checks against None. Again, if this isn't the desired behavior, I can change things back
- Everything else remains the same, so from a user standpoint, it shouldn't be too different

### Changes to the schemas
- It seems v2 renders fields like `field_name: Optional[str]` as `anyOf` in the schema. I think in v1 it just marked the field as `"requried": false`. I'm not sure if there's a way to prevent that change.
- This is a work in progress, but it's been difficult to use non-string and non-int enums. I changed all of them to "tagged" classes. For example, modalities and institutions are different. I'm working on making it so the users can still use the same interface (modality=Modality.ECEPHYS), but I need to do some refactoring. For the standard str and int enums, I kept them the same. I set the base model config to use_enum_values=True so the schema just renders the value and not the class name.
- pydantic raises warnings about adding extra information in the fields outside of the description. I removed all instances of "units=" and "additionalProperties=".
- I'll need to explore more how different the changes are between the old v1 and v2 schemas.
- I changed some of the class names that were the same across modules (session.Laser and device.Laser for example). I renamed the stuff in session to something like `LaserConfigs` but I can rename them back to their original name, but it was a bit confusing knowing which was which.

### Changes to the package organization
- I removed the deprecated packages
- I put all the core models into a single package to make it easier to know which classes are AindCoreModels
- I put models that get imported into different packages into a models package. This makes it easier to avoid circular imports and having to use future annotations.


examples/aibs_smartspim_instrument.json
```diff
@@ -25,12 +25,13 @@
          },
          "model": "CleanTop",
          "notes": null,
-         "length": 35,
-         "width": 29,
+         "length": "35",
```
saskiad, line 28: are all number fields now strings?

src/aind_data_schema/core/acquisition.py
```diff
@@ -51,26 +49,27 @@ class Axis(AindModel):
...
-
-        name_lookup = [AxisName.X, AxisName.Y, AxisName.Z]
-
-        axes = []
-        for i, c in enumerate(code):
-            axis = Axis(name=name_lookup[i], direction=direction_lookup[c], dimension=i)
-            axes.append(axis)
-
-        return axes
+    # #  TODO: It might make more sense to put this in the utils package
```
saskiad, line 52: did you move it to utils? or just remove it?

src/aind_data_schema/core/acquisition.py
```diff
@@ -111,13 +113,13 @@ class Acquisition(AindCoreModel):
     specimen_id: str = Field(..., title="Specimen ID")
     subject_id: Optional[str] = Field(None, title="Subject ID")
     instrument_id: str = Field(..., title="Instrument ID")
-    calibrations: Optional[List[Calibration]] = Field(
-        None,
+    calibrations: List[Calibration] = Field(
```
saskiad, line 116: it's unclear to me why making Optional fields required with default of an empty list is preferred over having them be optional. Is there a reason?

src/aind_data_schema/core/instrument.py
```diff
@@ -0,0 +1,96 @@
...
+from datetime import date
+from decimal import Decimal
+from enum import Enum
+from typing import List, Literal, Optional, Union
+
+from pydantic import Field, root_validator
+
+from aind_data_schema.base import AindCoreModel, AindModel
+from aind_data_schema.models.devices import (
+    LIGHT_SOURCES,
```
saskiad, line 12: why is this in caps?

src/aind_data_schema/core/instrument.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+    calibration_data: Optional[str] = Field(
+        None,
+        description="Path to calibration data from most recent calibration",
+        title="Calibration data",
+    )
+    com_ports: List[Com] = Field([], title="COM ports")
+    notes: Optional[str] = None
+
+    # @root_validator
```
saskiad, line 82: we're going to replace this, right?

src/aind_data_schema/core/acquisition.py
jtyoung84: I commented it out currently and haven't decided where it should go. There can be some issues self-referencing the class from a staticmethod. I think I'll probably end up moving the function into the utils package.

src/aind_data_schema/core/acquisition.py
jtyoung84: I'm on the fence about this change. I can revert it back if it's preferred. It's slightly easier to maintain using an empty list as a default, but I can understand why there might be a meaningful difference between "this field has no input" vs. "this field is an empty list"

src/aind_data_schema/core/instrument.py
jtyoung84: Light sources isn't a class, it's a constant referencing a list of classes a user can select from.

src/aind_data_schema/core/instrument.py
jtyoung84: Yeah, I'm going to refactor the validators last. They are going to require the most work since pydantic v2 has a completely different interface.

src/aind_data_schema/core/mri_session.py
```diff
@@ -2,56 +2,17 @@
...
+from pydantic import Field
 
-from aind_data_schema.base import AindCoreModel, AindModel, EnumSubset
-from aind_data_schema.device import Device
-from aind_data_schema.imaging.acquisition import Axis
+from aind_data_schema.base import AindCoreModel, AindModel
+from aind_data_schema.core.acquisition import Axis
+from aind_data_schema.core.procedures import Anaesthetic
 from aind_data_schema.imaging.tile import Scale3dTransform
-from aind_data_schema.procedures import Anaesthetic
```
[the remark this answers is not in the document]
jtyoung84: I moved the core models into a core package, so this import is now on line 11 as `from aind_data_schema.core.procedures import Anaesthetic`

examples/aibs_smartspim_instrument.json
jtyoung84: The Decimal types are now serialized to strings when converted to json in pydantic v2. There's a lengthy discussion about it in this thread: https://github.com/pydantic/pydantic/issues/1511

src/aind_data_schema/core/acquisition.py
jtyoung84: I ended up adding it as a field_validator on the Acquisition class. A user can either input a list of Axis objects or a string of direction codes:
https://github.com/AllenNeuralDynamics/aind-data-schema/blob/c3a9c8d95525cc14ab8071750dbd986bb4fd1df5/src/aind_data_schema/core/acquisition.py#L118

dyf (approved):
There are a few things we need to clean up still (notably, the syntax for specifying PIDNames), but I think this is fine to merge.

examples/ephys_session.json
```diff
@@ -83,38 +85,36 @@
...
       }
    ],
-   "stimulus_epochs": null,
-   "reward_delivery": null,
+   "stimulus_epochs": [],
    "reward_consumed_total": null,
    "reward_consumed_unit": "microliter",
+   "reward_delivery": null,
    "notes": null
 }
```
yosefmaru, line 120: EOF blank line is missing.

examples/exaspim_acquisition.json
```diff
@@ -162,12 +162,12 @@
...
+      "refractive_index": "1.33"
    },
    "sample_immersion": null,
    "active_objectives": null,
    "local_storage_directory": "D:",
    "external_storage_directory": "Z:",
-   "processing_steps": null,
+   "processing_steps": [],
    "notes": null
 }
```
yosefmaru, line 173: EOF blank line

examples/exaspim_instrument.json
```diff
@@ -414,5 +413,6 @@
          "com_port": "COM5"
       }
    ],
-   "notes": null
+   "notes": null,
+   "enclosure": null
 }
```
yosefmaru, line 418: EOF blank line

jtyoung84:
@mekhlakapoor @yosefmaru We can ignore github's new line warnings for the auto-generated json files

mekhlakapoor (approved):

src/aind_data_schema/core/mri_session.py
```diff
@@ -2,56 +2,17 @@
...
-
-
-class Scanner(Device):
-    """Description of a MRI Scanner"""
-
-    device_type: Literal["Scanner"] = Field("Scanner", const=True, readOnly=True)
-    scanner_location: ScannerLocation = Field(..., title="Scanner location")
-    magnetic_strength: MagneticStrength = Field(..., title="Magnetic strength (T)", units="T")
-    magnetic_strength_unit: str = Field("T", title="Magnetic strength unit")
+from aind_data_schema.models.devices import MriScanSequence, Scanner, ScanType
```
saskiad, line 13: I'm not a fan of moving these to devices because they aren't devices.

src/aind_data_schema/models/devices.py
```diff
@@ -305,104 +421,75 @@ class DAQChannel(AindModel):
...
+    channels: List[DAQChannel] = Field([], title="DAQ channels")
 
 
 class HarpDevice(DAQDevice):
     """DAQ that uses the Harp protocol for synchronization and data transmission"""
 
     # required fields
-    device_type: Literal["Harp device"] = Field("Harp device", const=True, readOnly=True)
+    device_type: Literal["Harp device"] = "Harp device"
     harp_device_type: HarpDeviceType = Field(..., title="Type of Harp device")
```
saskiad, line 449: I'm sure it's somewhere, but where is the enum of HarpDeviceType living now? I hvaen't been able to find it (probably how the diff gets shown) but the first place I'd look is with the Harp Device class, so it makes me uncomfortable having it not easily findeable.

src/aind_data_schema/core/procedures.py
```diff
@@ -334,11 +332,11 @@ class InjectionMaterial(AindModel):
     )
     genome_copy: Optional[Decimal] = Field(None, title="Genome copy")
     titer: Optional[Decimal] = Field(None, title="Titer (gc/mL)", description="Titer for viral materials")
-    titer_unit: Optional[str] = Field("gc/mL", title="Titer unit")
+    titer_unit: Literal["gc/mL"] = "gc/mL"
```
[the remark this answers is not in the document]
saskiad: titer might be the exception to this rule though, so we can leave it here, but in general

src/aind_data_schema/core/procedures.py
```diff
@@ -350,29 +348,29 @@ class InjectionMaterial(AindModel):
 class Injection(SubjectProcedure):
     """Description of an injection procedure"""
 
-    injection_materials: List[InjectionMaterial] = Field(
-        None, title="Injection material", unique_items=True, min_items=1
-    )
+    procedure_type: Literal["Injection"] = Field("Injection", title="Procedure type")
```
saskiad, line 351: Injection is never used without being used as part of a different procedure type (Nanoject or Iontophoresis or Retro-Orbital). We want those procedure types to be what is recorded and not "Injection".

src/aind_data_schema/core/procedures.py
```diff
@@ -350,29 +348,29 @@ class InjectionMaterial(AindModel):
...
+    procedure_type: Literal["Retro-orbital injection"] = Field("Retro-orbital injection", title="Procedure type")
+    injection_volume: Decimal = Field(..., title="Injection volume (uL)")
     injection_volume_unit: VolumeUnit = Field(VolumeUnit.UL, title="Injection volume unit")
     injection_eye: Side = Field(..., title="Injection eye")
 
 
 class BrainInjection(Injection):
     """Description of a brain injection procedure"""
 
+    procedure_type: Literal["BrainInjection"] = Field("BrainInjection", title="Procedure type")
```
saskiad, line 373: Ditto above - the procedure type should not be Brain Injection

src/aind_data_schema/core/procedures.py
```diff
@@ -98,12 +76,94 @@ class Fluorophore(Enum):
...
+    """Type of virus preparation"""
+
+    CRUDE = "Crude"
+    PURIFIED = "Purified"
+
+
+class SpecimenProcedure(AindModel):
+    """Description of surgical or other procedure performed on a specimen"""
+
+    specimen_procedure_type: Literal["SpecimenProcedure"] = Field(
```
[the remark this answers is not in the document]
saskiad: wait, we agreed to get rid of specimen_procedure_type, correct? We can do that later, but just so we're on the same page.

src/aind_data_schema/models/species.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class Species:
+    """Species classes"""
+
+    CALLITHRIX_JACCHUS = CallithrixJacchus()
+    HOMO_SAPIENS = HomoSapiens()
+    MACACA_MULATTA = MacacaMulatta()
+    MUS_MUSCULUS = MusMusculus()
+    ALL = tuple(_Species.__subclasses__())
```
saskiad, line 69: Can someone explain how the ALL thing works. We don't want people to choose ALL in these enums. I assume it is so we can use the list of enums somehow, but isn't that what the enum does in the first place?

src/aind_data_schema/core/acquisition.py
saskiad: part of my concern is how users understand what things are required and what are optional. I think this syntax is going to be too subtle.

src/aind_data_schema/core/session.py
```diff
@@ -159,45 +158,48 @@ class ManipulatorModule(DomeModule):
...
-    fiber_connections: List[FiberConnection] = Field(None, title="Fiber photometry devices")
+    fiber_connections: List[FiberConnectionConfig] = Field([], title="Fiber photometry devices")
 
 
-class Laser(AindModel):
+class LaserConfig(AindModel):
     """Description of laser settings in a session"""
 
-    device_type: Literal["Laser"] = Field("Laser", const=True, readOnly=True)
+    config_type: Literal["LaserConfig"] = "LaserConfig"
```
saskiad, line 183: I think we want to keep this as device_type. It's to indicate what type of lightsource is being used. Eg. the name needs to match a name of a laser in the rig schema.

saskiad: ditto for the LightEmittingDiode

src/aind_data_schema/core/session.py
```diff
@@ -45,10 +44,10 @@ class Detector(AindModel):
     trigger_type: TriggerType = Field(..., title="Trigger type")
 
 
-class LightEmittingDiode(AindModel):
+class LightEmittingDiodeConfig(AindModel):
     """Description of LED settings"""
 
-    device_type: Literal["LightEmittingDiode"] = Field("LightEmittingDiode", const=True, readOnly=True)
+    config_type: Literal["LightEmittingDiodeConfig"] = "LightEmittingDiodeConfig"
```
saskiad, line 50: see comment in LaserConfig - we want to stick with device_type here

saskiad:
I've left some comments - take a look. If you want to ignore them for now and turn them into issues to fix, that's fine. They're mostly small. A few might be worth fixing now. But if you do merge, let's turn them into issues so we fix them soon.

src/aind_data_schema/models/devices.py
jtyoung84: I put all of the enum definitions at the top of the devices.py file to avoid potential forward reference conflicts. We can re-organize them to different places though if it's desirable.

src/aind_data_schema/core/acquisition.py
jtyoung84: I can change it to something like calibrations: List[Calibrations] = Field(default = [], etc.) so that users know that there is a default value. Pydantic interprets `Optional` as something like a nullable field

src/aind_data_schema/core/mri_session.py
jtyoung84: I can move the MriScanSequence and ScanType classes to the mri_session module. The Scanner inherits from the Device class though.

src/aind_data_schema/models/manufacturers.py
```diff
@@ -0,0 +1,620 @@
...
+class Vortran(_Manufacturer):
+    """Vortran"""
+
+    name: Literal["Vortran"] = "Vortran"
+    abbreviation: Literal[None] = Field(None)
+    registry: Literal[None] = Field(None)
+    registry_identifier: Literal[None] = Field(None)
+
+
+class CarlZeiss(_Manufacturer):
```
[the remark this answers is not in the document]
jtyoung84: python classes traditionally start with upper case letters

src/aind_data_schema/core/procedures.py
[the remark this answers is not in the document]
jtyoung84: I think since this is how it is in the main branch, it might be better to update it in a future PR.

src/aind_data_schema/core/procedures.py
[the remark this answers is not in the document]
jtyoung84: I'll revert it back to an Optional[str]

src/aind_data_schema/core/procedures.py
jtyoung84: I'll remove the field here

src/aind_data_schema/core/procedures.py
jtyoung84: I'll remove the field here

src/aind_data_schema/core/session.py
jtyoung84: I'll revert the field back to the original

src/aind_data_schema/core/session.py
jtyoung84: I'll revert the field back to the original

src/aind_data_schema/models/species.py
jtyoung84: I imagine we'll need to revisit in a future PR how we want to manage the situation where users can select from a list of non-primitive types. Pydantic doesn't make it too easy. I can add an underscore to the `ALL` to mark it as a protected attribute that won't be exposed to the user. I added it to the classes where a user needs to select OneOf different values. For classes where users don't need to make a selection like that, I didn't include it.

jtyoung84: Also, most of the threads I've read recommended using tagged Unions instead of enums for non-primitive types. In pydantic v1, I added some hacks to the Enum class to get it to work, but I changed things to use tagged Unions for users to select a value from, but that also requires some additional work on our end.

jtyoung84:
> I've left some comments - take a look. If you want to ignore them for now and turn them into issues to fix, that's fine. They're mostly small. A few might be worth fixing now. But if you do merge, let's turn them into issues so we fix them soon.

I pushed a couple of commits that addresses your comments

src/aind_data_schema/models/manufacturers.py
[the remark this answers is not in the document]
jtyoung84: Although I can change this value since this is the value that gets exposed to the user:
```
name: Literal["Carl Zeiss"] = "Carl Zeiss"
```
Should that be something else?

PR #631, opened 2023-11-30
629 rat addition to subject species

miketaormina:
Adding `Rattus norvegicus` to enumerated subject species.


saskiad:
ewplaced by PR #647

PR #650, opened 2023-12-12
Feat 626 tars

### LINKED ISSUE #626: Revisit injection material fields now that TARS is integrated
Eg. the three name fields are a lot.
We want the virus prep alias, the plasmid alias, and when possible, the Addgene RRID...
might want to refer to them as TARS ID rather than name/alias?

Here's a current example:
```
{
                    "name": "AAV-PHP-eB_7x-TRE-tDTomato",
                    "material_id": "AiV300007",
                    "full_genome_name": "AAV-PHP-eB_7x-TRE-tDTomato",
                    "plasmid_name": "AiP300001",
                    "genome_copy": 180000000000,
                    "titer": 15000000000000,
                    "titer_unit": "gc/mL",
                    "concentration": null,
                    "concentration_unit": null,
                    "prep_lot_number": "VT3216g",
                    "prep_date": "2022-02-03",
                    "prep_type": "Purified",
                    "prep_protocol": "SOP#VC003"
}
```

"material_id" is the viral prep lot alias. Let's call this `virus_TARS_id`
"plasmid_name" is the plasmid alias. Let's call this `plasmid_TARS_id`
and there's no place for an external facing RRID (Addgene ID) which we'd ideally have in here.
Let's add `rrid` as a PIDName, Technically, the Addgene ID is a RRID (eg. "RRID:Addgene_198508") ... I feel like we need to go through our TARS entries and add rrid alongside the addgene ids. It's an existing field, just isn't populated.

saskiad:
closes #626

### DIFF
--- a/examples/aibs_smartspim_procedures.json
+++ b/examples/aibs_smartspim_procedures.json
@@ -1,6 +1,6 @@
 {
    "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/core/procedures.py",
-   "schema_version": "0.10.0",
+   "schema_version": "0.10.1",
    "subject_id": "651286",
    "subject_procedures": [],
    "specimen_procedures": [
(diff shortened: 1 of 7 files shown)


saskiad:
this gets the same error as PR#646 eventhough it's a separate branch. No idea what's going on with these.

PR #657, opened 2023-12-15
changed injection materials, update example and test

### LINKED ISSUE #626: Revisit injection material fields now that TARS is integrated

saskiad:
closes #626


src/aind_data_schema/core/procedures.py
```diff
@@ -315,40 +315,50 @@ class Headframe(SubjectProcedure):
...
-    full_genome_name: Optional[str] = Field(
-        None,
+    material_type: Literal["Virus"] = Field("Virus", title="Injection material type")
+    name: str = Field(
+        ...,
         title="Full genome name",
         description="Full genome for virus construct",
     )
-    plasmid_name: Optional[str] = Field(
+    virus_TARS_id: Optional[str] = Field(None, title="Virus ID, usually beings 'AiV'")
```
dyf, line 327: I don't particularly like putting an Allen-internal thing at the top level here. An alternative is to just have `identifiers: List[VirusIdentifier]`, and then:

```python
class VirusIdentifier(BaseModel):
   virus_id: str
   plasmid_id: str
   registry: VirusRegistry # <-- "TARS" goes here
```
addgene could be one of those?

I don't feel too strongly about this.

src/aind_data_schema/core/procedures.py
```diff
@@ -315,40 +315,50 @@ class Headframe(SubjectProcedure):
...
     prep_type: Optional[VirusPrepType] = Field(None, title="Viral prep type")
     prep_protocol: Optional[str] = Field(None, title="Prep protocol")
+    addgene_id: Optional[PIDName] = Field(None, title="Addgene id", description="Registry must be Addgene")
+    genome_copy: Optional[Decimal] = Field(None, title="Genome copy")
+    titer: Optional[int] = Field(None, title="Effective titer (gc/mL)",
+                                 description="Final titer of viral material, accounting for mixture/diliution")
+    titer_unit: str = Field("gc/mL", title="Titer unit")
+
+
+class OtherInjectionMaterial(Reagent):
```
dyf, line 348: I'm not clear why `OtherInjectionMaterial` is a `Reagent` but `ViralMaterial` is not.

dyf: maybe because we need to use the work `titer` instead of `concentration`?

src/aind_data_schema/core/procedures.py
saskiad: we want both Addgene and TARS information for viruses. But I agree that we move the TARS info into a subclass.

src/aind_data_schema/core/procedures.py
saskiad: The problem is that non-viral injection materials and viral injection materials use an orthogonal set of fields and it is really confusing for people to have a bunch of irrelevant fields morphed together like that. This provides clarity.
I agree on name - wasn't thrilled to call it other.

src/aind_data_schema/core/procedures.py
saskiad: because TARS isn't public facing I'm reluctant to make it a Registry

src/aind_data_schema/core/procedures.py
```diff
@@ -315,27 +315,16 @@ class Headframe(SubjectProcedure):
     well_type: Optional[str] = Field(None, title="Well type")
 
 
-class InjectionMaterial(AindModel):
-    """Description of injection material"""
+class TarsIds(AindModel):
```
dyf, line 318: Let's call this `TarsVirusIdentifiers` in case we need other stuff from TARS in the future?

src/aind_data_schema/core/procedures.py
```diff
@@ -315,27 +315,16 @@ class Headframe(SubjectProcedure):
...
-        description="Short name used to reference the plasmid",
-    )
-    genome_copy: Optional[Decimal] = Field(None, title="Genome copy")
-    titer: Optional[Decimal] = Field(None, title="Titer (gc/mL)", description="Titer for viral materials")
-    titer_unit: Optional[str] = Field("gc/mL", title="Titer unit")
-    concentration: Optional[Decimal] = Field(None, title="Concentration", description="Must provide concentration unit")
-    concentration_unit: Optional[str] = Field(None, title="Concentration unit")
-    prep_lot_number: Optional[str] = Field(None, title="Preparation lot number")
+    prep_lot_number: str = Field(..., title="Preparation lot number")
     prep_date: Optional[date] = Field(
```
dyf, line 328: should `pred_date` and `prep_protocol` be on `ViralMaterial`? I can imagine wanting to know these for non-TARS viruses.

Also wondering if `prep_lot_number` is a generally useful field, not TARS-specific.

dyf (changes requested):
couple questions

src/aind_data_schema/core/procedures.py
saskiad: I don't think we'll know `prep_date` and `prep_protocol` for viral prep lots that we don't make ourselves.

same for `prep_lot_number`. I can try to ask - that is one that might be more likely to have - but I'm guessing when we get viruses ready to inject we don't have that information...

dyf:
@saskiad and I discussed pausing this one to get confirmation from AIBS that this is a good way to describe viruses.

saskiad:
@dyf I just went over the `ViralMaterial`  and related classes with Bosiljka and she was happy with how this is organized. I might have forgotten some questions we had but I'd like to get this wrapped up so remind me if there are other things

dyf (approved):

Issue #622, opened 2023-11-27
Remove Specimen_procedure_type and use procedure_type instead


saskiad:
### Discussed in https://github.com/AllenNeuralDynamics/aind-data-schema/discussions/618

<div type='discussions-op-text'>

<sup>Originally posted by **saskiad** November 21, 2023</sup>
In the specimen procedures, we already had a procedure type, and the new specimen_procedure_type is adding not useful information. Is it possible to rely on the existing procedure_type and remove the specimen_procedure_type? @jtyoung84 </div>

PR #667, opened 2023-12-19
added Surgery class

### LINKED ISSUE #418: Surgical Procedures
Currently, surgical procedures like Headframe, Nanoinject Injections, and Fiber implants are treated as separate procedures such that for every injection and every fiber I am prompted to enter the date, and time under anesthesia repeatedly. Generally, these are all done as one "Surgical Procedure." To me, it would make more sense to have a level above these individual subject procedures called "Surgery" or "Surgical Procedure" where general info about time, date, anesthesia duration, animal recovery time etc. can be entered only once, then injections, craniotomies, fiber implants can be added to this "Surgical Procedure."

saskiad:
closes #418


saskiad:
also closes #622

saskiad:
this will mess up the metadata service so @mekhlakapoor should take a look too.

dyf (approved):

PR #669, opened 2023-12-19
adding two proc names

arielleleon:
Adding two new process names for the processing.json file:

- DF/F Caclulation
- Event Detection


src/aind_data_schema/models/process_names.py
```diff
@@ -33,6 +33,8 @@ class ProcessName(str, Enum):
     SKULL_STRIPPING = "Skull stripping"
     SPIKE_SORTING = "Spike sorting"
     SPATIAL_TIMESERIES_DEMIXING = "Spatial timeseries demixing"
+    VIDEO_DFF_CALCULATOIN = "Video dF/F calculation"
```
dyf, line 36: I think both of these actually have the fluorescence traces as input I think one of the F's in DF/F is fluorescence. How about we call these `fluorescence_event_detection` and `dff_estimation`?

dyf (changes requested):
I think that

src/aind_data_schema/models/process_names.py
arielleleon: Thanks for the naming help - this is fixed

dyf (approved):
looks good - go ahead and run he linters

PR #671, opened 2023-12-19
feat: updates tagged unions

jtyoung84:
Merges the recent changes to the main branch.

The Annotated Union fields needed updated. The tests highlighted issues related to classes missing the required Literal fields needed for them to be used in tagged Union fields. I created a "class_tag" field that should be renamed to something else. Since there is already a "procedure_type" being used in the generic SpecimenProcedure class, then either the "class_tag" field needs to be called something other than "procedure_type" or the "procedure_type" field in the SpecimenProcedure class needs to be called something else.


dyf (changes requested):
- rename `class_tag` -> `procedure_class`
- hard-code `procedure_class` and `procedure_type` for `SpecimenProcedure` subclasses
- add new values to `SpecimenProcedureName` as needed

jtyoung84:
Is immunostaining and immunolabeling the same thing?

dyf:
Immunostaining is a kind/subset of immunolabeling, I believe.

dyf (approved):

PR #677, opened 2024-01-02
adding optional filename in base

### LINKED ISSUE #676: DerivedDataDescription write_standard_file method writes the wrong filename
**Describe the bug**
I am creating a derived data description from an existing one, however, when I use the write_standard_file method it writes the json with the following filename `derived_data_description.json` when in fact it should be `data_description.json` with the type being derived.

**To Reproduce**
1. Read a data description json.
2. Create an object from the DerivedDataDescription class.
3. Use the `write_standard_file` method to reproduce.

camilolaiton:
I am adding an optional filename parameter in the default_filename method that fixes #676. The repository uses the CamelCase convention for the names of the classes and the method [default_filename](https://github.com/AllenNeuralDynamics/aind-data-schema/blob/main/src/aind_data_schema/base.py#L30) uses this to build the json filename. I personally don't find this wrong but for the derived data description it's an exception of this rule. 

PD: I could have modified the regular expression but having an optional parameter gives us more freedom. Any thoughts?

### DIFF
--- a/src/aind_data_schema/base.py
+++ b/src/aind_data_schema/base.py
@@ -34,19 +34,35 @@ class AindCoreModel(AindModel):
         name = cls.__name__
         return re.sub(r"(?<!^)(?=[A-Z])", "_", name).lower() + cls._FILE_EXTENSION.default
 
-    def write_standard_file(self, output_directory: Optional[Path] = None, prefix=None, suffix=None):
+    def write_standard_file(
+        self,
+        output_directory: Optional[Path] = None,
+        filename: Optional[str] = None,
+        prefix: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


dyf (changes requested):
I think we rather want to have `default_filename` look for the name of the `AindCoreModel` class it inherits from, if that's not too messy.

camilolaiton:
> I think we rather want to have `default_filename` look for the name of the `AindCoreModel` class it inherits from, if that's not too messy.

It was not messy at all, let me know what you think

dyf (approved):

dyf:
@jtyoung84 FYI

Issue #632, opened 2023-11-30
updates to HarpDevice/DaqDevice


saskiad:
- [x] add `firmware` field that uses Software class as optional in DAQ
- [x] add `hardware_version` in DAQ 
- [x] remove harp_device_version in Harp
- [x] Add `harp_device_type_whoAmI` in Harp (id of device type)
- [x] add manufacturers: AIND, Second Order Effects, Champalimaud Foundation (ror=[03g001n57](https://ror.org/03g001n57))
- [x] change `data_interface` to use USB by default, but not to be constant.
- [x] add boolean field `clock_generator`

anything I'm missing @bruno-f-cruz @poofjunior

bruno-f-cruz:
I would change `clock_generator` to `is_clock_generator`.
Also maybe add the field to the harp class `whoAmI`. This is a 4-digit number that fully specifies the device type.

For future reference here's how we are currently defining a device using in harp:

device schema example: https://github.com/harp-tech/device.behavior/blob/main/device.yml 
- The `device`, `whoAmI`, `firmwareVersion`, `hardwareTargets `are probably what we want to track in some of the aforementioned fields

For general harp documentation, here's the new website too:
https://harp-tech.org
and protocol
https://harp-tech.org/protocol/BinaryProtocol-8bit.html

saskiad:
@bruno-f-cruz is whoAmI redundant with `harp_device_type`? Do we really need both if they are redundant?

bruno-f-cruz:
I would rather have the whoami than the name. Makes it way easier to automate code since whoami are guaranteed to be unique and are reserved for each device. I suspect that users will still want the name. So to kill two birds with one stone, I suggest making it an enumerator of int.

bruno-f-cruz:
We do need the whoami and needs to be automatically populated. I can validate devices by whoami (immutable) but not my name (users can change the name of the device). All my workflows validate if the board I am connected to is of the type I expect. 
Also I just thought of a way to automatically instantiate devices that I want to try and this feature would make things so much easier and water tight!

PR #678, opened 2024-01-02
adding olfactometer and stimulus

### LINKED ISSUE #615: Olfactometer
new device:
- [ ] 5 channels: carrier channel + 4 odor channels
- [ ] channel type (odor/carrier)
- [ ] each channel has a flow_range (unit = mL/min)
- [ ] odor vial volume and odorant volume

add as a stimulus device in rig

In session:
- [ ] per channel: odor identity. dilution (what are units?). (flow rate is stimulus data, not metadta)

@bruno-f-cruz am I missing something here?

saskiad:
closes #615

### DIFF
--- a/src/aind_data_schema/models/devices.py
+++ b/src/aind_data_schema/models/devices.py
@@ -807,7 +807,7 @@ class OlfactometerChannel(AindModel):
 
     channel_index: int = Field(..., title="Channel index")
     channel_type: ChannelType = Field(default=ChannelType.ODOR, title="Channel type")
-    flow_capacity: Literal[100,1000] = Field(default=100, title="Flow capacity")
+    flow_capacity: Literal[100, 1000] = Field(default=100, title="Flow capacity")
     flow_unit: str = Field("mL/min", title="Flow unit")
 
 
--- a/src/aind_data_schema/models/devices.py
+++ b/src/aind_data_schema/models/devices.py
@@ -807,7 +807,7 @@ class OlfactometerChannel(AindModel):
 
     channel_index: int = Field(..., title="Channel index")
     channel_type: ChannelType = Field(default=ChannelType.ODOR, title="Channel type")
-    flow_range: set = Field(default={0, 100}, title="Flow range")
+    flow_capacity: Literal[100,1000] = Field(default=100, title="Flow capacity")
     flow_unit: str = Field("mL/min", title="Flow unit")
 
 
--- a/src/aind_data_schema/models/devices.py
+++ b/src/aind_data_schema/models/devices.py
@@ -10,7 +10,7 @@ from typing_extensions import Annotated
 
 from aind_data_schema.base import AindModel
 from aind_data_schema.models.coordinates import RelativePosition, Size3d
-from aind_data_schema.models.manufacturers import IMEC, Manufacturer
+from aind_data_schema.models.manufacturers import InteruniversityMicroelectronicsCenter, Manufacturer
 from aind_data_schema.models.reagent import Reagent
 from aind_data_schema.models.units import (
     FrequencyUnit,
@@ -513,7 +513,7 @@ class NeuropixelsBasestation(DAQDevice):
     # fixed values
     data_interface: Literal[DataInterface.PXI] = DataInterface.PXI
     manufacturer: Annotated[
-        Union[IMEC], Field(default=Manufacturer.IMEC, discriminator="name")
+        Union[InteruniversityMicroelectronicsCenter], Field(default=Manufacturer.IMEC, discriminator="name")
     ]
 
 
--- a/src/aind_data_schema/models/devices.py
+++ b/src/aind_data_schema/models/devices.py
@@ -839,6 +839,7 @@ class Olfactometer(HarpDevice):
 
     device_type: Literal["Olfactometer"] = "Olfactometer"
     manufacturer: Manufacturer.CHAMPALIMAUD
+    harp_device_type: HarpDeviceType.OLFACTOMETER
     channels: List[OlfactometerChannel]
 
 
--- a/src/aind_data_schema/models/devices.py
+++ b/src/aind_data_schema/models/devices.py
@@ -12,7 +12,7 @@ from aind_data_schema.base import AindModel
 from aind_data_schema.models.coordinates import RelativePosition, Size3d
 from aind_data_schema.models.manufacturers import InteruniversityMicroelectronicsCenter, Manufacturer
 from aind_data_schema.models.reagent import Reagent
-from aind_data_schema.models.units import(
+from aind_data_schema.models.units import (
     FrequencyUnit,
     PowerUnit,
     SizeUnit,
--- a/src/aind_data_schema/models/devices.py
+++ b/src/aind_data_schema/models/devices.py
@@ -823,7 +823,7 @@ class OlfactometerChannel(AindModel):
     def validate_channel(self):
         """Validator for channel type"""
 
-        if ChannelType.ODOR == self.channel_type:
+        if self.channel_type == ChannelType.ODOR.value:
             odor_vial_volume = self.odor_vial_volume
             odorant_volume = self.odorant_volume
             if not odor_vial_volume or not odorant_volume:
--- a/src/aind_data_schema/models/devices.py
+++ b/src/aind_data_schema/models/devices.py
@@ -12,7 +12,15 @@ from aind_data_schema.base import AindModel
 from aind_data_schema.models.coordinates import RelativePosition, Size3d
 from aind_data_schema.models.manufacturers import InteruniversityMicroelectronicsCenter, Manufacturer
 from aind_data_schema.models.reagent import Reagent
-from aind_data_schema.models.units import FrequencyUnit, PowerUnit, SizeUnit, SpeedUnit, TemperatureUnit, UnitlessUnit, VolumeUnit
+from aind_data_schema.models.units import(
+    FrequencyUnit,
+    PowerUnit,
+    SizeUnit,
+    SpeedUnit,
+    TemperatureUnit,
+    UnitlessUnit,
+    VolumeUnit
+    )
 
 
 class ImagingDeviceType(str, Enum):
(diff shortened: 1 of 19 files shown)


src/aind_data_schema/models/devices.py
```diff
@@ -787,11 +795,31 @@ class Speaker(Device):
...
+    odorant_volume: Decimal = Field(..., title="Odorant volume")
+    volume_unit: VolumeUnit = Field(VolumeUnit.ML, title="Volume unit")
+
+
+class Olfactometer(HarpDevice):
     """Description of an olfactometer for odor stimuli"""
 
     device_type: Literal["Olfactometer"] = "Olfactometer"
-    position: Optional[RelativePosition] = Field(None, title="Relative position of the monitor")
+    position: Optional[RelativePosition] = Field(None, title="Relative position of the olfactometer")
```
bruno-f-cruz, line 821: I am a bit confused about this. Is there an example of another device with a position? In this case, this is the position of the odor tube end not the olfactometer itself. Similar to a Valve / Reward spout.

src/aind_data_schema/models/devices.py
```diff
@@ -787,11 +795,31 @@ class Speaker(Device):
...
+    CARRIER = "Carrier"
+
+
+class OlfactometerChannel(AindModel):
+    """description of a Olfactometer channel"""
+
+    channel_index: int = Field(..., title="Channel index")
+    channel_type: ChannelType = Field(..., title="Channel type")
+    flow_range: int = Field(..., title="Flow range")
+    flow_range_unit: str = Field("mL/min", title="Flow range unit")
```
bruno-f-cruz, line 811: `channel_type` should have a default of `Odor`
`flow_range` should only allow two values: 0-100 or 0-1000 ml/min (make 0-100 default)
`flow_range_unit` - Can this be a const?

`odor_vial_volume` and `odorant_volume`: should be nullable, since for carrier type, these settings will not apply

bruno-f-cruz: Should there not be a cross-reference between `OlfactometerChannelConfig` and `OlfactometerChannel`?

src/aind_data_schema/models/stimulus.py
```diff
@@ -138,11 +138,27 @@ class PhotoStimulation(AindModel):
     notes: Optional[str] = Field(None, title="Notes")
 
 
+class OlfactometerChannelConfig(AindModel):
+    """Description of olfactometer channel configurations"""
+
+    channel_index: int = Field(..., title="Channel index")
```
bruno-f-cruz, line 144: Wouldnt it make more sense to have the reference to the `OlfactometerChannel` object, instead of the arbitrary index?

dyf (approved):
I think we need a generic stimulation model - lots of somewhat confusing overlap going on here.

bruno-f-cruz (changes requested):

bruno-f-cruz:
> I think we need a generic stimulation model - lots of somewhat confusing overlap going on here.

I think this goes back to one of my initial questions of why not implement abstract base classes in the schemas?

Should we not have:

```python
Stimulation(AindModel):
    pass
BehaviorStimulation(Stimulation)
    pass
```

It would also make extending the current schemas much easier

src/aind_data_schema/models/devices.py
saskiad: pretty much all devices have relative position. eg. monitor etc. 
But yes, we don't care where the olfactometer is - we care where the odor tube is. And we don't actually have any information about the odor tube in the schema at all...

src/aind_data_schema/models/devices.py
saskiad: The index serves as the cross-reference

src/aind_data_schema/models/devices.py
bruno-f-cruz: The easiest would be to change the property name to something more explicit. Maybe `odor_tube_position`?

src/aind_data_schema/models/stimulus.py
saskiad: oh, maybe...

saskiad:
> I think we need a generic stimulation model - lots of somewhat confusing overlap going on here.

issue #679

src/aind_data_schema/models/stimulus.py
saskiad: I don't think it can do this when the OlfactometerChannel is in a different schema

src/aind_data_schema/models/devices.py
```diff
@@ -259,7 +268,7 @@ class Device(AindModel):
     manufacturer: Optional[Manufacturer.ONE_OF] = Field(None, title="Manufacturer")
     model: Optional[str] = Field(None, title="Model")
     path_to_cad: Optional[str] = Field(None, title="Path to CAD diagram", description="For CUSTOM manufactured devices")
-    port_index: Optional[str] = Field(None, title="Port index")
+    port_index: Optional[int] = Field(None, title="Port index")
```
bruno-f-cruz, line 271: I think this should go back to being a string. I can see several cases where this index will not be an integer. Examples include: COM ports, Ip addresses and hashes.

src/aind_data_schema/models/devices.py
```diff
@@ -787,17 +796,57 @@ class Speaker(Device):
...
+    """Olfactometer channel types"""
+
+    ODOR = "Odor"
+    CARRIER = "Carrier"
+
+
+class FlowRange(Enum):
+    """Olfactometer flow ranges"""
+
+    LOW = "0-100 ml/min"
```
bruno-f-cruz, line 809: This should be a Set instead of an Enum. It will make it future proof. You can even call it max range or capacity, if you want to have MaxRange/MaxCapacity: Set[float]

src/aind_data_schema/models/devices.py
```diff
@@ -787,17 +796,57 @@ class Speaker(Device):
...
+    HIGH = "0-1000 ml/min"
+
+
+class OlfactometerChannel(AindModel):
+    """description of a Olfactometer channel"""
+
+    channel_index: int = Field(..., title="Channel index")
+    channel_type: ChannelType = Field(default=ChannelType.ODOR, title="Channel type")
+    flow_range: FlowRange = Field(..., title="Flow range")
```
bruno-f-cruz, line 818: I think it is fair for flow_range can have default 100, has most of the channels are odor channels and will most of the times have this setting. Also, all channels can be set to max 100, but not all channels can be set to 1000.

src/aind_data_schema/models/devices.py
```diff
@@ -787,17 +796,57 @@ class Speaker(Device):
...
+    odor_vial_volume: Optional[Decimal] = Field(None, title="Odor vial volume")
```
bruno-f-cruz, line 819: I am still a bit confused as to why the vial volume and odorant volume are part of the device and not of the stimuli. I don't see the distinction between these properties and, for instance, dilution percentage. Can you clarify? 

In general, I would reserve OlfactometerChannels (and this should probably inherit from an abstract Channel class) for specifying hardware (mostly electronics) I/O-related specs. In my mind, specifying the vial volume is akin to specifying the impedance of a BNC cable we connect to a nidaq board, which I don't see anywhere in the schema either. Seems like adding a bit too much detail at this point, IMO.

Regardless, if you do feel strongly about adding it I would ask to move it to the respective Stimulus class.

tests/test_rig.py
```diff
@@ -282,6 +282,36 @@ def test_validator(self):
...
+                modification_date=datetime.now(),
+                modalities=[Modality.TRAINED_BEHAVIOR],
+                stimulus_devices=[
+                    Olfactometer(
+                        name="Olfactometer",
+                        manufacturer=Manufacturer.OEPS,
+                        model="1234",
+                        serial_number="213456",
+                        harp_device_type="Behavior",
+                        harp_device_version="1",
```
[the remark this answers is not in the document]
saskiad: that's going to be addressed in a separate PR. Issue #632

src/aind_data_schema/models/devices.py
saskiad: this isn't part of this PR

tests/test_rig.py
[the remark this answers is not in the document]
bruno-f-cruz: Makes sense!

src/aind_data_schema/models/devices.py
saskiad: I did this because this is what we decided in our meeting in November. If that's not what you want, we should probably just delete the issue and start over.

src/aind_data_schema/models/devices.py
bruno-f-cruz: Not sure I follow. The diff tracker shows that the typing changed from a string to an integer. Sorry for the confusion.

src/aind_data_schema/models/devices.py
bruno-f-cruz: After talking with @cindypoo, we agreed on dropping both the vial volume and odorant volume.

src/aind_data_schema/models/devices.py
saskiad: I'm confused by this - I didn't touch this part of the code. But I'll do the harp device PR next that we discussed with Sonya so I'll get that cleaned up.

src/aind_data_schema/models/devices.py
```diff
@@ -787,17 +795,35 @@ class Speaker(Device):
...
+    ODOR = "Odor"
+    CARRIER = "Carrier"
+
+
+class OlfactometerChannel(AindModel):
+    """description of a Olfactometer channel"""
+
+    channel_index: int = Field(..., title="Channel index")
+    channel_type: ChannelType = Field(default=ChannelType.ODOR, title="Channel type")
+    flow_range: set = Field(default={0, 100}, title="Flow range")
```
bruno-f-cruz, line 810: Sorry I shoulda have made this point clearer! By set I meant something like:
replace:
``` python
flow_range: set = Field(default={0, 100}, title="Flow range")
```
with
 
``` python
flow_capacity: Literal[100, 1000] = Field(default=100, title="Flow range")
```
This way we don't have to conform to an Enumerator with predefined string values (i.e. Low and High) and we allow additional values to be added to the list if needed later.

saskiad: ah, I see.

bruno-f-cruz (approved):
lgtm! thanks

PR #680, opened 2024-01-03
importing functional validators

Sun-flow:
`field_validator` and `model_validators` were not importing properly, and needed to be pulled from a separate library. 

import statement has been updated.


Sun-flow:
Outdated issue, resolved with Pydantic V2 Upgrade. Closing PR with no merge.

Issue #674, opened 2023-12-20
RewardDelivery doesn't work with rig.Rig


mochic:
**Describe the bug**
[RewardDelivery](https://github.com/AllenNeuralDynamics/aind-data-schema/blob/caf2bf9bba8f8c91fec9970b4821d0d79e5f086f/src/aind_data_schema/models/devices.py#L774) cannot be added because it fails the validation step [here](https://github.com/AllenNeuralDynamics/aind-data-schema/blob/caf2bf9bba8f8c91fec9970b4821d0d79e5f086f/src/aind_data_schema/core/rig.py#L102), due to not being a subclass of device.Device and therefore not having a name field.

**Expected behavior**
RewardDelivery should be ommitted from the validation step or be modified to have a name field.

**Desktop (please complete the following information):**
 - Version: 0.19.0

PR #683, opened 2024-01-04
exclude reward delivery from standard device validation

mochic:
Attempts to address #674

### DIFF
--- a/src/aind_data_schema/models/devices.py
+++ b/src/aind_data_schema/models/devices.py
@@ -774,6 +774,7 @@ class RewardSpout(Device):
 class RewardDelivery(AindModel):
     """Description of reward delivery system"""
 
+    name: str = "Reward delivery"
     device_type: Literal["Reward delivery"] = "Reward delivery"
     stage_type: Optional[MotorizedStage] = Field(None, title="Motorized stage")
     reward_spouts: List[RewardSpout] = Field(..., title="Water spouts")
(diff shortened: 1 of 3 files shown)


saskiad:
I'd prefer to solve this by removing the `device_type` field from `RewardDelivery` (since it's not a device) and adding `name` that can have a default value of "Reward Delivery"

mochic:
@saskiad does this look ok? `device_type` appears to be required by STIMULUS_DEVICES which uses it as a discriminator.

dyf:
If you exclude delivery you won't be able to include reward delivery devices in the DAQ, e.g. the bottom of this:

https://github.com/AllenNeuralDynamics/aind-data-schema/pull/702

saskiad:
I think we need to rework the validator so that stimulus devices isn't listed in standard devices and is handles separately.

src/aind_data_schema/models/devices.py
```diff
@@ -774,6 +774,7 @@ class RewardSpout(Device):
 class RewardDelivery(AindModel):
     """Description of reward delivery system"""
 
+    name: str = "Reward delivery"
```
saskiad, line 777: if we're keeping the device_type then don't add the name.

saskiad (changes requested):
If we're keeping device_type for the Reward delivery then don't add the name

saskiad:
> @saskiad If you exclude delivery from validation you won't be able to include reward delivery subdevices (spouts, solenoids) in the DAQ, e.g. the bottom of this:
> 
> #702

Agreed, removing it completely from validation is not an option. We can rework the validation to handle it better though.

src/aind_data_schema/models/devices.py
saskiad: Oh, nevermind. We need the name. I'd still prefer to change the validation than to add this, but that will take time.

PR #685, opened 2024-01-04
Update processing upgrades for new schema version

### LINKED ISSUE #684: Update schema upgrade for new schema version
**Describe the bug**
The `DataDescriptionUpgrade` and `ProcessingUpgrade` do not take care of recent changes of the models:

- `DataDescription.investigators` must have at least 1 element
- `DataProcess.outputs` cannot be `None`, but it must be an empty dict

**To Reproduce**
Upgrade a `DataDescription` and `Processing` model from old schema versions.

**Expected behavior**
The `Upgrade` classes correctly upgrade these fields.

**EDIT**
As per #690, the `investigators` field should be enforced rather than fixed.

alejoe91:
Fixes #684


jtyoung84:
@saskiad Alessio is updating some code that upgrades old data description files. It appears that he's setting `investigators = [""]` for situations where the investigators are unknown. Is that the desired behavior?

saskiad:
No, we should know who the investigators are for things.

alejoe91:
There are some datasets with this missing information then, and that breaks the data description creation in the processing pipeline

alejoe91:
Just to add some context. Even recent sessions have missing investigators:

e.g.:
- `ecephys_668755_2023-08-29_15-10-28`

jtyoung84:
@alejoe91 There is a Discussion here about this topic more generally: https://github.com/AllenNeuralDynamics/aind-data-schema/discussions/690

alejoe91:
I agree with Saskia. Will remove the automatic "investigators" correction and just leave the processing one ;)

alejoe91:
@jtyoung84 @saskiad this should be ready to merge. I removed the auto-inferred investigators and just fixed the `null` outputs field in processing (should be `dict()` by default).

Let me know if you suggest other changes!

PR #686, opened 2024-01-04
adding image quality assessment process name

camilolaiton:
Including image quality assessment process name. I will be mainly using this as metadata to evaluate image stitching transformations and using no-reference image quality metrics for smartspim datasets. This is a new capsule that will be included in the pipeline.


saskiad:
I think we want a generic quality control process name and not something this granular

saskiad (approved):

PR #703, opened 2024-01-11
feat: ip (intraperitoneal) injection procedure

### LINKED ISSUE #643: Add IP injections to procedures
NOT a Brain Injection, NOT viruses

subject procedure
material - use Reagent?
dose
volume?

yosefmaru:
Closes #643

### DIFF
--- a/tests/test_procedures.py
+++ b/tests/test_procedures.py
@@ -10,6 +10,7 @@ from aind_data_schema.core.procedures import (
     OphysProbe,
     Procedures,
     RetroOrbitalInjection,
+    IntraperitonealInjection,
     SpecimenProcedure,
     Surgery,
     TarsVirusIdentifiers,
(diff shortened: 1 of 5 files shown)


tests/test_procedures.py
```diff
@@ -85,6 +86,22 @@ def test_injection_material_check(self):
                             injection_duration=1,
                             recovery_time=10,
                         ),
+                        IntraperitonealInjection(
+                            protocol_id="234",
+                            injection_materials=[
+                                ViralMaterial(
```
saskiad, line 92: Let's use a non-viral material for the example since these injections are usually drug administration rather than viral injections.

saskiad (changes requested):
looks good - one small change. Since we don't expect to use viruses in this context it'd probably be good for the test to look closer to what we expect.

saskiad (approved):
awesome!

PR #705, opened 2024-01-12
Feat 632 harp

### LINKED ISSUE #632: updates to HarpDevice/DaqDevice

saskiad:
closes #632


saskiad:
I think we're going to want a tool for mapping or validating that whoAmI values to harp_device_types so that people don't get them wrong.

bruno-f-cruz (changes requested):
First review. We can iterate Thanks!!!

src/aind_data_schema/models/devices.py
```diff
@@ -427,19 +429,20 @@ class DAQDevice(Device):
...
-    harp_device_type: HarpDeviceType = Field(..., title="Type of Harp device")
-    harp_device_version: str = Field(..., title="Device version")
-
-    # fixed values
     manufacturer: Manufacturer.DAQ_DEVICE_MANUFACTURERS = Field(default=Manufacturer.OEPS)
-    data_interface: Literal[DataInterface.USB] = DataInterface.USB
+    harp_device_type: HarpDeviceType = Field(..., title="Type of Harp device")
+    harp_device_whoami: int = Field(..., title="Harp device WhoAmI")
+    data_interface: DataInterface = Field(DataInterface.USB, title="Data interface")
+    is_clock_generator: bool = Field(..., title="Is Clock Generator")
```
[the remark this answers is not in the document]
saskiad: We currently have serial number, firmware version, hardware version. So I need to add Core version and Tag version? What's a tag version? (And I guess what's a core version too - I think I have an idea, but I'm not 100% sure)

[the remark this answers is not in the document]
saskiad: I wonder if we want to make an enum that includes both the Harp Device Type and the whoamI so that people who don't know the four number whoami but do know it's a olfactometer can make that easier?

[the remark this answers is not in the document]
bruno-f-cruz: Thats what I suggested above. I should have formated it sorry!

```python
DeviceWhoAmI(strEnum):
	Behavior: 1234
	LoadCells: 1235
	Lickometer: 1235
```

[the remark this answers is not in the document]
bruno-f-cruz: > We currently have serial number, firmware version, hardware version. So I need to add Core version and Tag version? What's a tag version? (And I guess what's a core version too - I think I have an idea, but I'm not 100% sure)

The firmware version seems overcomplicated tho, cant it simply be a string? We just need a major and minor versions.

Core version is the version of the harp core that the device is running. I.e. the Device has an application firmware and core protocol implementation, they are both versioned.

Tag is a concept that we don't yet have in Harp devices but we are reviewing the current version of the protocol to include it soon. It is basically a short string that can be used to `Tag` a release with additional information (e.g. pre-release, github hash that gave rise to it, etc...). I believe we discussed this with Sonya last time.

saskiad:
why does it say versions aren't updated and then when you update them it gives a merge conflict? That doesn't make any sense to me

src/aind_data_schema/models/harp_types.py
bruno-f-cruz, line 1: If we are going down the route of adding a class per device, can we then please add a generic device?
```python
class GenericHarpDevice(_HarpDeviceType):
    whoami: int = Field(..., ge=0, le=9999)
```

src/aind_data_schema/models/devices.py
```diff
@@ -440,19 +429,21 @@ class DAQDevice(Device):
...
     # optional fields
     channels: List[DAQChannel] = Field(default=[], title="DAQ channels")
+    firmware: Optional[Software] = Field(None, title="Firmware")
+    hardware_version: Optional[str] = Field(None, title="Hardware version")
 
 
 class HarpDevice(DAQDevice):
     """DAQ that uses the Harp protocol for synchronization and data transmission"""
 
     # required fields
```
bruno-f-cruz, line 439: If we are not ready to consider #708 , can we add this extra field to this class?

src/aind_data_schema/models/devices.py
```diff
@@ -440,19 +429,21 @@ class DAQDevice(Device):
 
     # optional fields
     channels: List[DAQChannel] = Field(default=[], title="DAQ channels")
+    firmware: Optional[Software] = Field(None, title="Firmware")
```
bruno-f-cruz, line 432: Any chance we can simplify the Firmware field? I am not sure what would I even enter under the required field "name". JUst trying to understand why would firmware and hardware versions be defined differently

bruno-f-cruz (changes requested):

bruno-f-cruz (changes requested):
Check comments above. I got two review requests for some reason...

src/aind_data_schema/models/harp_types.py
saskiad: Are there really generic harp devices? This contradicts my understanding of them.

src/aind_data_schema/models/devices.py
saskiad: sure - this was what we had decided together in our original plan but easy to change.

src/aind_data_schema/models/harp_types.py
bruno-f-cruz: I am worried that we dont have a way to annotate prototype devices and makes us extremely reliant on the data-schemas being updated. Remeber we want to use this to instantiate a task not only to annotate after the fact. @poofjunior might have some thoughts on this too.

src/aind_data_schema/models/devices.py
bruno-f-cruz: Ooops sorry I didnt check the data type you suggested back then. My bad!

src/aind_data_schema/models/harp_types.py
Poofjunior: @saskiad it is possible to create a harp device "ad hoc" by just uploading some firmware to a [Pi Pico](https://www.raspberrypi.com/documentation/microcontrollers/raspberry-pi-pico.html) board. 

We have done this before, and we will plan to do it more frequently to prototype more long-term devices that will later replace them. If useful, here are some promises that I can make about these ad hoc prototypes:
* they will have their own repository
  * example: [harp.pico.ephys-sync](https://github.com/AllenNeuralDynamics/harp.pico.ephys-sync)
  * example: [harp.pico.cam-trigger](https://github.com/AllenNeuralDynamics/harp.pico.cam-trigger)
* they will run a version of Harp Protocol which will enable us to poll for information about its firmware

Here are a couple options for annotating them.
* we call them a "generic harp device."
* we call them a "prototype harp device."
Since these devices will "speak Harp protocol" and have their own repository, most of the remaining fields can be filled in normally.

LMK what you think, or if I can clarify anything!

bruno-f-cruz: I would just leave it as a generic device and not tie it to the Pico. You never know what kind of devices we might need to spin in the future. Regarding the documentation of the devices, it should go elsewhere and not into this class. AFAIK this class is just a glorified enumerator of existing (known) harp devices.

saskiad:
closes #708

src/aind_data_schema/models/harp_types.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class TimestampGeneratorGen3(_HarpDeviceType):
+    """Timestamp Generator Gen 3"""
+
+    name: Literal["Timestamp Generator Gen 3"] = "Timestamp Generator Gen 3"
+    whoami: Literal[1158] = 1158
+
+
+class HarpDeviceType:
```
bruno-f-cruz, line 94: GenericHarpDevice wont validate against the discriminator

bruno-f-cruz (changes requested):

src/aind_data_schema/models/harp_types.py
saskiad: my guess is that we need to have a constant whoami for the generic harp device.

bruno-f-cruz: Not sure I follow. Why do you need a const WhoAmI? The discriminator takes a 'name' not WhoAmI.
I think all we need is to add a line to HarpDeviceType that establishes a default whoami for the generic device

```python
GENERIC_HARP_DEVICE= GenericHarpDevice(whoami=0)
```

Wouldnt this work?

saskiad: I don't know why. I know that when I removed the whoami field as literal that it created a whole slew of errors. You're welcome to hunt it down, but my guess is that the whoami for the generic needs to be a literal

bruno-f-cruz: Submitted a PR with a possible solution. Not sure if you are ok with it?

ephys_rig.json
bruno-f-cruz, line 1: This should probably be discarded before merging to main?

examples/ephys_rig.json
```diff
@@ -494,7 +513,7 @@
       },
       {
          "device_type": "Harp device",
-         "name": "Harp Behavior",
+         "name": "ThermistorPrototypeBoard",
```
bruno-f-cruz, line 516: I used this is a placeholder to test the code, I think the previous default makes more sense!

examples/ephys_rig.py
```diff
@@ -39,9 +40,10 @@
 analog_input = DAQChannel(channel_name="AI0", device_name="Running Wheel", channel_type="Analog Input")
 
 harp = HarpDevice(
-    name="Harp Behavior",
-    harp_device_type="Behavior",
-    harp_device_version="2.1",
+    name="ThermistorPrototypeBoard",
```
bruno-f-cruz, line 43: Same as above, change the default name so people don't get confused.

bruno-f-cruz (changes requested):

ephys_rig.json
bruno-f-cruz, line 1: I really think you want to delete this but I will approve :p

bruno-f-cruz (approved):
Check the comment on the rogue json file

ephys_rig.json
saskiad: which file?

saskiad: I see it

PR #696, opened 2024-01-08
in Subject schema removing MgiAlleleId in favor of PIDName

Sun-flow:
closes #621 

Converts field `Subject.mgi_allele_ids` from taking `MgiAlleleId` to taking `PIDName`, and adds `MouseGenomeInformatics` to `Registry`

PR #716, opened 2024-01-18
Species Upgrades & validators

### LINKED ISSUE #640: Make subject schema support different species better
- [x] make genotype optional
- [x] create a validator so that genotype is required for mouse subjects
- [x] make source required - uses Allen Institute for animals bred in house
- [x] Make the breeding information a specific class: breeding group, maternal id/genotype, paternal id/genotype. 
- - [ ] This has to be optional overall, validator to require when source is Allen Institute.

for macaque and human data we don't have date of birth, but we do have age - which doesn't live in a subject schema, since it's a moving target... But date of birth might need to be optional too. Maybe with validator to require it for mouse/rat subjects.

Sun-flow:
closes #640 

`Subject.source` optional -> REQUIRED
Breeding info removed from `Subject`, added to new class `BreedingInfo`

Add validators:
- `Subject.genotype` required for mouse subjects
- `Subject.breeding_info` required when `Subject.source == Allen Institute`


tests/test_metadata.py
```diff
@@ -60,7 +63,7 @@ def test_invalid_core_models(self):
         metadata_status as INVALID"""
 
         # Invalid subject model
-        d1 = Metadata(name="ecephys_655019_2023-04-03_18-17-09", location="bucket", subject=Subject.model_construct())
+        d1 = Metadata(name="ecephys_655019_2023-04-03_18-17-09", location="bucket", subject=Subject.model_construct()) #errors
```
Sun-flow, line 66: @jtyoung84 Can you look at this and give me input on solving it? `Subject.model_construct()` is not valid because it lacks required fields, and im not sure how I should manifest the same test goal here.

src/aind_data_schema/core/subject.py
```diff
@@ -116,11 +126,18 @@ class Subject(AindCoreModel):
...
+    @model_validator(mode="after")
+    def validate_breeding_info(self):
+        if self.source is Institution.AI:
+            if self.breeding_info is None:
+                raise ValueError("Breeding info should be provided for subjects bred in house")
+
+    def validate_genotype(self):
+        if self.species is Species.MUS_MUSCULUS:
+            if self.genotype is None:
+                raise ValueError("Genotype should be provided for mouse subjects")
```
saskiad, line 143: can you make the error say "Full genotype must be provided for mouse subjects"

tests/test_metadata.py
```diff
@@ -21,6 +22,8 @@ def test_valid_subject_info(self):
             subject_id="123345",
             sex=Sex.MALE,
             date_of_birth="2020-10-10",
+            source=Institution.AIND,
```
saskiad, line 25: make the Institute.AI

tests/test_metadata.py
```diff
@@ -21,6 +22,8 @@ def test_valid_subject_info(self):
...
+            breeding_info=BreedingInfo(),
             genotype="Emx1-IRES-Cre;Camk2a-tTA;Ai93(TITL-GCaMP6f)",
```
saskiad, line 27: this is not actually a full genotype. which I know is subtle, and maybe not necessary for a test, but it's good for us to use it properly. Full genotype has two copies of each allele. So it might be "Emx1-IRES-Cre/wt;Camk2a-tTA/Camk2a-tTA; Ai93(TITL..)/wt" for each allele there needs to be a slash and either the same thing or a wt

tests/test_metadata.py
```diff
@@ -69,6 +72,8 @@ def test_invalid_core_models(self):
             subject_id="123345",
             sex=Sex.MALE,
             date_of_birth="2020-10-10",
+            source=Institution.AIND,
```
saskiad, line 75: Institution.AI

tests/test_metadata.py
```diff
@@ -69,6 +72,8 @@ def test_invalid_core_models(self):
...
+            breeding_info=BreedingInfo(),
             genotype="Emx1-IRES-Cre;Camk2a-tTA;Ai93(TITL-GCaMP6f)",
```
saskiad, line 77: ditto above

saskiad (commented):
great start - a few small comments

src/aind_data_schema/core/subject.py
```diff
@@ -116,11 +126,18 @@ class Subject(AindCoreModel):
...
     wellness_reports: List[WellnessReport] = Field(default=[], title="Wellness Report")
     housing: Optional[Housing] = Field(None, title="Housing")
     notes: Optional[str] = Field(None, title="Notes")
+
+    @model_validator(mode="after")
+    def validate_breeding_info(self):
+        if self.source is Institution.AI:
+            if self.breeding_info is None:
+                raise ValueError("Breeding info should be provided for subjects bred in house")
+
```
jtyoung84, line 139: You need the model_validator annotation here. Although using a field_validator on breeding_info might be more appropriate

src/aind_data_schema/core/subject.py
```diff
@@ -116,11 +126,18 @@ class Subject(AindCoreModel):
...
-    paternal_genotype: Optional[str] = Field(None, title="Paternal genotype")
+    breeding_info: Optional[BreedingInfo] = Field(None, title="Breeding Info")
     wellness_reports: List[WellnessReport] = Field(default=[], title="Wellness Report")
     housing: Optional[Housing] = Field(None, title="Housing")
     notes: Optional[str] = Field(None, title="Notes")
+
+    @model_validator(mode="after")
+    def validate_breeding_info(self):
+        if self.source is Institution.AI:
+            if self.breeding_info is None:
```
jtyoung84, line 137: Most style guides recommend avoiding nested if statements if it can be avoided. You can group this into one.

src/aind_data_schema/core/subject.py
```diff
@@ -116,11 +126,18 @@ class Subject(AindCoreModel):
         description="Any restrictions on use or publishing based on subject source",
         title="Restrictions",
     )
-    breeding_group: Optional[str] = Field(None, title="Breeding Group")
-    maternal_id: Optional[str] = Field(None, title="Maternal specimen ID")
-    maternal_genotype: Optional[str] = Field(None, title="Maternal genotype")
-    paternal_id: Optional[str] = Field(None, title="Paternal specimen ID")
-    paternal_genotype: Optional[str] = Field(None, title="Paternal genotype")
+    breeding_info: Optional[BreedingInfo] = Field(None, title="Breeding Info")
```
jtyoung84, line 129: Since all of the fields in BreedingInfo are defaulted to None, as an alternative we could consider just
```
breeding_info: BreedingInfo = Field(BreedingInfo(), title="Breeding Info")
```
but it isn't a huge deal

src/aind_data_schema/core/subject.py
Sun-flow: If i used a field validator, I don't think I could check the source field for it's value? or could i?

src/aind_data_schema/core/subject.py
Sun-flow: will do

src/aind_data_schema/core/subject.py
jtyoung84: If you do use a model_validator, it requires a `return self`

tests/test_metadata.py
jtyoung84: I'll look into it

src/aind_data_schema/core/subject.py
Sun-flow: I am waiting on @saskiad to give input on what our defaults for that class should be. I would not expect it to be default empty ultimately, I think some/all of the fields would be required. Just a dummy for now so she can have thoughts

src/aind_data_schema/core/subject.py
jtyoung84: You can inspect other fields using a field validator as long as they are listed above the field in the fields list

tests/test_metadata.py
Sun-flow: The error is showing up again, with my most recent round of edits.

src/aind_data_schema/models/institutions.py
```diff
@@ -111,6 +111,7 @@ class Institution:
     NYU = NewYorkUniversity()
     SIMONS = SimonsFoundation()
     _ALL = tuple(_Institution.__subclasses__())
+    SUBJECT_SOURCES = Annotated[Union[AI, JAX], Field(discriminator="name")]
```
jtyoung84, line 114: This need to be the actual classes and not the instances. So something like
`Annotated[Union[AllenInstitute, JacksonLabs], Field(discriminator="name")]`

src/aind_data_schema/core/subject.py
Sun-flow: so there is a hierarchy based on order , okay. if we were using the default `validator(cls, v)` import method, would i find those fields in `cls.field`?

src/aind_data_schema/models/institutions.py
jtyoung84: Let me know if this fixes the error downstream

src/aind_data_schema/core/subject.py
jtyoung84: The function signature looks like (cls, v: (Some input type like str), info: ValidationInfo) -> expected output type

There is an example here: https://docs.pydantic.dev/latest/concepts/validators/#field-validators

jtyoung84: There is also an example for the  "notes" fields in a few classes here: https://github.com/AllenNeuralDynamics/aind-data-schema/blob/main/src/aind_data_schema/core/session.py

src/aind_data_schema/models/institutions.py
Sun-flow: Ok, all other errors are resolved, the error I tagged on originally still exists. I think it will continue to exist as long as we are generating an invalid model for subject.

src/aind_data_schema/models/institutions.py
```diff
@@ -111,6 +111,7 @@ class Institution:
     NYU = NewYorkUniversity()
     SIMONS = SimonsFoundation()
     _ALL = tuple(_Institution.__subclasses__())
+    SUBJECT_SOURCES = Annotated[Union[AllenInstitute(), JacksonLaboratory()], Field(discriminator="name")]
```
jtyoung84, line 114: With the parentheses, just
```
SUBJECT_SOURCES = Annotated[Union[AllenInstitute, JacksonLaboratory], Field(discriminator="name")]
```

tests/test_metadata.py
jtyoung84: I have a rough idea of why it's breaking. I'll add a fix tomorrow. Converting to field validators might help.

src/aind_data_schema/core/subject.py
saskiad: all these fields should be required for the `BreedingInfo` class

src/aind_data_schema/models/institutions.py
saskiad: I'd prefer we not limit the Institutions we can use and have a validator to exclude AIBS and AIND.

saskiad: Mice that are bred at the institute should have Allen Institute as a source and not Allen Institute for Neural Dynamics or Allen Institute for Brain Science.

jtyoung84: The main drawback of using validators is that users won't know which Institutions are allowed for that field until after they've already tried constructing the model. For example, unless we maintain the same list in the metadata-entry-app, the user will see the full list of Institutions in a dropdown menu, and only after they have tried saving the json will they get an error saying that the institution they selected isn't valid.

Sun-flow: @saskiad lets sort out what our ideal implementation is here. I am lightly on Jon's side for this one, as it will make the GUI make most sense. If there was a way to do an exclusive union (ie, all of these items except these items, @jtyoung84), this would be really easy to do without needing to update a list when we add new institutions. Otherwise, i will just make a long list of all of them except AIND and AIBS.

saskiad: Maintaining lists becomes a bottleneck. The reality is most Allen users aren't writing the subject metadata themselves and are pulling it from the service. It's collaborators and external users that are going to need this changed.

jtyoung84: > @saskiad lets sort out what our ideal implementation is here. I am lightly on Jon's side for this one, as it will make the GUI make most sense. If there was a way to do an exclusive union (ie, all of these items except these items, @jtyoung84), this would be really easy to do without needing to update a list when we add new institutions. Otherwise, i will just make a long list of all of them except AIND and AIBS.

It's simple enough to create a list from another list with a couple of items excluded though

jtyoung84: Oh, I think I get what you're saying. This is only for validating that if a mouse isn't bred in-house, then we want to check the institution?

saskiad: The list of Institutions will grow, and each time we add to it we will need to update this enum as well. The validation is that people like to put "AIND" instead of "AI" but our mouse breeding is "AI" and not "AIND". So internal mice need to have the source as "AI". Presumably, json will be generated by the service for these data, which should be doing it correctly, but it would be good to have a validator to make sure if someone does make it by hand they don't do it wrong.
When there are mice from other sources, the service will not be making the json, and people need to specify the Institution that the mouse is from, and if the list of possible Institutions they can enter for this field isn't up to date, it will be a problem for them.

dyf: It is annoying to have two lists, but if we handle this in a validator then the metadata UI won't know about it.

saskiad: I recommend we not have a list then and let people make this mistake. I really don't want to maintain extra lists if we can avoid it and the reality is most Allen subject will be generated using the service that can be made to avoid this mistake.

src/aind_data_schema/core/subject.py
```diff
@@ -92,11 +109,13 @@ class Subject(AindCoreModel):
         description="Genotype of the animal providing both alleles",
         title="Genotype",
     )
-    alleles: List[PIDName] = Field(default=[], title="Alleles", description="Allele names and persistent IDs")
+    species: Species.ONE_OF = Field(..., title="Species")
+    mgi_allele_ids: List[MgiAlleleId] = Field(default=[], title="MGI allele ids")
```
saskiad, line 113: I'd like to keep the field name as alleles and I'd like to continue using a PIDName structure where the MGI is the registry that is being used.

src/aind_data_schema/core/subject.py
```diff
@@ -73,13 +74,29 @@ class Housing(AindModel):
     )
 
 
+class MgiAlleleId(AindModel):
+    """Mouse Genome Informatics IDs for genotype alleles"""
+
+    allele_name: str = Field(..., title="Name")
+    mgi_id: str = Field(..., title="MGI ID")
```
saskiad, line 81: Does this need a class? Can't we use PIDName?

src/aind_data_schema/core/subject.py
Sun-flow: Hmm,  Ithink this came from when someone pushed to my branch........... frustrating. I'll try to track down what our new implementation should be.

saskiad (changes requested):
I think you're undoing the recent changes we made to using PIDNames for alleles, and I don't want to undo that. See PR #696

src/aind_data_schema/core/subject.py
saskiad: PR #696

src/aind_data_schema/core/subject.py
```diff
@@ -73,13 +74,22 @@ class Housing(AindModel):
     )
 
 
+class BreedingInfo(AindModel):
+    """Description of breeding info for subject"""
+
+    breeding_group: str = Field(None, title="Breeding Group")
```
saskiad, line 80: make all these fields required within this class. That manes in needs to have `Field(..., title="")`. If the default is `None` then it's not actually required.

src/aind_data_schema/core/subject.py
```diff
@@ -92,11 +102,13 @@ class Subject(AindCoreModel):
         description="Genotype of the animal providing both alleles",
         title="Genotype",
     )
-    alleles: List[PIDName] = Field(default=[], title="Alleles", description="Allele names and persistent IDs")
+    species: Species.ONE_OF = Field(..., title="Species")
+    alleles: List[PIDName] = Field(default=[], title="MGI allele ids")
```
saskiad, line 106: teeny edit for the alleles:
`title="Alleles", description="Allele names and persistent IDs"

saskiad (changes requested):
we want the fields within BreedingInfo to be required. (the presence/absence of the word Optional is not actually what determines if a field is required or optional ... that's something I've learned recently)

src/aind_data_schema/core/subject.py
```diff
@@ -92,11 +102,13 @@ class Subject(AindCoreModel):
...
         title="Genotype",
     )
+    species: Species.ONE_OF = Field(..., title="Species")
     alleles: List[PIDName] = Field(default=[], title="Alleles", description="Allele names and persistent IDs")
     background_strain: Optional[BackgroundStrain] = Field(None, title="Background strain")
-    source: Optional[Organization.SUBJECT_SOURCES] = Field(
-        None,
-        description="If the subject was not bred in house, where was it acquired from.",
+    breeding_info: Optional[BreedingInfo] = Field(None, title="Breeding Info")
+    source: Organization.ONE_OF = Field(
```
saskiad, line 109: make this
'source: Organization.SUBJECT_SOURCES'

this is a sublist of possible sources

saskiad (commented):
one small edit since we did add a sublist in the Organization class for possible sources for mouse, let's use that.

saskiad (approved):

PR #752, opened 2024-01-31
metadata validator for smartspim

### LINKED ISSUE #547: metadata completeness check for SmartSPIM data
# User story
As a user, I want to ensure metadata is complete when uploading ephys data, so I can avoid uploading data assets with incomplete metadata.

# Acceptance criteria
- [x] Given a user creates a Metadata model, when the data_description is not None and the data_description.platform field is set to SmartSPIM, then an error will be raised if subject, procedures, acquisition, and/or instrument is None. (This can be accomplished through an extra model_validator in the Metadata class)
- [x] Given a user creates a Metadata model, when the data_description is not None and the data_description.platform field is set to SmartSPIM, then an error will be raised if procedures is missing Injection Materials

# Sprint Ready Checklist 
- [x] 1. Acceptance criteria defined 
- [x] 2. Team understands acceptance criteria 
- [x] 3. Team has defined solution / steps to satisfy acceptance criteria 
- [x] 4. Acceptance criteria is verifiable / testable 
- [x] 5. External / 3rd Party dependencies identified
- [x] 6. Ticket is prioritized and sized

# Notes
Does it make sense to move the upload job models into this repo?

## Original issue:
```
Internally collected `SmartSPIM` data are mature enough that we can require that metadata be complete.  

For `SmartSPIM`, this means:
- `data_description.json` exists
- `subject.json` exists for Allen Institute mice
- `procedures.json` exists for Allen Institute mice
- `acquisition.json` exists
- `instrument.json` exists

Acceptance criteria:
- `SmartSPIM` experiment type data cannot be uploaded if the above criteria are not true.
    - "Allen Institute mice" are those that do not have an `AK` prefix in their subject ID.
```

mekhlakapoor:
closes #547

### DIFF
--- a/src/aind_data_schema/core/metadata.py
+++ b/src/aind_data_schema/core/metadata.py
@@ -12,7 +12,7 @@ from aind_data_schema.base import AindCoreModel
 from aind_data_schema.core.acquisition import Acquisition
 from aind_data_schema.core.data_description import DataDescription
 from aind_data_schema.core.instrument import Instrument
-from aind_data_schema.core.procedures import BrainInjection, Procedures, Surgery
+from aind_data_schema.core.procedures import Injection, Procedures, Surgery
 from aind_data_schema.core.processing import Processing
 from aind_data_schema.core.rig import Rig
 from aind_data_schema.core.session import Session
@@ -197,7 +197,7 @@ class Metadata(AindCoreModel):
             and self.data_description.platform == SmartSpim
             and self.procedures
             and any(
-                isinstance(surgery, BrainInjection) and getattr(surgery, "injection_materials", []) == []
+                isinstance(surgery, Injection) and getattr(surgery, "injection_materials", []) == []
                 for subject_procedure in self.procedures.subject_procedures
                 if isinstance(subject_procedure, Surgery)
                 for surgery in subject_procedure.procedures
@@ -223,7 +223,7 @@ class Metadata(AindCoreModel):
             and self.data_description.platform == Ecephys
             and self.procedures
             and any(
-                isinstance(surgery, BrainInjection) and getattr(surgery, "injection_materials", []) == []
+                isinstance(surgery, Injection) and getattr(surgery, "injection_materials", []) == []
                 for subject_procedure in self.procedures.subject_procedures
                 if isinstance(subject_procedure, Surgery)
                 for surgery in subject_procedure.procedures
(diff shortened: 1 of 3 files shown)


src/aind_data_schema/core/metadata.py
```diff
@@ -181,6 +181,25 @@ def validate_metadata(self):
...
+        ):
+            raise ValueError(
+                "Missing some metadata for SmartSpim. Requires subject, procedures, acquisition, and instrument."
+            )
+        if (
+            self.data_description
+            and self.data_description.platform == SmartSpim
+            and (self.procedures and getattr(self.procedures, "injection_materials", None) is None)
+        ):
+            raise ValueError("Procedures is missing injection materials.")
```
saskiad, line 200: do all SmartSpim procedures involve injections? Would it be better to hinge this warning on if there is an Injection procedure in the list of subject procedures?

mekhlakapoor: The injection materials check is based on the 2nd acceptance criteria here: https://github.com/AllenNeuralDynamics/aind-data-schema/issues/547 , but let me know if we should update that criteria.

src/aind_data_schema/core/metadata.py
```diff
@@ -181,6 +181,25 @@ def validate_metadata(self):
...
+        return self
+
```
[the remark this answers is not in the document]
mekhlakapoor: Also was not in the acceptance criteria but I can add that if we want. Should we move this to discussions thread?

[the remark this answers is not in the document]
jtyoung84: I think it should be easy enough to add that check in this PR if we want that

src/aind_data_schema/core/metadata.py
```diff
@@ -195,7 +221,13 @@ def validate_ecephys_metadata(self):
         if (
             self.data_description
             and self.data_description.platform == Ecephys
-            and (self.procedures and getattr(self.procedures, "injection_materials", None) is None)
+            and self.procedures
+            and any(
+                isinstance(surgery, BrainInjection) and getattr(surgery, "injection_materials", []) == []
```
saskiad, line 226: ditto

saskiad (changes requested):
I think we're better off using `Injection` rather than `BrainInjection`. Otherwise I think it looks good.

saskiad (approved):

PR #760, opened 2024-02-09
changed relative position, moved transforms from tile to coordinates

### LINKED ISSUE #728: Add support for COPA coordinates
The relative position class uses x,y,z,pitch,yaw,roll to describe a device's position. Alex Cahoon has created and shared a COPA object position definition format before the winter break which uses:
- Rotation Matrix [3x3] (a-i)​
a,b,c​
d,e,f​
g,h,i​
- Location Matrix [1x3] (xyz)​
x,y,z​

Our rig devices are currently defined in this format. Hopefully changes can be made to the RelativePosition class or a new position class is implemented to support COPA.

saskiad:
closes #728

### DIFF
--- a/src/aind_data_schema/core/rig.py
+++ b/src/aind_data_schema/core/rig.py
@@ -7,6 +7,7 @@ from pydantic import Field, ValidationInfo, field_validator
 from typing_extensions import Annotated
 
 from aind_data_schema.base import AindCoreModel
+from aind_data_schema.models.coordinates import Axis, Origin
 from aind_data_schema.models.devices import (
     LIGHT_SOURCES,
     Calibration,
@@ -48,7 +49,7 @@ class Rig(AindCoreModel):
 
     _DESCRIBED_BY_URL = AindCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/rig.py"
     describedBy: str = Field(_DESCRIBED_BY_URL, json_schema_extra={"const": _DESCRIBED_BY_URL})
-    schema_version: Literal["0.2.11"] = Field("0.2.11")
+    schema_version: Literal["0.2.12"] = Field("0.2.12")
     rig_id: str = Field(..., description="room_stim apparatus_version", title="Rig ID")
     modification_date: date = Field(..., title="Date of modification")
     mouse_platform: MOUSE_PLATFORMS
@@ -76,6 +77,8 @@ class Rig(AindCoreModel):
         title="CCF coordinate transform",
         description="Path to file that details the CCF-to-lab coordinate transform",
     )
+    origin_point: Optional[Origin] = Field(None, title="Origin point for rig position transforms")
+    rig_axes: Optional[List[Axis]] = Field(default=[], title="Rig axes", min_length=3, max_length=3)
     modalities: Set[Modality.ONE_OF] = Field(..., title="Modalities")
     notes: Optional[str] = Field(None, title="Notes")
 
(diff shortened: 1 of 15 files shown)


src/aind_data_schema/core/rig.py
```diff
@@ -76,6 +77,8 @@ class Rig(AindCoreModel):
         title="CCF coordinate transform",
         description="Path to file that details the CCF-to-lab coordinate transform",
     )
+    origin_point: Optional[Origin] = Field(None, title="Origin point for rig position transforms")
```
dyf, line 80: `rig_origin`, or maybe just `origin`?

src/aind_data_schema/models/coordinates.py
```diff
@@ -16,20 +16,70 @@ class CcfVersion(str, Enum):
...
-    coordinate_system: Optional[str] = Field(None, title="Description of the coordinate system used")
+class AxisName(str, Enum):
+    """Image axis name"""
+
+    X = "X"
+    Y = "Y"
+    Z = "Z"
+
+
+class Direction(str, Enum):
```
dyf, line 35: should we rename this `AnatomicalDirection`?

src/aind_data_schema/models/coordinates.py
```diff
@@ -92,3 +142,41 @@ class CcfCoords(AindModel):
...
+        ...,
+        description="Tissue direction as the value of axis increases. If Other describe in notes.",
+    )
+    unit: SizeUnit = Field(SizeUnit.UM, title="Axis physical units")
+
+
+class RelativePosition(AindModel):
+    """Position and rotation of a device in a rig or instrument"""
+
+    device_position: Translation3dTransform = Field(..., title="Device position")
```
dyf, line 173: I think doing `coordinate_transformations: List[...]` like we did with imaging is a good idea. It's important to apply the transforms in a particular order, and that it would make it clear.

src/aind_data_schema/models/coordinates.py
```diff
@@ -92,3 +142,41 @@ class CcfCoords(AindModel):
...
+    position_unit: SizeUnit = Field(SizeUnit.MM, title="Position unit")
```
dyf, line 174: the units are covered in the `Axis` definitions - I think you can get rid of `position_unit` and `angle_unit` here.

src/aind_data_schema/models/coordinates.py
```diff
@@ -92,3 +142,41 @@ class CcfCoords(AindModel):
...
+    device_rotation: Rotation3dTransform = Field(..., title="Device rotation")
+    angle_unit: AngleUnit = Field(AngleUnit.DEG, title="Angle unit")
+    device_reference_point: str = Field(
```
dyf, line 177: `device_origin` to be consistent?

dyf (changes requested):

src/aind_data_schema/models/coordinates.py
saskiad: Alex had suggested the rotation and position be separate, but that might have been because of how we had organized the previous iteration of relative position...

src/aind_data_schema/models/coordinates.py
```diff
@@ -92,3 +142,41 @@ class CcfCoords(AindModel):
...
+        ...,
+        description="Tissue direction as the value of axis increases. If Other describe in notes.",
+    )
+    unit: SizeUnit = Field(SizeUnit.UM, title="Axis physical units")
+
+
+class RelativePosition(AindModel):
+    """Position and rotation of a device in a rig or instrument"""
+
+    device_position_transformations: Union[
```
dyf, line 173: very close, this needs to be a `List` then gtg.

dyf (changes requested):

PR #769, opened 2024-02-14
Feat 757 mri

### LINKED ISSUE #757: Add a few fields to MRI class
in `MRIScan` add:

- [x] `rare_factor` (title= "RARE factor). int
- [x] `vc_orientation` can use `coordinates.Rotation3DTransform`
- [x] `vc_position` can use `coordinates.Translation3DTransform`
- [x] `vc_subject_position` enum prone/supine 
- [x] `resolution`

remove:

- [x] `axes`

I will update as some of these will use classes in the Tile (or coordinates) file. I will detail asap.

saskiad:
closes #757

### DIFF
--- a/src/aind_data_schema/core/mri_session.py
+++ b/src/aind_data_schema/core/mri_session.py
@@ -39,6 +39,7 @@ class SubjectPosition(str, Enum):
 class MRIScan(AindModel):
     """Description of a 3D scan"""
 
+    scan_index: int = Field(..., title="Scan index")
     scan_type: ScanType = Field(..., title="Scan type")
     primary_scan: bool = Field(
         ..., title="Primary scan", description="Indicates the primary scan used for downstream analysis"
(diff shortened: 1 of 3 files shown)


src/aind_data_schema/core/mri_session.py
```diff
@@ -30,30 +29,41 @@ class ScanType(str, Enum):
...
-        ..., title="Voxel sizes", description="Size of voxels in order as specified in axes"
-    )
+    rare_factor: Optional[int] = Field(None, title="RARE factor")
+    echo_time: Decimal = Field(..., title="Echo time (ms)")
+    effective_echo_time: Decimal = Field(..., title="Effective echo time (ms)")
+    echo_time_unit: TimeUnit = Field(TimeUnit.MS, title="Echo time unit")
+    repetition_time: Decimal = Field(..., title="Repetition time (ms)")
+    repetition_time_unit: TimeUnit = Field(TimeUnit.MS, title="Repetition time unit")
+    # fields required to get correct orientation
+    vc_orientation: Rotation3dTransform = Field(..., title="Scan orientation")
```
dyf, line 55: do you want to combine position/orientation/scale into `coordinate_transformations` like we did elsewhere?

saskiad: I don't. The bruker2nifti package uses these specific fields to orient the image in RAS orientation, so I want to keep these as accessible for that as possible

PR #773, opened 2024-02-14
feat: can have space in project name

### LINKED ISSUE #761: Allow spaces in Project name
The `project_name` field in `data_description` doesn't allow spaces. We want this to be able to use the project names from the funding smartsheet which do have spaces. Can we change the regex to allow spaces?

micahwoodard:
fixes #761 
- allow project name to have spaces


micahwoodard:
I changed the NO_SPECIAL_CHARS to allow spaces but can create new variable if that is what's needed

saskiad (approved):

micahwoodard:
Changed to have an additional variable in DataRegex to allow no special characters but space

src/aind_data_schema/core/data_description.py
```diff
@@ -38,6 +38,7 @@ class DataRegex(str, Enum):
     )
     NO_UNDERSCORES = "^[^_]+$"
     NO_SPECIAL_CHARS = '^[^<>:;"/|? \\_]+$'
+    NO_SPECIAL_CHARS_EXCEPT_SPACE = '^[^<>:;"/|? \\_]+$'
```
jtyoung84, line 41: This is the same regex pattern as the line above

jtyoung84 (changes requested):
Can you add a couple of unit tests that test the regex patterns are doing what's expected? You don't need to create entire models; you can pull the regex patterns by `DataDescription.NO_SPECIAL_CHARS`, etc.

micahwoodard:
Added some test for the different patterns but may need to add or change to better test

jtyoung84 (approved):

PR #782, opened 2024-02-17
refactor stimulus class

### LINKED ISSUE #679: Refactor Stimulus schema
- [ ] BehaviorStimulation as the base class.
- [ ] Need inputs to behavior to allows instances of visual/olfactory/opto conditions
- [ ] clean up the Opto stim class - much of this information doesn't belong in metadata and should be in the NWB file instead

saskiad:
closes #679


src/aind_data_schema/models/stimulus.py
```diff
@@ -171,3 +152,31 @@ class StimulusEpoch(AindModel):
         title="Stimulus end time",
         description="When a specific stimulus ends. This might be the same as the session end time.",
     )
+    behavior_name: str = Field(..., title="Behavior name")
```
dyf, line 155: do we need to couple behavior and stimulus this way? should behavior live at a higher/different level?

src/aind_data_schema/models/stimulus.py
```diff
@@ -150,17 +125,23 @@ class OlfactometerChannelConfig(AindModel):
...
+    """Description of an auditory stimulus"""
+
+    tone_frequency: int = Field(..., title="Tone frequency")
+    tone_frequency_unit: FrequencyUnit = Field(FrequencyUnit.HZ, title="Tone frequency unit")
+    tone_intensity: Decimal = Field(..., title="Tone volume")
+    tone_intensity_unit: SoundIntensityUnit = Field(SoundIntensityUnit.DB, title="Tone intensity unit")
+    notes: Optional[str] = Field(None, title="Notes")
 
 
 class StimulusEpoch(AindModel):
```
dyf, line 142: I am wondering if this actually just wants to be `Epochs`

src/aind_data_schema/models/stimulus.py
```diff
@@ -150,17 +125,23 @@ class OlfactometerChannelConfig(AindModel):
 class OlfactoryStimulation(AindModel):
     """Description of a olfactory stimulus"""
 
-    stimulus_type: Literal["Olfactory"] = "Olfactory"
     channels: List[OlfactometerChannelConfig]
+    notes: Optional[str] = Field(None, title="Notes")
+
+
+class AuditoryStimulation(AindModel):
```
dyf, line 132: Should this rather be `AudibleToneStimuluation`? Wondering b/c there are probably other auditory cues and I'm not sure there are any fields shared between them.

Also this feels like an image stimulation somehow.  I wonder if this should look like that.

src/aind_data_schema/models/stimulus.py
```diff
@@ -171,3 +152,31 @@ class StimulusEpoch(AindModel):
...
+        title="Behavior software",
+        description="The software used to control the behavior (e.g. Bonsai)",
+    )
+    script: Software = Field(
+        ...,
+        title="Behavior script",
+        description="provide URL to the commit of the script and the parameters used",
+    )
+    stimulus_modalities: List[StimulusModality] = Field(..., title="Stimulus modalities")
+    olfactory_stimulus: Optional[OlfactoryStimulation] = Field(None, title="Olfactory stimulus")
```
dyf, line 168: Should we be consistent on `stimulus` vs `stimulation`?

src/aind_data_schema/models/stimulus.py
saskiad: behavior and stimulus are interchangeable

saskiad:
when this merges with main, replace Dict with AindGeneric...

bjhardcastle:
A single `StimulusEpoch` can contain many stimuli, so the names and descriptions in these lines that seem to refer to a specific stimulus are confusing:
https://github.com/AllenNeuralDynamics/aind-data-schema/blob/feat-679-Stimulus/src/aind_data_schema/core/session.py#L387-L397

I think replacing "stimulus" with "epoch" would be clearer (or for the names just have `start_time`/`end_time`/`name`).

bjhardcastle:
Thinking about epochs for recording spontaneous activity, with no stimulus modality per se:
can `stimulus_modalities: List[StimulusModality] = Field(..., title="Stimulus modalities")` be an empty list?

https://github.com/AllenNeuralDynamics/aind-data-schema/blob/feat-679-Stimulus/src/aind_data_schema/core/session.py#L409

dyf (approved):

saskiad:
> Thinking about epochs for recording spontaneous activity, with no stimulus modality per se: can `stimulus_modalities: List[StimulusModality] = Field(..., title="Stimulus modalities")` be an empty list?
> 
> https://github.com/AllenNeuralDynamics/aind-data-schema/blob/feat-679-Stimulus/src/aind_data_schema/core/session.py#L409

`StimulusModality` can be `NONE` 
Which really should only be used if there is truly NO stimulus. Spontaneous activity when the visual monitor is held at a constant value should have that defined.

PR #795, opened 2024-02-26
791 add fields to fieldofview in session

arielleleon:
Added new fields to the FieldOfView session object required for multiplane sessions.


src/aind_data_schema/core/session.py
```diff
@@ -73,6 +77,14 @@ class FieldOfView(AindModel):
     frame_rate: Optional[Decimal] = Field(None, title="Frame rate (Hz)")
     frame_rate_unit: FrequencyUnit = Field(FrequencyUnit.HZ, title="Frame rate unit")
     coupled_fov_index: Optional[int] = Field(None, title="Coupled FOV", description="Coupled planes for multiscope")
+    power: Optional[Decimal] = Field(None, title="Power (mW)")
+    power_unit: Optional[PowerUnit] = Field(PowerUnit.MW, title="Power unit")
```
jtyoung84, line 81: I think we've been keeping the Unit fields as non-Optional and set with a default

src/aind_data_schema/core/session.py
```diff
@@ -73,6 +77,14 @@ class FieldOfView(AindModel):
...
+    scanfield_z: Optional[int] = Field(
+        None,
+        title="Z stage position of the fastz actuator for a given targeted depth",
+    )
+    scanfield_z_unit: Optional[SizeUnit] = Field(None, title="Z stage position unit")
```
jtyoung84, line 86: I think we've been keeping the Unit fields as non-Optional and set with a default

jtyoung84 (changes requested):
The examples in the examples folder may need to be updated

src/aind_data_schema/core/session.py
arielleleon: @jtyoung84  Got it and fixed

arielleleon:
@jtyoung84 Examples have been added

jtyoung84 (approved):

arielleleon:
After meeting with @saskiad , I fixed / added the following:

- using coupled_fov_index
- removed scanimage_resolution for the provided fov_scale_factor
- default power for fov set to PERCENT by default

examples/multiplane_ophys_session.py
```diff
@@ -0,0 +1,212 @@
...
+                    fov_coordinate_ml=1.5,
+                    fov_coordinate_ap=1.5,
+                    fov_coordinate_unit=SizeUnit.UM,
+                    fov_reference="Bregma",
+                    fov_width=512,
+                    fov_height=512,
+                    fov_size_unit=SizeUnit.UM,
+                    magnification="10x",
+                    fov_scale_factor=0.78,
+                    fov_scale_factor_unit=SizeUnit.UM,
```
saskiad, line 79: fov_scale_factor_unit should be um/pixel, correct?

saskiad (changes requested):
Minor thing with the `fov_scale_factor_unit`

examples/multiplane_ophys_session.py
arielleleon: Good catch - this is fixed now

saskiad (approved):

PR #800, opened 2024-02-27
fix error in ephys_rig.json

bendichter:
"rig_axes: []" violates the schema, which specifies a minimum length of 3. The underlying issue is the model, which has [] as a default where the default should instead be `None`.


saskiad:
covered in #802

PR #803, opened 2024-02-27
Feat 802 optional

### LINKED ISSUE #802: Validation issues with optional fields
We have a couple of instances where the default of None of [] is creating validation errors in the `Rig` class. 
Right now we've isolated:

- [ ] `rig_axes` where I expect the `min_length=3` probably doesn't like an empty list
- [ ] in `camera`/`detector` having None for `cooling` which takes the enum

These errors don't show up in pydantic (e.g. we can create our examples with these things), but is creating problems with validation to the json_schema.

saskiad:
closes #802


src/aind_data_schema/models/devices.py
```diff
@@ -193,6 +193,7 @@ class Cooling(str, Enum):
 
     AIR = "Air"
     WATER = "Water"
+    NONE = None
```
jtyoung84, line 196: Since this is a string enum, the value might need to be "None"

saskiad: I need something so that cooling can be optional. I think the validation error comes because the default is None, but None isn't an option. I don't think a string of "None" will do the same thing, will it?

jtyoung84: It will have to be set at the field level like:
```python
cooling: Optional[Cooling] = Field(None, title="Cooling")
```

jtyoung84: Alternatively, if you want to avoid nulls, you can do something like:
```python
class Cooling(str, Enum):
    """Cooling medium name"""

    AIR = "Air"
    WATER = "Water"
    NONE = "None"
```
and then
```python
cooling: Cooling = Field(Cooling.NONE, title="Cooling")
```

jtyoung84: Either way should be fine

saskiad:
replaced with #813

Issue #804, opened 2024-02-27
Consider switching rig daq validation implementation


mochic:
It appears that daqs are being validated here: https://github.com/AllenNeuralDynamics/aind-data-schema/blob/c593512d06d7e76a25cb94735d1989ce226dc5e1/src/aind_data_schema/core/rig.py#L85
I'm not sure if this was intended but a field validator doesn't have predictable access to all fields in the info object according to [documentation](https://docs.pydantic.dev/latest/concepts/validators/#field-validators). Could we potentially switch to using a model_validator instead? This also appears to be breaking the validation of my serialized rig data which I was kinda hoping to rely on.

Would happily implement this change to move this along.

mochic:
Resolved in #805

PR #805, opened 2024-02-27
use model validator instead

mochic:
Addresses #804


src/aind_data_schema/core/rig.py
```diff
@@ -78,41 +78,40 @@ class Rig(AindCoreModel):
         description="Path to file that details the CCF-to-lab coordinate transform",
     )
     origin: Optional[Origin] = Field(None, title="Origin point for rig position transforms")
-    rig_axes: Optional[List[Axis]] = Field(default=[], title="Rig axes", min_length=3, max_length=3)
+    rig_axes: Optional[List[Axis]] = Field(None, title="Rig axes", min_length=3, max_length=3)
```
dyf, line 81: any reason your are changing the default here?

dyf (changes requested):

src/aind_data_schema/core/rig.py
mochic: Yeah. Sorry this is included in this pr and also not in it's own commit. I think that the current default is incorrect and causes future deserialization issues. It is a list that doesn't match the min length requirement for the field. I can remove it as it is out of the scope of this pr.

mochic: Resolved in 1ce710e but I think that this change needs to occur at some point to allow for models to be serialized with the default and then be deserialized with later issues.

mochic:
@dyf I added a new test to supplement the loss of coverage but it does bother me that this change is significant enough that what was covering this in the past no longer is.

dyf:
@mochic I was just looking at this issue - is there a reason you closed it?

mochic:
@dyf I think that the bug I was experiencing was related to something else on my end. It's possible that model validators are more "correct" but I don't think we get anything out of it.

PR #806, opened 2024-02-28
feat:injection materials is optional list

### LINKED ISSUE #765: Injection class materials type should be an List[Annotated[Union[...]]] type
**Describe the bug**
This defintion:
```python
injection_materials: Annotated[
        List[Union[ViralMaterial, NonViralMaterial]],
        Field(title="Injection material", min_length=1, discriminator="material_type"),
    ] = []
```
will lead to incorrect validation behavior. It should be defined as
```python
injection_materials: Optional[List[
        Annotated[
           Union[ViralMaterial, NonViralMaterial], 
           Field(discriminator="material_type")
        ]
    ]] = Field(None, title="Injection material", min_length=1)
```
An Optional tag is needed if the desired behavior is to make the field optional, but if a user creates a list, then the list needs to have at list one item. Otherwise, if the field is required to have a list with at least one item, then the above should suffice.

**Additional context**

micahwoodard:
fixes #765 
- injection material optional 
- if list, the list needs to have at list one item


src/aind_data_schema/core/procedures.py
```diff
@@ -350,10 +350,9 @@ class NonViralMaterial(Reagent):
 class Injection(AindModel):
     """Description of an injection procedure"""
 
-    injection_materials: Annotated[
-        List[Union[ViralMaterial, NonViralMaterial]],
-        Field(title="Injection material", min_length=1, discriminator="material_type"),
-    ] = []
+    injection_materials: Optional[
+        List[Annotated[Union[ViralMaterial, NonViralMaterial], Field(discriminator="material_type")]]
+    ] = Field(None, title="Injection material", min_length=1)
```
jtyoung84, line 355: Is there a unit test that verifies these validate under a few conditions. We should check that the Injection model should:
 - Pass if injection_materials is None
 - Fail if injection_materials is an empty list []
 - Pass if injection_materials is a list with at list one ViralMaterial or NonViralMaterial item

jtyoung84 (changes requested):
It might be helpful to add a unit test to check some validations

micahwoodard:
I added tests for None and empty list. It seemed like there was already a test checking the a list with at list one ViralMaterial or NonViralMaterial item, but I can add one in if you were thinking of something in

jtyoung84 (approved):

saskiad:
For the record: in the long run we don't want injection_materials to be optional. If an injection is performed, we need to know what was injected. If we need to make it optional temporarily because we don't have the systems yet, we can do that, but we will need to revert this in the future

saskiad (approved):

PR #807, opened 2024-02-28
feat: added orcid and invesigator name now pid name

### LINKED ISSUE #762: Add ORCID for Investigators
in `DataDescription`, make `investigator` a list of `PIDName`
Add ORCID to `Registry`. Name = "Open Researcher and Contributor ID", abbreviation = "ORCID"

micahwoodard:
fixes #762 
- add orcid
- invesigators now list of pid names


saskiad:
I'm not keen on making it a union with string. If we aren't ready to use this, let's not do it. As I said when I made the issue, this is not an urgent priority, but something we'll need in the long run. 
But I'm also unclear on how this makes upgrades complicated when almost no data descriptions currently have investigators in them to begin with.

dyf:
@saskiad I agree and we can merge this. The place we've run into trouble is that all the code using `DataDescription` has to change slightly. We ran into this with pipelines writing data descriptions for results breaking when reading stuff out of older versions of raw data descriptions to make derived data descriptions. So let's go ahead and add a ticket to update aind-metadata-upgrader for this change.

PR #812, opened 2024-03-05
add ground_wire procedure

### LINKED ISSUE #776: ground wire position
The ephys team adds a ground wire the the implant and it would be good to have the ground wire position tracked.
This is a string description.
This probably should live in the procedure schema, but it's not a procedure that is requested so I'm not sure how to ingest that information.

saskiad:
closes #776


jsiegle:
Can we add a ground wire material (`str`) field?

saskiad:
> Can we add a ground wire material (`str`) field?

Sure thing. Do you want this to be a free field string or an enum? E.g. similar to HeadframeMaterial

jsiegle:
Tagging @galenlynch – I think we currently just use silver wires, so the enum could be short

galenlynch:
That's right, we just use silver.

galenlynch (commented):
I think this procedure should be renamed for clarity, and some of the fields should be made optional. The thing that must happen with this procedure is that the sortaclear must be removed, and replaced with some material that allows for recordings. We currently use DuraGel, but that could change in the future. Dynamic Routing uses agarose instead.

Concretely, that would mean changing this procedure name to "SortaClearRemoval" or something similar, the "duragel_replacement" field should instead be a "replacement_material" field that accepts a string or enum, the ground_wire fields should all be optional (sometimes the ground wire fails or is not used). I also think there should be a comment field to allow notes about bleeding or holes being blocked etc.

Just to make your life more difficult, we have also removed a ground wire and well, sent a mouse to the MRI, and then attached a second well to replace the first, and inserted a ground wire again. Not sure how to represent these edge cases. Maybe the separate steps should be different procedures, or all steps should be optional? The well removal procedure would be difficult to represent with this class.

saskiad:
@galenlynch take a look now. One note - this procedure will be part of a `Surgery` procedure, so that already provide a note location.
My recommendation for well removal would be to use the `OtherSubjectProcedure` for removals and then have another `ProtectiveMaterialReplacement` surgery when they are replaced.

src/aind_data_schema/core/procedures.py
```diff
@@ -299,6 +307,20 @@ class Headframe(AindModel):
     well_type: Optional[str] = Field(None, title="Well type")
 
 
+class ProtectiveMaterialReplacement(AindModel):
+    """Description of a proective material replacement procedure in preparation for ephys recording"""
+
+    procedure_type: Literal["Ground wire"] = "Ground wire"
+    protocol_id: str = Field(..., title="Protocol ID", description="DOI for protocols.io")
+    protective_material: ProtectiveMaterial = Field(..., title="Protective material")
```
galenlynch, line 315: Is this the material being removed, or the material that it is replaced with?

galenlynch (approved):

src/aind_data_schema/core/procedures.py
saskiad: replaced with - I'll make that clear

dyf (approved):

PR #813, opened 2024-03-06
cooling and rig axes option

### LINKED ISSUE #802: Validation issues with optional fields

saskiad:
closes #802


saskiad:
replaces #803

jtyoung84 (approved):

PR #816, opened 2024-03-06
Acquisition Immersion description

miketaormina:
- added `description` field (optional string) to `Acquisition.Immersion` class
This class is used for both `chamber_immersion` and `sample_immersion`

### DIFF
--- a/src/aind_data_schema/core/acquisition.py
+++ b/src/aind_data_schema/core/acquisition.py
@@ -18,11 +18,6 @@ class Immersion(AindModel):
 
     medium: ImmersionMedium = Field(..., title="Immersion medium")
     refractive_index: Decimal = Field(..., title="Index of refraction")
-    description: Optional[str] = Field(
-        '',
-        title="Description",
-        examples=["Cargille", "EasyIndex", "Ethyl Cinnimate"]
-    )
 
 
 class ProcessingSteps(AindModel):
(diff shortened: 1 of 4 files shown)


src/aind_data_schema/core/acquisition.py
```diff
@@ -18,6 +18,11 @@ class Immersion(AindModel):
 
     medium: ImmersionMedium = Field(..., title="Immersion medium")
     refractive_index: Decimal = Field(..., title="Index of refraction")
+    description: Optional[str] = Field(
```
saskiad, line 21: I'd MUCH prefer to expand the ImmersionMedium to include these options.

miketaormina: I'm open to that as well (I actually thought they were in there already, but maybe my memory is bad). Off the top of my head, this would mean adding
- EASYINDEX
- ETHYLCINNIMATE
- ACB
- OPTICLEAR (OPTIPREP ?)
- CUBICR

Those last three are things that Jayaram's lab is using or developing, so I need to ask Kevin for the correct names.

The Cargille oil can just be "OIL" with the specifics located in the SOP

miketaormina: Maybe it used to be a sting instead of an enumerated list? An acquisition from last year reads `Cargille oil` in the `acquisition.json`

miketaormina: @saskiad I've undone the changes from yesterday and added three new immersion mediums to `devices.ImmersionMedium`

saskiad (approved):

Issue #825, opened 2024-03-12
MRIScan.validate_primary_scan is not returning itself


Sun-flow:
add a line to `return self` at the end of the validator

PR #828, opened 2024-03-12
adding example for MRISession

### LINKED ISSUE #827: generate examples for MRISession

Sun-flow:
closes #827 

Adds an example for `MRISession`, which includes examples of `MRIScan`. Will need to be updated when the validator change in #825 goes through, to include non-null scans.


examples/mri_session.py
```diff
@@ -0,0 +1,59 @@
...
+sess = MriSession(
+    subject_id="",
+    session_start_time="2024-03-12T16:27:55.584892",
+    session_end_time="2024-03-12T16:27:55.584892",
+    experimenter_full_name=["test"],
+    protocol_id="",
+    iacuc_protocol="",
+    mri_scanner=Scanner(
+        name="test_scanner",
+        scanner_location="Fred Hutch",
```
dyf, line 52: let's make up a name

PR #831, opened 2024-03-13
timezones required for times

### LINKED ISSUE #484: Specify timezones for datetimes
For datetimes to be useful, we need to specify the timezone.  

Acceptance criteria:

- [ ] all `datetime` type fields include the timezone when serialized
- [ ] the serialized string is in the timezone local to the client that generated the string. For example: "Jan 1 2023 8am", when generated in Seattle, should look like this: `2023-01-01T08:00:00-07:00`

Notes:
- light cycle should be a `time`, not a `datetime`, as it is not a specific point in time.

micahwoodard:
fixes #484 
- replace datetime fields with AwareDatetime
- requires .astimezone() for datetime inputs


examples/ephys_rig.py
```diff
@@ -170,15 +170,15 @@
 )
 
 red_laser_calibration = Calibration(
-    calibration_date=datetime.datetime(2023, 10, 2, 10, 22, 13),
+    calibration_date=datetime(2023, 10, 2, 10, 22, 13, tzinfo=timezone.utc).astimezone(tz=timezone.utc),
```
jtyoung84, line 173: Does it work if we just do:
```python
calibration_date=datetime(2023, 10, 2, 10, 22, 13, tzinfo=timezone.utc)
```
I think as long as it get's stored with an offset, we should be good.

micahwoodard: oh so leave out the astimezone? I'm not sure I'll check

micahwoodard: Yep that works. Changing accordingly

examples/bergamo_ophys_session.json
```diff
@@ -5,8 +5,8 @@
    "experimenter_full_name": [
       "John Doe"
    ],
-   "session_start_time": "2022-07-12T07:00:00",
-   "session_end_time": "2022-07-12T07:00:00",
+   "session_start_time": "2022-07-12T07:00:00Z",
```
saskiad, line 8: This is probably my ignorance, but I don't understand the Z nomenclature. How does this indicate the timezone?

saskiad (commented):
I'm just confused about how this works - the code looks fine to me, but I want to understand before merging.

jtyoung84:
> I'm just confused about how this works - the code looks fine to me, but I want to understand before merging.

We can add a step to use the locale of the user's computer settings as a default, as per the discussion yesterday. With the current implementation, a user will need to supply the timezone information whenever they create a datetime object, otherwise an error will be raised.
```python
from datetime import datetime, timezone

# Current implementation doesn't add any tzinfo
session_start_time=datetime(2023, 10, 2, 10, 22, 13)

# After this gets merged, a user will need to add this line to get the timezone of their os settings
session_start_time=datetime(2023, 10, 2, 10, 22, 13).astimezone()

# We can add that "astimezone()" as a default if a user just defines
# a datetime object as above. Users shouldn't need to change anything
# in that case. 

# Alternatively, a user can supply a timezone if they know it
session_start_time=datetime(2023, 10, 2, 10, 22, 13, tzinfo=timezone.utc)

# We may need to update the metadata-entry app to add timezone information
```

saskiad:
Can we update the examples to use this to help make it clear?
`session_start_time=datetime(2023, 10, 2, 10, 22, 13).astimezone()`

dyf:
> The main issue is that the unit tests will use different timezone settings when run locally vs. in github actions, which can cause discrepancies in the example.json files if we don't specify a timezone in the example.py files. We can modify the unit tests to ignore the datetime fields if we want though.

IMO having the examples always use UTC will be fine.

saskiad (approved):

jtyoung84 (changes requested):
Add AwareDatetime with Default

jtyoung84 (approved):

PR #834, opened 2024-03-13
Feat 827 mri session example

Sun-flow:
re-pulling on issue #827 due to a bug in `MRIScan` generating nulls for the json


jtyoung84:
It looks like the example was automatically updated. Closing this for now.

PR #866, opened 2024-03-28
Feat 793 sectioning

### LINKED ISSUE #793: Sectioning procedure
We need a specimen procedure class called `Sectioning`.

it needs a field for `output _specimen_ids` which should be a list of strings

saskiad:
closes #793

### DIFF
--- a/src/aind_data_schema/core/procedures.py
+++ b/src/aind_data_schema/core/procedures.py
@@ -87,6 +87,21 @@ class Side(str, Enum):
     RIGHT = "Right"
 
 
+class SectioningOrientation(str, Enum):
+    """Orientation of sectioning"""
+
+    CORONAL = "Coronal"
+    SAGITTAL = "Sagittal"
+    TRANSVERSE = "Transverse"
+
+
+class SliceStrategy(str, Enum):
+    """"""
+
+    WHOLE = "Whole Brain"
+    HEMI = "Hemi Brain"
+
+
 class ProtectiveMaterial(str, Enum):
     """Name of material applied to craniotomy"""
 
@@ -113,6 +128,7 @@ class CoordinateReferenceLocation(str, Enum):
 
     BREGMA = "Bregma"
     LAMBDA = "Lambda"
+    MIDLINE = "Midline"
 
 
 class HeadframeMaterial(str, Enum):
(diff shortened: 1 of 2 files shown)


src/aind_data_schema/core/procedures.py
```diff
@@ -87,6 +87,21 @@ class Side(str, Enum):
...
 
+class SectioningOrientation(str, Enum):
+    """Orientation of sectioning"""
+
+    CORONAL = "Coronal"
+    SAGITTAL = "Sagittal"
+    TRANSVERSE = "Transverse"
+
+
+class SliceStrategy(str, Enum):
```
dyf, line 98: Do we use the term `slice` anywhere else? Should this be `SectioningStrategy`?

src/aind_data_schema/core/procedures.py
```diff
@@ -224,6 +240,33 @@ class Immunolabeling(AindModel):
...
+
+    procedure_type: Literal["Sectioning"] = "Sectioning"
+    number_of_slices: int = Field(..., title="Number of slices")
+    output_specimen_ids: List[str] = Field(..., title="Output specimen ids", min_length=1)
+    sectioning_orientation: SectioningOrientation = Field(..., title="Sectioning orientation")
+    section_thickness: Decimal = Field(..., title="Section thickness")
+    section_thickness_unit: SizeUnit = Field(SizeUnit.MM, title="Section thickness unit")
+    section_distance: Decimal = Field(..., title="Section distance from reference")
+    section_distance_unit: SizeUnit = Field(SizeUnit.MM, title="Distance unit")
+    distance_reference: CoordinateReferenceLocation = Field(..., title="Distance reference")
```
dyf, line 254: The term `distance_reference` was a little confusing to me. Maybe:
```
reference_location: ReferenceLocation
section_reference_distance: Decimal
```

dyf (changes requested):

src/aind_data_schema/core/procedures.py
saskiad: I'm leaving CoordinateReferenceLocation as that is used in other places as well. Changed other.

src/aind_data_schema/core/procedures.py
saskiad: going with `SectionStrategy`

PR #883, opened 2024-04-03
Feat 847 sessionmri

### LINKED ISSUE #847: make MRI part of session

saskiad:
closes #847


saskiad:
this will impact the mri ingest function slightly - just needs to point it to `Session` and add a few extra fields

src/aind_data_schema/core/session.py
```diff
@@ -306,6 +389,8 @@ class Stream(AindModel):
     ophys_fovs: List[FieldOfView] = Field(default=[], title="Fields of view")
     slap_fovs: Optional[SlapFieldOfView] = Field(None, title="Slap2 field of view")
     stack_parameters: Optional[Stack] = Field(None, title="Stack parameters")
+    mri_scanner: Optional[Scanner] = Field(None, title="MRI scanner")
```
dyf, line 392: Do we actually need `Scanner` here? Should that not instead be part of a `Rig`?

dyf (changes requested):

src/aind_data_schema/core/session.py
saskiad: I don't think it makes sense to make a rig.json for a Scanner.

examples/mri_session.json
```diff
@@ -1,105 +1,142 @@
...
-         "processing_steps": [],
-         "additional_scan_parameters": {},
+         "mri_scans": [
+            {
+               "scan_index": 1,
+               "scan_type": "Set Up",
+               "primary_scan": false,
+               "scan_sequence_type": "RARE",
+               "rare_factor": 8,
+               "echo_time": "3.4199999999999999289457264239899814128875732421875",
```
jtyoung84, line 59: This happens when using floats directly instead of `Decimal("3.42")` when constructing these fields

examples/session.json
saskiad, line 1: I'm confused by this file - let's keep the mri example as mri_session.py. I don't think we want a generic session example as there are some many different variants on the sessions.

examples/mri_session.json
Sun-flow: This is weird because it is using Decimal. It is not using decimal as a string? Would that be the issue? `Decimal(3.42)`

saskiad: yeah, I think that's what Jon means. If you feed Decimal a string it won't have that floating point error.

jtyoung84 (approved):

src/aind_data_schema/core/session.py
saskiad, line 392: @Sun-flow let's move the mri_scanner to the `MRIScan` field and then it can be a required field and we don't need a validator for it.

Sun-flow: Let's move this into MRIScan

Sun-flow (approved):

dyf (approved):

dyf:
@Sun-flow go ahead and resolve these merge conflicts and you can merge

PR #887, opened 2024-04-10
fix: removes erd tests from local cov

### LINKED ISSUE #878: Consider skipping `test_diagram_builder.py` test
**Describe the bug**
A new test was added to the test suite that breaks the dev dependencies in the project.toml file.  `test_diagram_builder.py` requires the erdantic package that is only part of the docs build. While in theory one could install erdantic, this package is known to be a bit annoying to install in windows so it does make sense to not add it to the dev dependencies. However, it does mean that the test suite will always fail on that test since the module will not be found.

**To Reproduce**
1. Run the test suite with the dev environment

**Expected behavior**
I expect to be able to run all tests with the dev environment

**Screenshots**

 - OS: win10

**Additional context**
Alternatively one could use a different test module or even a conditional skip (e.g.  https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) that checks for the existence of the module.

jtyoung84:
Closes #878 

- Omits running erdantic tests locally
- Still runs erdantic tests during Pull Requests
- Adds constraint to erdantic version since >=1.0 seems to break things

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,4 +38,4 @@ jobs:
         sudo apt install graphviz libgraphviz-dev -y
         python -m pip install -e .[dev] -e .[docs] --no-cache-dir
     - name: Run tests and coverage
-      run: coverage run -m unittest discover && coverage report
+      run: coverage run --omit "*__init__*" -m unittest discover && coverage report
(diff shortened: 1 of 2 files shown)


Sun-flow (approved):
Looks good. Just to confirm, these tests were not failing locally, but in the PR. Will merge this pull and see if my branch is fixed.

PR #903, opened 2024-04-29
Feat 852 multiplanepower

### LINKED ISSUE #852: Power Recordings for Multiplane Imaging
The mesoscope contains two Pockels cells which modulate the beam via a voltage controlled wave plate. All planes are imaged in pairs and each pair require a different modulation (or percent power) over each of the Pockels cells. The modulation of a laser beam is not linear meaning the power at the output of each Pockels cell needs to be recorded for each fov pair. 

We need to have a way to record both powers and their split ratio for each fov.

saskiad:
closes #852


saskiad:
We should consider adding a final power field, which would require that someone (and this can likely be automated) use the calibration field along with the power and power ratio to calculate what the power at the plane was...

arielleleon (approved):
Yes this looks good. I will make sure the weekly laser calibration is stored in the sesison file so we can back calculate power per plane

PR #904, opened 2024-04-29
adding fip behavior rig example

### LINKED ISSUE #872: add behavior rig template to the example folder
It would be great if you could add the following rig template for the foraging behavior team to the `example` folder and maintain for version updates.
https://github.com/AllenNeuralDynamics/dynamic-foraging-task/blob/develop/src/setting_templates/447_FIPBehaviorOpt_FullModalityTemplate.py

saskiad:
closes #872


hagikent:
Just checking in if it's going forward!
Behavior team would benefit a lot if this template keeps being updated with schema version update.

saskiad:
> Just checking in if it's going forward! Behavior team would benefit a lot if this template keeps being updated with schema version update.

There's an issue with the pydantic version that is making the tests fail. We have an issue to resolve that - When I'm back in Seattle next week I'll find out if we've found a solution to that. I think this example is fine, it's something with how the tests are run.

jtyoung84 (approved):
This is probably fine. We can consider adding this serializer to all "Sets" or to create a custom SortedSet class. There are some libraries like `https://github.com/drivendataorg/sortedcontainers-pydantic` but they don't seem to work with objects like Modality.

hagikent:
@alexpiet 
This is the PR to have SciComp make the rig.json template "catching up" with the ever-changing schema ecosystem.
You might want to replace this previous template prepared by me with newer one. Otherwise, we (the GUI) need to keep using very old schema ver.

PR #906, opened 2024-04-30
feat: added optiopnal list of Software to acquisition

### LINKED ISSUE #539: Logging software versions in acquisition?
This is not related to a problem. Instead an inquiry, see below.

It would be beneficial for SPIM systems (probably all of our systems) to record software information about the given acquisition. For example package versions for imported Python libraries, and for any internal libraries that may be undergoing ongoing development, git hashes. Is this something we already have added into aind-data-schema - or do we think this is worth adding/recording?

We currently stores these in the log files for each scan, so it is recorded. This is more of a question about whether we think this is important to push into aind-data-schema.

**Additional context**
None!

micahwoodard:
fixes #539 
- added optional list of Software to acquisition

### DIFF
--- a/src/aind_data_schema/core/acquisition.py
+++ b/src/aind_data_schema/core/acquisition.py
@@ -78,7 +78,7 @@ class Acquisition(AindCoreModel):
         title="Processing steps",
         description="List of downstream processing steps planned for each channel",
     )
-    software: Optional[Software] = Field(None, title="Acquisition software version data")
+    software: Optional[List[Software]] = Field([], title="Acquisition software version data")
     notes: Optional[str] = Field(None, title="Notes")
 
     @field_validator("axes", mode="before")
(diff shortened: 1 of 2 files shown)


src/aind_data_schema/core/acquisition.py
```diff
@@ -78,6 +78,7 @@ class Acquisition(AindCoreModel):
         title="Processing steps",
         description="List of downstream processing steps planned for each channel",
     )
+    software: Optional[Software] = Field(None, title="Acquisition software version data")
```
saskiad, line 81: Needs to be `Optional[List[Software]] = Field([], title=etc)`

saskiad (changes requested):
Same as other PR - needs to be a list

saskiad (approved):

Issue #861, opened 2024-03-28
Add software to data stream/session


saskiad:
We need a way to document the software used in data collection. Eg. OpenEphys, sync, etc. Stimulus control software is already in the StimulusEpoch class and software for specific devices (e.g. cameras) are in those devices. But we need to document recording software.
Probably it belongs in `Stream` but could also be in the `Session` class. It should be a list.

dyf:
I think `Stream` makes more sense than `Session`.

dyf:
Should be a list, can be optional.

mekhlakapoor:
closing this issue as completed

PR #907, opened 2024-04-30
feat: added optional software list to stream

micahwoodard:
feat #861 
- added optional software list to stream


src/aind_data_schema/core/session.py
```diff
@@ -307,6 +307,7 @@ class Stream(AindModel):
     slap_fovs: Optional[SlapFieldOfView] = Field(None, title="Slap2 field of view")
     stack_parameters: Optional[Stack] = Field(None, title="Stack parameters")
     stream_modalities: List[Modality.ONE_OF] = Field(..., title="Modalities")
+    software: Optional[Software] = Field(None, title="Data stream software information")
```
saskiad, line 310: I think the syntax needs to be `Optional[List[Software]] = Field([], title="...")

saskiad (changes requested):
need to make it a list

saskiad (approved):

PR #922, opened 2024-05-07
Feat 893 laser assembly updates

### LINKED ISSUE #893: Laser Assembly additions
- [ ] Add a field for `collimator` that uses the `Devices` class
- [ ] Add a field for `fiber` that uses the `Patch` class
- [ ] both should be required. This makes this a minor change (rather than a patch)

Sun-flow:
closes #893 

Adding 2 fields to `LaserAssembly`:

`LaserAssembly.collimator` = Device
`LaserAssembly.fiber` = Patch


saskiad (changes requested):
you need to manually update the schema versions because this is a minor (as opposed to a patch, which is automatic) changes to rig

saskiad (approved):

PR #948, opened 2024-05-28
Remove software

### LINKED ISSUE #892: Remove software from devices
We currently have `Software` as an optional field for some devices (e.g. detectors). As we're adding software to session and acquisition, we should probably remove it from the device.

micahwoodard:
fixes #892 
- removed recording_software from Detector class 
- removed encoder_firmware from Disc class
-  removed software  and script from StimulusEpoch


src/aind_data_schema/core/session.py
```diff
@@ -500,16 +500,6 @@ class StimulusEpoch(AindModel):
     )
     stimulus_name: str = Field(..., title="Stimulus name")
     session_number: Optional[int] = Field(default=None, title="Session number")
-    software: Optional[List[Software]] = Field(
```
saskiad, line 503: don't remove the software/script from the session. Thanks!

saskiad (changes requested):
let's not remove software, etc, from the session class.

src/aind_data_schema/core/session.py
micahwoodard: Okay added back in!

jtyoung84:
Decided not to remove this software field

PR #954, opened 2024-05-30
switch max_frame_rate to frame_rate

### LINKED ISSUE #839: Camera frame rate
We currently have the `max_frame_rate` as a property of a camera/detector, but I think we really want a representation of the frame rate that the camera is being recorded at. This would live in the `session` and `acquisition` files. I think it means we need to add a `camera_config` class (and detector). We should determine if there are other config properties that we want to know. It's possible some of the things in the `detector` class in the `device` file should actually be config settings rather than device settings.

saskiad:
closes #839


saskiad:
I decided not to make this a config setting in the Session as it adds a layer of complexity. My experience is that frame rate isn't something that is adjusted session to session but is stable for the device. So I'm keeping it in the rig, but specify that we want the actual frame rate and not the maximum setting.

dyf (approved):

PR #975, opened 2024-06-11
change annotation to timeline

### LINKED ISSUE #955: Add procedure name to timeline plot
Currently subject procedures are labeled with "Surgery" when we want "Nanoject Injection" or "Craniotomy"

saskiad:
closes #955


saskiad:
I haven't tested this - other than I used this code when I first developed it. But I haven't test the function in the schema - see my teams thread.

Issue #941, opened 2024-05-21
Synchronize the ephys rig / session examples


saskiad:
Eg. get names matching. This might have been true in past, but after merging of issue #932 we'll need to fix a few places. I'll document more.

saskiad:
I recommend lumping this into #951

mekhlakapoor:
closed by #983

PR #983, opened 2024-06-28
Feat 965 rigsession compatibility

### LINKED ISSUE #695: DataDescription.data_level needs a validator for valid tags
Users are currently still uploading data which includes outdated `data_level` tags, which need to be then updated manually (via a script that is already written) to our current tag standards.

Introduce a validator to `DataDescription` that checks whether the tag input is valid or not, and converts it to a valid analog if not. If there is no valid analog, reject the model.

We could also solve this via running the `data_description_upgrade` over all `DataDescription`s regularly, but this requires a lot of oversight.

### LINKED ISSUE #965: Rig and session compatibility: Rest of fields
# User story
As a user I want my rig and session models to be synced. Once the standard of how we're doing this is complete (#964 #951), check the rest of the fields

# Acceptance criteria
- [x] Given camera_names in session, check camera_assembly.name OR camera_name in rig (create list and compare)
- [x] Given light_sources session, create a list of names and match against light_source.name in rig
- [x] Given ephys_module.assembly_name in session, match against ephys_assembly.name
- [x] Given stick_microsocope.name in session, match against stick_microscope.name in rig
- [x] Given manipulator_modules, Assembly_name needs to match manipulator_assembly.name
- [x] Given detectors, match detector name
- [x] Given fiber connections in session, match fiber_connection.patch_cord_name with patch_cord.name in rig
- [x] Given fiber_connection.fiber_name, match against fiber_assembly.fibers.name in rig
- [x] Given fiber_module.name, match against fiber_assembly.name
- [x] Given stimulus_epoch.stimulus_device_name match again device.name and reward spout.name in rig

# Sprint Ready Checklist 
- [x] 1. Acceptance criteria defined 
- [x] 2. Team understands acceptance criteria 
- [x] 3. Team has defined solution / steps to satisfy acceptance criteria 
- [x] 4. Acceptance criteria is verifiable / testable 
- [x] 5. External / 3rd Party dependencies identified
- [x] 6. Ticket is prioritized and sized

# Notes
Most of these fields are lists, so will need to create and iterate through the lists of names

mekhlakapoor:
closes #965  and #941 

- Completes compatibility check for session and rig (rest of fields listed in #965)
- For all lists fields, its ensuring that whatever is in session exists in corresponding rig field
- Updates ephys_rig and ephys_session in examples and adds a test to ensure that with any change, they are still in sync
- Adds models in resources


src/aind_data_schema/utils/compatibility_check.py
```diff
@@ -40,9 +40,185 @@ def _compare_daq_names(self) -> Optional[ValueError]:
...
 
+    def _compare_camera_names(self) -> Optional[ValueError]:
+        """Compares camera names"""
+        session_cameras = [
+            camera
+            for stream in getattr(self.session, "data_streams", [])
+            for camera in getattr(stream, "camera_names", [])
+        ]
+        camera_devices = getattr(self.rig, "cameras", []) + getattr(self.rig, "stick_microscopes", [])
+        rig_cameras = [camera_device.camera.name for camera_device in camera_devices]
```
saskiad, line 51: do we need to allow the name to match either the camera assembly name (as here) OR the camera name within the camera assembly. That is, we should probably allow either name to be used in the session.

src/aind_data_schema/utils/compatibility_check.py
```diff
@@ -40,9 +40,185 @@ def _compare_daq_names(self) -> Optional[ValueError]:
...
             )
 
+    def _compare_camera_names(self) -> Optional[ValueError]:
+        """Compares camera names"""
+        session_cameras = [
+            camera
+            for stream in getattr(self.session, "data_streams", [])
+            for camera in getattr(stream, "camera_names", [])
+        ]
+        camera_devices = getattr(self.rig, "cameras", []) + getattr(self.rig, "stick_microscopes", [])
```
saskiad, line 50: since the stick_microscopes have to match the stick_microscopes in the rig, I don't think we want them to match to cameras listed here. Let's remove the stick microscopes from here. We maybe can add a description to the camera field in the session to make it clear these are non-stick microscope cameras.

src/aind_data_schema/utils/compatibility_check.py
```diff
@@ -40,9 +40,188 @@ def _compare_daq_names(self) -> Optional[ValueError]:
...
+            for fiber_connection in getattr(stream, "fiber_connections", [])
+        ]
+        rig_patch_cords = [patch_cord.name for patch_cord in getattr(self.rig, "patch_cords", [])]
+        if not set(session_patch_cords).issubset(set(rig_patch_cords)):
+            return ValueError(
+                f"patch cord names in session do not match patch cord names in rig. "
+                f"session_patch_cords: {set(session_patch_cords)} rig_patch_cords: {set(rig_patch_cords)}"
+            )
+
+    def _compare_fiber_names(self) -> Optional[ValueError]:
```
saskiad, line 152: would it make more sense to compare both fiber_assemblies and patch_cords to the fiber_connections at the same time? If a rig has both patch cords and fiber assemblies (which seems unlikely to me... but probably not impossible) so then a given session ... if it only has the patch cord, then in `compare_patch_cords` it will be fine, but in `_compare_fiber_names` it will throw an error. So I think we need it to make sure that `fiber connections` are a subset of the combination of `patch_cords` and `fiber_assemblies`. Right?

saskiad, line 152: Okay, I've been working through this a bit. I think we don't need this comparison. Patch_cords map onto fiber_connections. And Fiber_assemblies map to fiber_modules. But I don't think we can have fiber_assemblies to fiber_connections. They are two separate ways of having fibers in the brain. So I think we remove the `_compare_fiber_names` class.

examples/ephys_rig.py
```diff
@@ -102,10 +146,31 @@
...
-microscope = CameraAssembly(
+microscope_1 = CameraAssembly(
+    name="Stick_assembly",
+    camera_target=CameraTarget.BRAIN_SURFACE,  # NEEDS TO BE FILLED OUT
+    camera=probe_camera_1,
+    lens=stick_lens,
+)
+
+microscope_2 = CameraAssembly(
+    name="Stick_assembly",
```
saskiad, line 157: these assemblies should have unique names too. Stick_assembly_1/2/3 ...

src/aind_data_schema/utils/compatibility_check.py
```diff
@@ -40,9 +40,188 @@ def _compare_daq_names(self) -> Optional[ValueError]:
...
+
+    def _compare_stick_microscopes(self) -> Optional[ValueError]:
+        """Compares stick microscopes"""
+        session_stick_microscopes = [
+            stick_microscope.assembly_name
+            for stream in getattr(self.session, "data_streams", [])
+            for stick_microscope in getattr(stream, "stick_microscopes", [])
+        ]
+        rig_stick_microscopes = [
+            stick_microscope.camera.name for stick_microscope in getattr(self.rig, "stick_microscopes", [])
```
saskiad, line 100: like the camera assembly above, I think we want this to compare against both the assembly name and the camera name. I think in this case, I'd expect people to use the assembly name more than the camera name, but I'm not certain.

saskiad (changes requested):
A few things, I think mostly this is good, but I think there's a cross-comparison in the fibers (which is how we drew it so that's on me) and then the stick microscope should accept both assembly and camera names.

saskiad (approved):
I think this looks good! I'm sure we'll run into things if we have something askew.

PR #985, opened 2024-07-02
rig id validator

### LINKED ISSUE #970: Validator for rig_id
We'd like to add a validator for `rig_id`. The rules for the id is:
room number_description_date of modification in YYYYMMDD format.

We'd like a validator that confirms:

- [x] room number is a sequence of numbers? (This is a string... but we expect a 3 digit number)
- [x] the date at the end is YYYYMMDD
- [x] There is a str_str_str structure. The description could be something like EPHYS1-OPTO so we should allow additional gaps, but we need the three parts with underscores.

mekhlakapoor:
closes #970

### DIFF
--- a/src/aind_data_schema/core/rig.py
+++ b/src/aind_data_schema/core/rig.py
@@ -2,7 +2,7 @@
 
 from datetime import date
 from typing import List, Literal, Optional, Set, Union
-
+import re
 from aind_data_schema_models.modalities import Modality
 from pydantic import Field, ValidationInfo, field_serializer, field_validator, model_validator
 from typing_extensions import Annotated
@@ -43,6 +43,7 @@ STIMULUS_DEVICES = Annotated[Union[Monitor, Olfactometer, RewardDelivery, Speake
 RIG_DAQ_DEVICES = Annotated[
     Union[HarpDevice, NeuropixelsBasestation, OpenEphysAcquisitionBoard, DAQDevice], Field(discriminator="device_type")
 ]
+RIG_ID_PATTERN = r"^\d{3}_[a-zA-Z0-9-]+_\d{8}$"
 
 
 class Rig(AindCoreModel):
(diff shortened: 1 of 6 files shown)


src/aind_data_schema/core/rig.py
```diff
@@ -43,6 +44,7 @@
 RIG_DAQ_DEVICES = Annotated[
     Union[HarpDevice, NeuropixelsBasestation, OpenEphysAcquisitionBoard, DAQDevice], Field(discriminator="device_type")
 ]
+RIG_ID_PATTERN = r"^\d{3}_[a-zA-Z0-9-]+_\d{8}$"
```
dyf, line 47: I don't think we should restrict room names to 3 digits (I know this was ambiguous in the original issue). That's very 615-Westlake-specific.

I think it it should be any string of alphanumeric characters. @saskiad ?

dyf (changes requested):

src/aind_data_schema/core/rig.py
saskiad: Yeah, our room names are 3 digits, but I think any string of numbers is fine.

mekhlakapoor: Should it be a string of digits (any length) or should it be alphanumeric

src/aind_data_schema/core/rig.py
```diff
@@ -106,6 +108,16 @@ def validate_cameras_other(self):
 
         return self
 
+    @field_validator("rig_id", mode="after")
```
dyf, line 111: on second though, I think you can do this with `Field(..., pattern='blah')` instead of needing to write a field_validator.  And then it will make it into the jsonschema.

dyf (changes requested):

src/aind_data_schema/core/rig.py
mekhlakapoor: just changed it to use the pattern constraint! It throws a validation error with this error message: 

`String should match pattern '^[a-zA-Z0-9]+_[a-zA-Z0-9-]+_\d{8}$' [type=string_pattern_mismatch, input_value='123', input_type=str]`

dyf (approved):

PR #987, opened 2024-07-03
Feat 986 change external links to list of dicts

### LINKED ISSUE #986: Change `Metadata.external_links` from list of dicts to dict of lists
Currently, `Metadata.external_links` takes in `List[Dict[Platform, str]]`, which forces a user to sift through each dict to fin dthe platform they are looking for. Additionally, it is unclear in what situation a separate dict would be necessary for storing relevant info.

Proposing a change to take in `Dict[Platform, List[str]]`

Sun-flow:
closes #986 

`Metadata.external_links` input changed: `List[Dict[Platform, str]]` --> `Dict[Platform, List[str]]`

This might impact how data is drawn from the metadata service / how data is indexed, so @helen-m-lin and @mekhlakapoor should asses those effects.

### DIFF
--- a/src/aind_data_schema/core/metadata.py
+++ b/src/aind_data_schema/core/metadata.py
@@ -78,7 +78,7 @@ class Metadata(AindCoreModel):
     metadata_status: MetadataStatus = Field(
         default=MetadataStatus.UNKNOWN, title=" Metadata Status", description="The status of the metadata."
     )
-    external_links: List[Dict[ExternalPlatforms, str]] = Field(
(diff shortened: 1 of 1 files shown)


mekhlakapoor:
This shouldn't affect the metadata-service

jtyoung84:
This is technically a breaking change. If users have code expecting this field to be in a certain shape, then they'd need to update their code base. If we implement a stricter version control release cycle, then this is something that would need to be discussed. I'll leave it up to Saskia and David to decide on whether to merge this. If we haven't released v1.0, then maybe we can add this before then.

dyf (approved):

PR #999, opened 2024-07-24
moved QC into current schema

saskiad:


saskiad:
still don't understand these error messages. Most of them are for files this PR doesn't touch. The one about this is uninterpretable.

saskiad:
TODO: remove
MANUAL_ANNOTATION = "Manual annotation"
QUALITY_CONTROL_AND_ASSESSMENT = "Quality control and assessment"
from ProcessName Model in aind-data-schema-models

dyf (changes requested):

src/aind_data_schema/core/quality_metrics.py
```diff
@@ -0,0 +1,44 @@
...
+    UNKNOWN = "Unknown"
+
+
+class QCEvaluation(AindModel):
+    """Description of one evaluation stage"""
+
+    evaluation_modality: Modality.ONE_OF = Field(..., title="Modality")
+    evaluation_stage: str = Field(..., title="Evaluation stage")
+    evaluator_full_name: str = Field(..., title="Evaluator full name")
+    evaluation_date: date = Field(..., title="Evaluation date")
```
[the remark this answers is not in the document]
saskiad: I think date.

src/aind_data_schema/core/quality_metrics.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class QualityMetrics(AindCoreModel):
+    """Description of quality metrics for a data asset"""
+
+    _DESCRIBED_BY_URL = AindCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/quality_metrics.py"
+    describedBy: str = Field(_DESCRIBED_BY_URL, json_schema_extra={"const": _DESCRIBED_BY_URL})
+    schema_version: Literal["0.0.1"] = Field("0.0.1")
+    overall_status: Status = Field(..., title="Overall status")
+    overall_status_date: date = Field(..., title="Date of status")
```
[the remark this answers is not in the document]
saskiad: I still think date.

src/aind_data_schema/core/quality_metrics.py
```diff
@@ -0,0 +1,44 @@
...
+from aind_data_schema_models.modalities import Modality
+from aind_data_schema.base import AindCoreModel, AindGeneric, AindGenericType, AindModel
+
+
+class Status(Enum):
+    """QC Status"""
+
+    FAIL = "Fail"
+    PASS = "Pass"
+    UNKNOWN = "Unknown"
```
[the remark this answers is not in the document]
saskiad: I expect we're going to add a bunch of other options to this.

dyf:
@jtyoung84  I am good with this for now, it will change. But it's a new core schema so want your eyes on it.

dyf (approved):

src/aind_data_schema/core/quality_metrics.py
```diff
@@ -0,0 +1,43 @@
...
+from enum import Enum
+from typing import List, Literal, Optional
+
+from pydantic import Field
+
+from aind_data_schema_models.modalities import Modality
+from aind_data_schema.base import AindCoreModel, AindGeneric, AindGenericType, AindModel
+
+
+class Status(Enum):
```
jtyoung84, line 15: We can make this `class Status(str, Enum)` to be able to use some helpful string enum methods

src/aind_data_schema/core/quality_metrics.py
```diff
@@ -0,0 +1,43 @@
...
+
+class QCEvaluation(AindModel):
+    """Description of one evaluation stage"""
+
+    evaluation_modality: Modality.ONE_OF = Field(..., title="Modality")
+    evaluation_stage: str = Field(..., title="Evaluation stage")
+    evaluator_full_name: str = Field(..., title="Evaluator full name")
+    evaluation_date: date = Field(..., title="Evaluation date")
+    qc_metrics: AindGenericType = Field(AindGeneric(), title="QC metrics")
+    stage_status: Status = Field(..., title="Stage status")
```
jtyoung84, line 30: Since this a required field, do we want an `Unknown` option? Will it always be Pass/Fail?

tests/test_quality_metrics.py
```diff
@@ -0,0 +1,41 @@
...
+class QualityControlTests(unittest.TestCase):
+    """test quality metrics schema"""
+
+    def test_constructors(self):
+        """testing constructors"""
+
+        with self.assertRaises(ValidationError):
+            q = QualityControl()
+
+        now = datetime.datetime.now()
```
jtyoung84, line 21: It's better to avoid using values that change each run in unit tests. Can we use a fixed datetime?

jtyoung84:
Do you want to add this in the core.Metadata class also? It won't show up in DocDB otherwise.

jtyoung84:
Should we also add something in the examples folder?

dyf:
> Do you want to add this in the core.Metadata class also? It won't show up in DocDB otherwise.

Yes

dyf:
> Should we also add something in the examples folder?

yes

src/aind_data_schema/core/quality_metrics.py
saskiad: you'll need to duke that out with @dyf - he removed the `Unknown` option. My hunch is that we're going to end up with a handful of Status options very quickly, but I don't think we want one to be Unknown.

jtyoung84: My worry is, that since it's required, if we ever upgrade old metadata records, we'll just fall back to using `QCEvaluation.model_construct()` everywhere, which is frustrating.

jtyoung84 (approved):

PR #1000, opened 2024-07-26
added arena

### LINKED ISSUE #610: Add Arena as Mouseplatform option
arena needs:

- [x] size3D
- [x] material
- [x] shape (enum of rectangle/ellipse/other)

objects in arena include: ladders, wheels, cubes. These will need to be devices. I wonder if we can have a generic "object" class (better name?) for cubes/ladders/etc

(this is part of Motor Observatory needs)

saskiad:
closes #610

### DIFF
--- a/src/aind_data_schema/components/devices.py
+++ b/src/aind_data_schema/components/devices.py
@@ -772,6 +772,14 @@ class Treadmill(MousePlatform):
     width_unit: SizeUnit = Field(default=SizeUnit.CM, title="Width unit")
 
 
+class Arena(MousePlatform):
+    """Description of a rectangular arena"""
+
+    device_type: Literal["Arena"] = "Arena"
+    size: Size3d = Field(..., title="3D Size")
+    objects_in_arena: List[Device = Field(default=[], title="Objects in arena")]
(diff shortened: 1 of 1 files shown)


src/aind_data_schema/components/devices.py
```diff
@@ -772,6 +772,14 @@ class Treadmill(MousePlatform):
     width_unit: SizeUnit = Field(default=SizeUnit.CM, title="Width unit")
 
 
+class Arena(MousePlatform):
+    """Description of a rectangular arena"""
+
+    device_type: Literal["Arena"] = "Arena"
+    size: Size3d = Field(..., title="3D Size")
+    objects_in_arena: List[Device] = Field(default=[], title="Objects in arena")
```
dyf, line 780: Are there specific classes for affordances/objects like this?

saskiad: no. There are going to be a range of objects, some that we make in house and some that are purchased. We decided that using the Device class would be the place to start and we might be able to add other information in the future.

dyf (approved):

PR #1002, opened 2024-07-27
Add (de)serialization tests for AindGeneric

bruno-f-cruz:
This PR adds serialization and deserialization tests for the AindGeneric class. This is to ensure that round-trip conversions are not lossy and can be recovered from the subclasses as well as the AindGeneric type.


jtyoung84 (changes requested):
- It looks like some of the doc strings were copied from another module
- Is it possible to use `self.assertEqual(<expected>, <actual>)` instead of `self.assertTrue`

jtyoung84 (approved):

PR #1011, opened 2024-08-03
anantomical reference

saskiad:
closes #926

Issue #770, opened 2024-02-14
Consider enforcing the use of the `default` keyword


bruno-f-cruz:
**Is your feature request related to a problem? Please describe.**
Consider the following example:

```python

from pydantic import Field, BaseModel


class Foo(BaseModel):
    foo_property: int = Field(0)


class Bar(BaseModel):
    bar_property: int = Field(default=0)


foo = Foo()
# Intellisense output: (*, foo_property: int) -> None

bar = Bar()
# Intellisense output: (*, bar_property: int = 0) -> None

```

Without the explicit use of the `default` keyword, intellisense fails to provide the default value of the class. (I am using vscode, not sure if other editors suffer from the same problem). 

**Describe the solution you'd like**
Enforce the use of the `default` keyword to make the lives of people using the schemas easier. Having the default value be provided as part of the IntelliSense string is a really helpful hint! We should also investigate if it is possible to run a check on all models and see if the keyword exists. Could even be part of the unit tests.

**Describe alternatives you've considered**
N/A

**Additional context**

dyf:
I'm okay with this. @jtyoung84?

jtyoung84:
> I'm okay with this. @jtyoung84?

I'm cool with this also. I'm not sure how to programmatically enforce it without adding the mypy library or something, which might turn into a larger task.

bruno-f-cruz:
Can we please keep using this standard? It makes really difficult to use statically typed libraries if the default keyword is never used. For instance, the `Session` class seems to be a mix between properties with this standard and others without.

PR #1013, opened 2024-08-03
Fix feat 926 coordinates

bruno-f-cruz:
This PR tries to fix #1011/#1012 errors. The issue was related to using enumerator instances (which should be treated as value-types) as inputs to an Union/type signature. If you want to limit the choices users have in terms of value types, use a Literal type instead.
 
Test errors related to the "Bregma" error it was throwing before seem to disappear but couple of other errors remain related with missing fields that I didn't take a look at.

Additionally, following #945 and #770 it would be great if the `default` keyword was preserved in the current schemas if possible.

Hope it helps!

### DIFF
--- a/src/aind_data_schema/core/session.py
+++ b/src/aind_data_schema/core/session.py
@@ -219,7 +219,7 @@ class ManipulatorModule(DomeModule):
     )
     anatomical_coordinates: Optional[Coordinates3d] = Field(None, title="Anatomical coordinates relative to reference")
     anatomical_reference: Optional[
-        Union[CoordinateReferenceLocation.BREGMA, CoordinateReferenceLocation.LAMBDA]
+        Literal[CoordinateReferenceLocation.BREGMA, CoordinateReferenceLocation.LAMBDA]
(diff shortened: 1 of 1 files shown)


src/aind_data_schema/core/session.py
```diff
@@ -351,6 +352,7 @@ class MRIScan(AindModel):
     notes: Optional[str] = Field(default=None, title="Notes", validate_default=True)
 
     @field_validator("notes", mode="after")
+    @classmethod
```
jtyoung84, line 355: Adding this might be outside the scope of this fix. There's debate as to whether this is necessary since `field_validator` already returns a `classmethod` decorator. For some IDEs it raises a warning when removed, in other IDEs it raises a warning when added.

bruno-f-cruz: I am fine with either, but it should be kept consistent and documented somewhere which standard the library is using (same problem with the default keyword). Do you want me to force push or just make a commit to change this?

jtyoung84:
Duplicate of https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1016

PR #1030, opened 2024-08-20
added EMGInsertion

### LINKED ISSUE #996: Options for muscle targets and myomatrix arrays
I'm starting a new issue here to discuss how to specify myomatrix arrays and the muscles they target. 

Mafalda Vicente and I were thinking it could be something like the following for myomatrix array:

- number of branches
  - 4 branches 8 contacts currently
  - 8 branches 4 contacts eventually
  - even more Contacts?
- for each branch
  - number of contact in branch
  - muscle it's in

For the muscles, it could be an enum of the muscles allowed by our protocol to start and we can expand it as we add more muscles to the protocol.

saskiad:
closes #996

### DIFF
--- a/src/aind_data_schema/components/devices.py
+++ b/src/aind_data_schema/components/devices.py
@@ -253,6 +253,13 @@ class LickSensorType(str, Enum):
     PIEZOELECTIC = "Piezoelectric"
 
 
+class MyomatrixArrayType(str, Enum):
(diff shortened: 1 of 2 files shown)


dbirman:
@saskiad I guess the Union types don't need to have `.ONE_OF` appended to them, they just work directly as type annotations

dyf (approved):

PR #1036, opened 2024-08-26
Feat emg side

saskiad:
adding side option for Myomatrix Contact

### DIFF
--- a/src/aind_data_schema/core/procedures.py
+++ b/src/aind_data_schema/core/procedures.py
@@ -560,6 +560,7 @@ class MyomatrixContact(AindModel):
     """ "Description of a contact on a myomatrix thread"""
 
     body_part: MouseAnatomicalStructure.BODY_PARTS = Field(..., title="Body part of contact insertion")
+    side: Side = Field(..., title="Body side")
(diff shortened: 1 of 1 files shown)


dbirman (approved):

dbirman (commented):
Actually after approving this I realized that I don't know why the `schema_version` value is changing? That seems to be causing the error in the examples

src/aind_data_schema/core/procedures.py
```diff
@@ -650,7 +652,7 @@ class Procedures(AindCoreModel):
     _DESCRIBED_BY_URL = AindCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/procedures.py"
     describedBy: str = Field(_DESCRIBED_BY_URL, json_schema_extra={"const": _DESCRIBED_BY_URL})
 
-    schema_version: Literal["1.0.0"] = Field("1.0.0")
+    schema_version: Literal["0.13.14"] = Field("0.13.14")
```
dbirman, line 655: I think this is supposed to be 1.0.0 and then the example will work

saskiad:
ah, that helps. Thanks for catching that.

PR #1037, opened 2024-08-26
Refactor to add QCMetric class and list of URLs for reference images/figures

dbirman:


examples/quality_control.json
```diff
@@ -12,11 +12,18 @@
...
-         "qc_metrics": {
-            "Video_1_num_frames": 662,
-            "Video_2_num_frames": 662,
-            "Frame_match": true
-         },
+         "qc_metrics": [
+            {
+               "name": "Video_frame_match",
+               "value": {
+                  "Video_1_num_frames": 662,
```
saskiad, line 19: I guess I thought that each metrics would be a separate instance of `qc_metrics`, is that not the idea?

src/aind_data_schema/core/quality_control.py
```diff
@@ -19,22 +19,30 @@ class Status(str, Enum):
     PASS = "Pass"
 
 
+class QCMetric(BaseModel):
+    """Description of a single quality control metric"""
+    name: str = Field(..., title="Metric name")
+    value: Any = Field(..., title="Metric value")
+    description: Optional[str] = Field(None, title="Metric description")
```
saskiad, line 26: add the `default=` for the optional fields.

saskiad (changes requested):
fix the defaults. Also a quick question on how you envision using it

examples/quality_control.json
dbirman: Oh I see what you were imagining, so the evaluation here would have two metrics (video 1 frames / video 2 frames) and then the evaluation is passed/failed based on whether they match? (let me know if that's wrong). I'll change the example.

dbirman: For evaluations that have rules like this, do we need a place to put that? I feel like "Notes" doesn't imply that there should be a description of the evaluation rule.

dbirman:
Added quality_control.Stage per discussion, and modified examples to better reflect our concept. Also added name to QCEvaluation and made descriptions optional, to separate stage/name/desc from each other.

src/aind_data_schema/core/quality_control.py
```diff
@@ -19,25 +19,46 @@ class Status(str, Enum):
...
 class QCEvaluation(AindModel):
-    """Description of one evaluation stage"""
+    """Description of one evaluation stage, with one or more metrics"""
 
     evaluation_modality: Modality.ONE_OF = Field(..., title="Modality")
-    evaluation_stage: str = Field(..., title="Evaluation stage")
-    evaluator_full_name: str = Field(..., title="Evaluator full name")
+    evaluation_stage: Stage = Field(..., title="Evaluation stage")
+    evaluation_name: str = Field(..., title="Evaluation name")
+    evaluation_desc: Optional[str] = Field(default=None, title="Evaluation description")
```
saskiad, line 47: can we make this `evaluation_description` - I'm reluctant to shorten things, even obvious things.

src/aind_data_schema/core/quality_control.py
```diff
@@ -19,25 +19,46 @@ class Status(str, Enum):
...
+    evaluator: str = Field(..., title="Evaluator full name")
     evaluation_date: date = Field(..., title="Evaluation date")
-    qc_metrics: AindGenericType = Field(AindGeneric(), title="QC metrics")
+    qc_metrics: List[QCMetric] = Field(title="QC metrics")
```
saskiad, line 50: I think this needs the `...` in the Field

saskiad (changes requested):
super minor things, but otherwise I think it's good.

saskiad (approved):

PR #1041, opened 2024-08-29
chore: making all use of Field(default=None) consistent for Optional …

dbirman:
…fields


dbirman:
This way I don't have to be confused when I look at older files for examples!

saskiad (approved):
Nice!

saskiad:
Thanks for doing this - I thought it had been done earlier, but obviously it missed some places.

PR #1048, opened 2024-09-06
feat: add pending option to QC status

dbirman:


dbirman:
@jtyoung84 any insight about why the tests might be suddenly failing with this single-line addition? I can't get these failures to reproduce on my local python install.

jtyoung84:
Which version of pydantic are you running locally? We don't have an upper bound, so github actions is using 2.9.0, which is pretty new.

dbirman:
Yep that's it, runs fine on 2.8. Looks like there might be breaking changes in 2.9. I can make a separate PR to restrict to <=2.8 and open an issue for the upgrade?

jtyoung84:
> Yep that's it, runs fine on 2.8. Looks like there might be breaking changes in 2.9. I can make a separate PR to restrict to <=2.8 and open an issue for the upgrade?

Maybe set it to <2.9?

saskiad (approved):

PR #1060, opened 2024-09-12
1043 enforce default formatting

dbirman:
@saskiad requested that I get this working, and as long as I limit it to checking for default= and not required fields that are missing Field(...) it seems to work well.

This PR installs a [dev] package `aind-flake8-extensions` published on pypi as a flake8 extension. The extension generates warnings like these:

```
src/aind_data_schema/components/devices.py:530:5: PF001 Field 'coupling_efficiency' should use 'default=None' for optional fields
src/aind_data_schema/components/stimulus.py:42:5: PF001 Field 'pulse_train_interval' should use 'default=None' for optional fields
src/aind_data_schema/core/instrument.py:40:5: PF001 Field 'instrument_id' should use 'default=None' for optional fields
src/aind_data_schema/core/instrument.py:60:5: PF001 Field 'calibration_date' should use 'default=None' for optional fields
src/aind_data_schema/core/instrument.py:65:5: PF001 Field 'calibration_data' should use 'default=None' for optional fields
src/aind_data_schema/core/procedures.py:243:5: PF001 Field 'procedure_name' should use 'default=None' for optional fields
src/aind_data_schema/core/subject.py:100:5: PF001 Field 'genotype' should use 'default=None' for optional fields
src/aind_data_schema/core/subject.py:114:5: PF001 Field 'rrid' should use 'default=None' for optional fields
src/aind_data_schema/core/subject.py:119:5: PF001 Field 'restrictions' should use 'default=None' for optional fields
```

When it sees code like this (from devices.py:530):

```
    coupling_efficiency: Optional[Decimal] = Field(
        None,
        title="Coupling efficiency (percent)",
        ge=0,
        le=100,
    )
```

It also auto-highlights this issue in your editor as you code, if you have a flake8 extension in your IDE.

The extension doesn't fix the issue for you (I've gone through and fixed all of the above).

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -5,3 +5,7 @@ exclude =
     build
 max-complexity = 10
 max-line-length = 120
+
+[flake8:local-plugins]
+extension =
+    PF = aind_flake8_extensions.plugin:run_ast_checks
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,8 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    'aind_data_schema[linters]'
+    'aind_data_schema[linters]',
+    'aind-flake8-extensions'
 ]
 
 linters = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,8 +24,7 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    'aind_data_schema[linters]',
-    'aind-flake8-extensions'
+    'aind_data_schema[linters]'
 ]
 
 linters = [
@@ -34,7 +33,8 @@ linters = [
     'coverage',
     'flake8',
     'interrogate',
-    'isort'
+    'isort',
+    'aind-flake8-extensions'
 ]
 
 docs = [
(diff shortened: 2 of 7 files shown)


dbirman:
@jtyoung84 any suggestion for how to get this working? The extension works fine but I'd prefer to only require it as a [dev] dependency, but this crashes in the workflows because they install the package as-is. The `.flake8` file doesn't have a built-in way to skip that extension if it isn't installed it just errors out.

I could replace the call to flake8 directly with a python script that conditionally runs flake8 w/ w/out the extension enabled.

jtyoung84 (approved):

bruno-f-cruz (commented):
Great tool/idea! I don't have experience with these custom linters so I am not comfortable evaluating the PR. 
I would just ask if running flake8 via CI/CD will raise an exception if any of these are missing, just to make sure we can use this to enforce the keyword on PRs. Otherwise lgtm thanks for the PR :)

dbirman:
> Great tool/idea! I don't have experience with these custom linters so I am not comfortable evaluating the PR. I would just ask if running flake8 via CI/CD will raise an exception if any of these are missing, just to make sure we can use this to enforce the keyword on PRs. Otherwise lgtm thanks for the PR :)

It should raise an exception yes!

PR #1068, opened 2024-09-19
Blocking pydantic versions 2.9.0/2.9.1 in [dev]

### LINKED ISSUE #1058: Breaking changes in pydantic 2.9 (part 2)
```
======================================================================
FAIL: test_validate_ecephys_metadata (__main__.TestMetadata.test_validate_ecephys_metadata)
Tests that ecephys validator works as expected
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/dan/proj/aind/aind-data-schema/tests/test_metadata.py", line 189, in test_validate_ecephys_metadata
    self.assertIn(
AssertionError: 'Missing some metadata for Ecephys. Requires subject, procedures, session, rig, and processing.' not found in "1 validation error for Metadata\n  Value error, Expected 21 fields but got 14 for type `NanojectInjection` with value `NanojectInjection(recovery_time=None, recovery_tim...jection_volume_unit=<VolumeUnit.NL: 'nanoliter'>)` - serialized value may not be as expected. [type=value_error, input_value={'name': 'ecephys_655019_..._axes=None, notes=None)}, input_type=dict]\n    For further information visit https://errors.pydantic.dev/2.9/v/value_error"

======================================================================
FAIL: test_validate_smartspim_metadata (__main__.TestMetadata.test_validate_smartspim_metadata)
Tests that smartspim validator works as expected
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/dan/proj/aind/aind-data-schema/tests/test_metadata.py", line 148, in test_validate_smartspim_metadata
    self.assertIn(
AssertionError: 'Missing some metadata for SmartSpim. Requires subject, procedures, acquisition, and instrument.' not found in "1 validation error for Metadata\n  Value error, Expected 21 fields but got 14 for type `NanojectInjection` with value `NanojectInjection(recovery_time=None, recovery_tim...jection_volume_unit=<VolumeUnit.NL: 'nanoliter'>)` - serialized value may not be as expected. [type=value_error, input_value={'name': 'ecephys_655019_...oftware=[], notes=None)}, input_type=dict]\n    For further information visit https://errors.pydantic.dev/2.9/v/value_error"

```

These remaining issues are due to the use of model_construct() on classes with required fields (that then don't get generated. Something about pydantic 2.9 changed and these are sneaking past our attempt to ignore validation errors. Separating this from the issues Bruno fixed already

dbirman:
This PR ~~increases the pydantic minimum to 2.9.2~~ sets two versions of pydantic to ignore now that 2.9.2 is released, only on the [dev] dependencies

In versions 2.9.0 and 2.9.1 errors are thrown when model_construct() is used on models with Unions. 

Closes #1058

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dynamic = ["version"]
 dependencies = [
     'aind-data-schema-models>=0.3.2',
     'dictdiffer',
-    'pydantic>=2.7,<2.9',
+    'pydantic>=2.7',
     'inflection',
     'jsonschema',
     'semver'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dynamic = ["version"]
 dependencies = [
     'aind-data-schema-models>=0.3.2',
     'dictdiffer',
-    'pydantic>=2.7',
+    'pydantic>=2.9.2',
     'inflection',
     'jsonschema',
     'semver'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dynamic = ["version"]
 dependencies = [
     'aind-data-schema-models>=0.3.2',
     'dictdiffer',
-    'pydantic>=2.7, !=2.9.0, !=2.9.1',
+    'pydantic>=2.7',
     'inflection',
     'jsonschema',
     'semver'
@@ -24,7 +24,8 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    'aind_data_schema[linters]'
+    'aind_data_schema[linters]',
+    'pydantic>=2.7, !=2.9.0, !=2.9.1'
 ]
 
 linters = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ dynamic = ["version"]
 dependencies = [
     'aind-data-schema-models>=0.3.2',
     'dictdiffer',
-    'pydantic>=2.9.2',
+    'pydantic>=2.7, !=2.9.0, !=2.9.1',
     'inflection',
     'jsonschema',
     'semver'


jtyoung84:
In your comment, if you add `Closes #1058`, then it should automatically close the issue when the PR is merged.

dbirman:
> In your comment, if you add `Closes #1058`, then it should automatically close the issue when the PR is merged.

Huh weird, it auto-links when you generate branches from an issue but I think because I changed the name it un-linked it?

jtyoung84:
> > In your comment, if you add `Closes #1058`, then it should automatically close the issue when the PR is merged.
> 
> Huh weird, it auto-links when you generate branches from an issue but I think because I changed the name it un-linked it?

Oh, maybe it will auto close the issue then. I usually create a branch locally.

bruno-f-cruz:
My two cents: I don't think this pin is warranted. Sure 2.9.1 will not work with ADS 1.0.0 but this is largely (well totally if you consider the users) due to a bug in ADS that should be fixed anyway. I would suggest leaving the pin as is as it simplifies the bootstrap of dependencies for consumers. Moreover, the chances that someone pins pydantic on their side to 2.9.1  is very very small since the 2.9.2 patch was released a few days after 2.9.1 anyway.

dbirman:
I think != actually does work in the pyproject.toml file? Changed it to specify the versions to ignore

bruno-f-cruz:
That's even better! 

However I still think that this only makes sense if people want to run tests. Once the next version of ads gets released the other problems should be fixed anyway and the pin stops making sense since it can't be applied retroactively.

I guess we can compromise and apply the specific version rejection until the unit tests get fixed.

jtyoung84:
Is this an issue then with just the unit tests or is it something we'll have to modify the source code for? If it's just the unit tests, we should just update them if it's a relatively quick fix.

dbirman:
It would be a lot of work to fix the tests I think, when I tried the only way I could do it was by generating full definitions for the objects instead of using model_construct(). Arguably this is better since as we've learned the behavior of model_construct is unreliable, but it would take a while to update the tests.

And yes it's true that the version restrictions here are specific to running tests, so we could drop this PR now and tell everybody to update pydantic to 2.9.2.

jtyoung84:
@dyf Do have an opinion? A lot of this is also related to the core.metadata module, which I'm hoping will be updated soon. Maybe we can resolve the unit tests as part of that project?

dbirman:
Okay one more time on this -- I moved the two pydantic versions that cause issues with running tests into the [dev] requirements. Should be fine to merge this now and then open a separate ticket for removing `model_construct()` in as many places as possible.

dyf (approved):

PR #1074, opened 2024-09-23
1072 qc automate state properties in qualitycontrol and evaluation based on metric states

dbirman:
This PR refactors the `overall_status` and `evaluation_status` fields to built from the status of the `QCMetric` objects in each evaluation. The complication to this PR is that it would require people to build QualityControl objects to use these features, if you dump to JSON and you don't re-validate as a model you can only access the private status lists indirectly. You can still use the schema without building the classes you just don't get access to the convenience functions that are defined using the `@property` tags.

Obviously the hope is that nobody has to deal with this stuff since the QC portal handles it for them.

I'll walk through the tests to explain how this works:

```
        test_eval = QCEvaluation(
            evaluation_name="Drift map",
            evaluation_modality=Modality.ECEPHYS,
            evaluation_stage=Stage.PROCESSING,
            qc_metrics=[
                QCMetric(
                    name="Multiple values example",
                    value={"stuff": "in_a_dict"},
                    metric_status_history=[
                        QCStatus(evaluator="Bob", timestamp=datetime.fromisoformat("2020-10-10"), status=Status.PASS)
                    ],
                ),
                QCMetric(
                    name="Drift map pass/fail",
                    value=False,
                    description="Manual evaluation of whether the drift map looks good",
                    reference="s3://some-data-somewhere",
                    metric_status_history=[
                        QCStatus(evaluator="Bob", timestamp=datetime.fromisoformat("2020-10-10"), status=Status.PASS)
                    ],
                ),
            ],
        )


        q = QualityControl(
            evaluations=[test_eval, test_eval],
        )

``` 

This object has status set for the metrics, but not for the evaluation or the full quality control object. If you now call:

```
        q.evaluate_status()
```

You'll find that `self.assertEqual(q.overall_status.status, Status.PASS)`

If you append a metric to one of the evaluations that is set to `Status.PENDING` or `FAIL` the evaluation status and the overall status will subsequently resolve to that state:

```
        # Add a pending metric to the first evaluation
        q.evaluations[0].qc_metrics.append(
            QCMetric(
                name="Drift map pass/fail",
                value=False,
                description="Manual evaluation of whether the drift map looks good",
                reference="s3://some-data-somewhere",
                metric_status_history=[
                    QCStatus(
                        evaluator="Automated", timestamp=datetime.fromisoformat("2020-10-10"), status=Status.PENDING
                    )
                ],
            )
        )

        q.evaluate_status()
        self.assertEqual(q.overall_status.status, Status.PENDING)
```

Timestamps automatically get set to `datetime.now()` unless the user specifies a timestamp, to allow for generating the qc.json separately from the actual processing that happened.


src/aind_data_schema/core/quality_control.py
```diff
@@ -36,7 +36,7 @@ class QCStatus(BaseModel):
 
     evaluator: str = Field(..., title="Status evaluator full name")
     status: Status = Field(..., title="Status")
-    timestamp: date = Field(..., title="Status date")
+    timestamp: datetime = Field(..., title="Status date", default_factory=datetime.now)
```
saskiad, line 39: let's use the `AwareDatetimeWithDefault` consistent with rest of schema.
(I should have pointed this out earlier)

saskiad: I also don't know the the default to now should be in the schema but rather in the qc app

src/aind_data_schema/core/quality_control.py
```diff
@@ -46,6 +46,18 @@ class QCMetric(BaseModel):
...
+    @property
+    def metric_status(self) -> QCStatus:
+        """Get the latest status object for this metric
+
+        Returns
+        -------
+        QCStatus
+            Most recent status object
+        """
+        return self.metric_status_history[-1]
```
saskiad, line 60: this assumes that the list will maintain a chronological order. Do we know if this behavior is true? I feel like JSON moves things into weird orders. But if the python list is in the right order, then json doesn't matter...

saskiad (changes requested):
let's use the same datetime that we use throughout the schema. 

I'm still a little confused by the structure overall - I might have you explain it for me again (or look with fresh eyes in the morning)

src/aind_data_schema/core/quality_control.py
dbirman: I'm not sure if they do, but I'll add a test that does the round trip (pydantic -> json -> pydantic) and confirms that they stay in the same order.

If we ever have issues I'll add a check here that sorts to find the last timestamp

dbirman:
After discussion:

- [ ] Add to QCEvaluation an option to allow fails
- [ ] Add to app a display that says "3 Evaluations, 1 Passed with Failures"

dbirman:
Closes #1085

dbirman:
This PR now also closes two other issues that involved adding `asset_id` fields to QCEvaluation and adding a validator to QCMetric to make sure that a `QCStatus` object is included when pushed (and not just an empty list in metric_status_history)

src/aind_data_schema/core/quality_control.py
```diff
@@ -56,8 +75,62 @@ class QCEvaluation(AindModel):
     evaluation_name: str = Field(..., title="Evaluation name")
     evaluation_description: Optional[str] = Field(default=None, title="Evaluation description")
     qc_metrics: List[QCMetric] = Field(..., title="QC metrics")
-    evaluation_status: List[QCStatus] = Field(..., title="Evaluation status")
+    asset_id: Optional[str] = Field(default=None, title="DocDB asset ID that this evaluation is associated with")
     notes: Optional[str] = Field(default=None, title="Notes")
+    evaluation_status_history: List[QCStatus] = Field(default=[], title="Evaluation status history")
+    allow_failed_metrics: bool = Field(default=False, title="Allow metrics to fail")
```
saskiad, line 81: can we add a description that explains what this is for a bit - I think it will be confusing to people.

src/aind_data_schema/core/quality_control.py
```diff
@@ -56,8 +75,62 @@ class QCEvaluation(AindModel):
     evaluation_name: str = Field(..., title="Evaluation name")
     evaluation_description: Optional[str] = Field(default=None, title="Evaluation description")
     qc_metrics: List[QCMetric] = Field(..., title="QC metrics")
-    evaluation_status: List[QCStatus] = Field(..., title="Evaluation status")
+    asset_id: Optional[str] = Field(default=None, title="DocDB asset ID that this evaluation is associated with")
```
saskiad, line 78: Let's remove this from this PR. I have reservations about using asset_id in the metadata - largely because it's not something that exists in the schema. I'd prefer to resolve that separately from the rest of this

saskiad (changes requested):
I want to deal with the asset id separately. Otherwise I think it looks good.

saskiad (approved):

PR #1075, opened 2024-09-23
1047 qc docs

dbirman:
This PR does two things:

1. Adds .md file support with myst-parser
2. Adds the QC documentation draft, it doesn't include an example workflow because that's blocked by the automated status PR. I'll add that later


docs/source/quality_control.md
```diff
@@ -0,0 +1,33 @@
+# Quality control
+
+**Q: What is the QualityControl object?**
+
+Quality control is a collection of **evaluations** based on sets of **metrics** about the data. You build a QualityControl object by first building ``QCEvaluation`` objects and attaching them to the ``QualityControl`` object.
+
+The QC object reflects the state of the entire data asset and its status is set based on the worst status of its evaluations.
```
saskiad, line 7: this means the overall status is set by the worst status of all the evaluations. but I don't think we want this. We can decide that some evaluations have different priorities. E.g. the tail video can fail but the data asset as a whole can pass.

docs/source/quality_control.md
```diff
@@ -0,0 +1,33 @@
...
+
+We enforce this minimal set of states to prevent ambiguity and to make it easier to build tools that can interpret the status of a data asset.
+
+**Q: When does the state get set for the QualityControl and QCEvaluation objects?**
+
+Call the ``QualityControl.evaluate_status()`` function to append a new status to the object. The status will be timestamped with the current time and reflect the worst status of the attached evaluations or metrics. It's best practice to evaluate the status anytime you add or update evaluations and/or metrics.
+
+Note that while you can evaluate the status of ``QCEvaluation`` objects manually, this could lead to inconsistencies in the overall status of the ``QualityControl`` object.
+
+## QC Portal
```
saskiad, line 29: This should maybe be in a `Questions for AIND users` section as we do for some of the other sections (e.g. the Procedures)

saskiad (changes requested):
I think the overall status needs a little more clarity - or perhaps I do, but want to resolve this before we publicize it as such.

docs/source/quality_control.md
dbirman: This is what I was suggesting in our meeting last week with the idea of "fail" vs "critical fail", where only a critical fail causes the parent evaluation/qc object to also fail. 

Other people have suggested other words for it like "flag" "warning" etc but it seems like we're converging on needing this concept in the status.

dbirman:
- [ ] Need to add more information about Reference URLs

docs/source/quality_control.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Details
+
+**Q: What is an evaluation?**
+
+Each `QCEvaluation` should be thought of as a single aspect of the data asset that can be evaluated for quality, at a specific stage in data acquisition and processing. For example, the brain moves a small amount during both electrophysiology and optical recordings. This drift can be measured by various automated quantiative metrics, as well as qualititative metrics evaluated by a human observer. The state of an evaluation depends on the state of its metrics according to these rules:
+
+- If all metrics pass the evaluation passes
+- If any metric fails the evaluation fails (except when the `allow_failed_metrics` flag is set)
```
saskiad, line 18: I think it would be valuable to unpack how we expect to use `allow_failed_metrics` - I anticipate people making wild assumptions about this and using it in different ways than we intend.

docs/source/quality_control.md
```diff
@@ -0,0 +1,84 @@
...
+- If any metric is pending (and the rest pass) the evaluation is pending
+
+Make sure to annotate `QCEvaluation`s with the `Stage` and `Modality` they are related to.
```
saskiad, line 21: This sounds like a suggestion while these are required fields, so phrase this along the lines of: 
Each `QCEvaluation` applies to a specific data `Modality` and `Stage` (e.g. raw data acquisition or different data processing steps) as specified.

docs/source/quality_control.md
```diff
@@ -0,0 +1,84 @@
...
+
+**Q: What is a metric?**
+
+A metric is any single value or set of values that can be computed, or observed, about a set of data as part of an evaluation. These can have any type, but the QC apps that scientific computing supports expect you to use metrics that are numerical, string, bool, arrays of these values. See the AIND section for special rules for annotating metrics with options.
+
+Metrics have a status which is set by a rule, usually some function of the metric value. The rule should be encoded in the metric description. Metrics can be evaluated multiple times, in which case the new status will be appended the `QCMetric.metric_status_history`. The latest status can be retrived by using `QCMetric.metric_status`.
```
saskiad, line 27: `description` - make it clear you mean in this field rather than in the name of the metric.

saskiad: maybe use `QCMetric` rather than metric - make it clear you are refering to the field/class rather than a broad concept.

docs/source/quality_control.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Build your QualityControl object
+qc = QualityControl(evaluations=evaluations)
+
+qc.write_standard_file()
+```
+
+**Alternate workflow**
+
+If need to build evaluations for a _different_ data asset, for example making evaluations for a primary data asset that you are derived from, but need to be attached to that original asset, then you need to set the `asset_id` field of each QCEvaluation. There is an `aind-data-access-api` function that can help you pull all the relevant `QCEvaluation` objects for a specific `_id`. Note that you can point the `asset_id` field at _any_ asset, so your capsule does not need to be directly linked to a particular primary data asset to generate evaluations for it.
```
saskiad, line 59: I think this needs to be much clearer. What kind of evaluations of primary data come from derived results? I'm VERY VERY worried this is going to be abused, so if we don't establish what this is for, it is going to be chaotic.

I also wonder whether the `asset_id` (and I'm not stoked on using asset_id, and we can discuss separately) should be in the evaluation itself. Can it be somethign in the QCportal that points the evaluation without ending up in the evaluation itself? (A big part of my concern is that there's no asset_id information. So if I see a QC evaluation in a file that says it's for a given asset_id, I have no idea whether that asset_id is the asset I'm looking at or is a derived asset from the one I'm looking at or something else. I'd rather use asset name - that's at least in the schema. 
BUT my gusenseess this is so the QCportal attaches the evaluation to the right asset, and can be part of the process without being in the evaluation itself.

saskiad (changes requested):
a few places need a smidge more clarity.

saskiad (approved):

PR #1084, opened 2024-09-26
1061 refactor core file validators based on aind data schema models expected modalities

dbirman:
This PR takes advantage of the new `ExpectedFiles` model in `aind-data-schema-models` to refactor the core metdata validator to check for required/excluded files for all modalities. Currently the validators only check for ecephys/smartspim platform to confirm they have a few required files. The code now checks by modality (not platform) and confirms that all expected/required files are included.

In the event that two modalities are in conflict the code prioritizes REQUIRED > OPTIONAL > EXCLUDED

### DIFF
--- a/src/aind_data_schema/core/metadata.py
+++ b/src/aind_data_schema/core/metadata.py
@@ -6,8 +6,8 @@ from enum import Enum
 from typing import Dict, List, Literal, Optional, get_args
 from uuid import UUID, uuid4
 
-from aind_data_schema_models.platforms import Platform
 from aind_data_schema_models.modalities import ExpectedFiles, FileRequirement
+from aind_data_schema_models.platforms import Platform
 from pydantic import Field, PrivateAttr, ValidationError, ValidationInfo, field_validator, model_validator
 
 from aind_data_schema.base import AindCoreModel
@@ -21,7 +21,6 @@ from aind_data_schema.core.rig import Rig
 from aind_data_schema.core.session import Session
 from aind_data_schema.core.subject import Subject
 from aind_data_schema.utils.compatibility_check import RigSessionCompatibility
-
 
 CORE_FILES = [
     "subject",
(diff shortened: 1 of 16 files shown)


dbirman:
There's some QC file changes that snuck into this branch... I'll remove those

dbirman:
There's a bug in this PR -- it can't handle modalities with a '-' in the name (need to .replace('-','_'))

jtyoung84:
> There's a bug in this PR -- it can't handle modalities with a '-' in the name (need to .replace('-','_'))

I wonder if it'd be helpful to add the modality folder name explicitly in https://github.com/AllenNeuralDynamics/aind-data-schema-models/blob/main/src/aind_data_schema_models/models/modalities.csv

dbirman:
> > There's a bug in this PR -- it can't handle modalities with a '-' in the name (need to .replace('-','_'))
> 
> ~I wonder if it'd be helpful to add the modality folder name explicitly in https://github.com/AllenNeuralDynamics/aind-data-schema-models/blob/main/src/aind_data_schema_models/models/modalities.csv~
> 
> I guess I'm just using the abbreviation

Yeah I think we only ever use the abbreviation everywhere

src/aind_data_schema/core/metadata.py
```diff
@@ -189,17 +195,43 @@ def validate_metadata(self):
...
+                        file_requirement == FileRequirement.OPTIONAL
+                        and requirement_dict[file] == FileRequirement.EXCLUDED
+                    ):
+                        # override, required wins over all else, and optional wins over excluded
+                        requirement_dict[file] = file_requirement
+
+            for file in CORE_FILES:
+                # Check required case
+                if requirement_dict[file] == FileRequirement.REQUIRED and not getattr(self, file):
+                    raise ValueError(f"{modality.abbreviation} metadata missing required file: {file}")
```
jtyoung84, line 224: This `modality.abbreviation` raises a warning about `referencing before assignment`. Does the modality.abbreviation need to be tracked in the `requirement_dict` in the previous loop?

dbirman: fixed in https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1084/commits/529bf44000cb35c0e1d673e5b4f3b559e2ca60c7

jtyoung84 (approved):

PR #1106, opened 2024-10-16
Support multi session quality control by tracking metric provenance

dbirman:
This PR adds a new Stage "MULTI_SESSION" and `evaluated_assets` field that together differentiate QC that is about this data asset from QC that was calculated in a way that depended on some external data asset. The names of those assets are expected to be included in the evaluated_assets list.

### DIFF
--- a/src/aind_data_schema/core/quality_control.py
+++ b/src/aind_data_schema/core/quality_control.py
@@ -47,7 +47,7 @@ class QCMetric(BaseModel):
     provenance: Optional[List[str]] = Field(
         default=None,
         title="List of asset names that this metric depends on",
-        description="Set to None except when a metric is calculated relative data coming from a different data asset."
+        description="Set to None except when a metric is calculated relative data coming from a different data asset.",
     )
 
     @property
--- a/src/aind_data_schema/core/quality_control.py
+++ b/src/aind_data_schema/core/quality_control.py
@@ -133,8 +133,9 @@ class QCEvaluation(AindModel):
 
     @model_validator(mode="after")
     def validate_multi_session(cls, v):
-        stage = v.get('stage')
-        metrics = v.get('metrics')
+        """Ensure that the evaluated_assets field in any attached metrics is set correctly"""
+        stage = v.stage
+        metrics = v.metrics
 
         if stage == Stage.MULTI_SESSION:
             for metric in metrics:
@@ -147,7 +148,8 @@ class QCEvaluation(AindModel):
             for metric in metrics:
                 if metric.evaluated_assets:
                     raise ValueError(
-                        f"Metric '{metric.name}' is in a single-session QCEvaluation and should not have evaluated_assets"
+                        f"Metric '{metric.name}' is in a single-session ",
+                        "QCEvaluation and should not have evaluated_assets",
                     )
         return v
 
--- a/src/aind_data_schema/core/quality_control.py
+++ b/src/aind_data_schema/core/quality_control.py
@@ -132,7 +132,7 @@ class QCEvaluation(AindModel):
             return failing_metrics
 
     @model_validator(mode="after")
-    def validate_multi_session(cls, v):
+    def validate_multi_asset(cls, v):
         """Ensure that the evaluated_assets field in any attached metrics is set correctly"""
         stage = v.stage
         metrics = v.metrics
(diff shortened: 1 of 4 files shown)


jeromelecoq:
I think this makes senes. I would try to distinguish "multi-asset" for "multi-modalies" somehow. I think you are thinking about "multi_sessions"?

dbirman:
> I think this makes senes. I would try to distinguish "multi-asset" for "multi-modalies" somehow. I think you are thinking about "multi_sessions"?

Yeah this would be multi-session. Multi-modal could be supported by changing the QCEvaluation modality to a list. I'll put that in this same PR and we'll see what people think

seanmcculloch:
I like this addition, and think it is a great way to document the relationship of data-assets on a per-metric basis.

To be clear, these QCMetrics will reside only in this multi-asset quality_control.json and will not get added to the respective single-asset quality_control.jsons, correct?

bjhardcastle:
In the title of `provenance` it says 
> "List of asset names that this metric depends on",

I think this would have to be asset IDs, no? 
AFAIK asset names aren't unique.

dbirman:
> In the title of `provenance` it says
> 
> > "List of asset names that this metric depends on",
> 
> I think this would have to be asset IDs, no? AFAIK asset names aren't unique.

The reason to prefer name is that it's human understandable and easier to search for, and also shared across the data asset on S3 and the record in DocDB. I think the intention is that names should be unique by the way (and we're going back and fixing names that are not unique!)

src/aind_data_schema/core/quality_control.py
```diff
@@ -41,9 +42,14 @@ class QCMetric(BaseModel):
 
     name: str = Field(..., title="Metric name")
     value: Any = Field(..., title="Metric value")
+    status_history: List[QCStatus] = Field(default=[], title="Metric status history")
     description: Optional[str] = Field(default=None, title="Metric description")
     reference: Optional[str] = Field(default=None, title="Metric reference image URL or plot type")
-    status_history: List[QCStatus] = Field(default=[], title="Metric status history")
+    provenance: Optional[List[str]] = Field(
+        default=None,
```
saskiad, line 49: Should the default be an empty List?

saskiad: is it worth having a validator so that when the stage is multi-asset it makes sure the list has length>1?

saskiad: and I'm not sure I like `provenance` as the name for this field. I think of provenance as the history of an asset or result. Maybe `evaluated_assets` or `related_assets`?

saskiad (changes requested):
couple comments on `provenance`

src/aind_data_schema/core/quality_control.py
dbirman: Yeah a validator is a good idea, I'll add that. I'll probably keep the default to None so that when the stage is not multi-asset the field is guaranteed to be blank.

src/aind_data_schema/core/quality_control.py
```diff
@@ -26,6 +26,7 @@ class Stage(str, Enum):
     RAW = "Raw data"
     PROCESSING = "Processing"
     ANALYSIS = "Analysis"
+    MULTI_SESSION = "Multi-session"
```
saskiad, line 29: no, this should be multi-asset.

saskiad (changes requested):
stage should be multi-asset

src/aind_data_schema/core/quality_control.py
dbirman: Whoops thought we had discussed this and decided to change it, just a confusion

src/aind_data_schema/core/quality_control.py
```diff
@@ -125,6 +131,30 @@ def failed_metrics(self) -> Optional[List[QCMetric]]:
 
             return failing_metrics
 
+    @model_validator(mode="after")
+    def validate_multi_session(cls, v):
```
saskiad, line 135: validate_multi_asset

src/aind_data_schema/core/quality_control.py
```diff
@@ -125,6 +131,30 @@ def failed_metrics(self) -> Optional[List[QCMetric]]:
...
+                    raise ValueError(
+                        f"Metric '{metric.name}' is in a multi-session QCEvaluation and must have evaluated_assets set."
+                    )
+        else:
+            # make sure all evaluated assets are None
+            for metric in metrics:
+                if metric.evaluated_assets:
+                    raise ValueError(
+                        (
+                            f"Metric '{metric.name}' is in a single-session QCEvaluation"
```
saskiad, line 152: asset

saskiad (changes requested):
few more places

saskiad (approved):

PR #1109, opened 2024-10-16
Feat 1078 modelschema

### LINKED ISSUE #1078: Schema for models
Define schema for trained models based on features outlined by group discussion and also looking at huggingface and onnx schemas.

saskiad:
closes #1078


src/aind_data_schema/core/model.py
```diff
@@ -0,0 +1,95 @@
...
+    """Description of an analysis model"""
+
+    _DESCRIBED_BY_URL = AindCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/model.py"
+    describedBy: str = Field(_DESCRIBED_BY_URL, json_schema_extra={"const": _DESCRIBED_BY_URL})
+    schema_version: Literal["0.0.1"] = Field("0.0.1")
+
+    name: str = Field(..., title="Name")
+    developer_full_name: Optional[str] = Field(default=None, title="Name of developer")
+    developer_institution: Optional[Organization.ONE_OF] = Field(default=None, title="Institute where developed")
+    modality: Modality.ONE_OF = Field(..., title="Modality")
```
dyf, line 87: should be a list

src/aind_data_schema/core/model.py
```diff
@@ -0,0 +1,95 @@
...
+    model_architecture: ModelArchitecture = Field(..., title="Model architecture")
+    software: Software = Field(..., title="software")
+    direct_use: str = Field(..., title="Intended model use", description="Semantic description of intended use")
```
dyf, line 90: let's call this `intended_use`

src/aind_data_schema/core/model.py
```diff
@@ -0,0 +1,95 @@
...
+    limitations: Optional[str] = Field(default=None, title="Model limitations")
+    training: List[ModelTraining] = Field(..., title="Training")
+    evaluations: Optional[List[ModelEvaluation]] = Field(default=[], title="Evaluations")
+    cumulative_performance: Optional[List[CumulativeEvaluation]] = Field(default=[], title="Cumulative performance")
```
dyf, line 94: can we keep this simple and just have a list of `metrics`?

src/aind_data_schema/core/model.py
```diff
@@ -0,0 +1,95 @@
...
+
+    _DESCRIBED_BY_URL = AindCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/model.py"
+    describedBy: str = Field(_DESCRIBED_BY_URL, json_schema_extra={"const": _DESCRIBED_BY_URL})
+    schema_version: Literal["0.0.1"] = Field("0.0.1")
+
+    name: str = Field(..., title="Name")
+    developer_full_name: Optional[str] = Field(default=None, title="Name of developer")
+    developer_institution: Optional[Organization.ONE_OF] = Field(default=None, title="Institute where developed")
+    modality: Modality.ONE_OF = Field(..., title="Modality")
+    model_architecture: ModelArchitecture = Field(..., title="Model architecture")
```
dyf, line 88: I'm not sure there is a general way to describe the architecture of a model. I would make this a generic dictionary (`AindGeneric`).

src/aind_data_schema/core/model.py
```diff
@@ -0,0 +1,95 @@
...
+    f1_score: Union[Decimal, ScoreStatistics] = Field(..., title="F1 score")
+
+
+class ModelTraining(AindModel):
+    """Description of model training"""
+
+    training_data: str = Field(..., title="Path to training data")
+    training_data_description: Optional[str] = Field(default=None, title="Description of training data") 
+    training_date: AwareDatetimeWithDefault = Field(..., title="Date trained") #not sure we need datetime
+    validation_folds: int = Field(..., title="Validation folds") #is the validation methods always x-fold? or Enum?
```
dyf, line 54: no, I don't think people always do x-fold validation.

src/aind_data_schema/core/model.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class ScoreStatistics(AindModel):
+    """Statistics for x-fold validation scores"""
+
+    mean: Decimal = Field(..., title="Mean")
+    std: Decimal = Field(..., title="Standard deviation")
+
+
+class PerformanceScore(AindModel):
```
dyf, line 40: do these only apply to supervised models?

dyf (commented):
I feel this structure is too specific and nested. 

More generally, how do others describe models? We should talk to Laura.

src/aind_data_schema/core/model.py
dyf: it could even just be a string

src/aind_data_schema/core/model.py
tmchartrand: I wonder if we want a more generic version of modality as well, especially for models like an LLM that is not taking experimental data specifically as input (ie text, image, video etc). Not sure what an alternative name for that would be though.

src/aind_data_schema/core/model.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class PerformanceMetric(AindModel):
+    """Description of a performance metric"""
+
+    name: str = Field(..., title="Metric name")
+    value: Any = Field(..., title="Metric value")
+
+
+class ModelEvaluation(AindModel):
```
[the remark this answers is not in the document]
tmchartrand: actually, now I'm wondering if these should both subclass DataProcess, with ModelEvaluation just adding the performance field? It's a bit off conceptually in the case of unsupervised methods, but I think we're already considering simulation as within the scope of DataProcess so maybe no different from that.

src/aind_data_schema/core/model.py
lauradriscoll: If this is meant to include all models, I agree with Tom that there won't always be training data. In 'task trained' RNNs, the model is trained on a supervised target output. I imagine some models might be 'pre-trained' in this setup and then trained on data at a later stage of training.

src/aind_data_schema/core/model.py
```diff
@@ -0,0 +1,62 @@
...
+from aind_data_schema.base import AindCoreModel, AindGenericType, AindModel, AwareDatetimeWithDefault
+from aind_data_schema.components.devices import Software
+
+
+class ModelArchitecture(AindModel):
+    """Description of model architecture"""
+
+    backbone: ModelBackbone = Field(..., title="Backbone")
+    layers: int = Field(..., title="Layers")
+    parameters: AindGenericType = Field(..., title="Parameters")
```
lauradriscoll, line 22: would it be useful to explicitly include information about whether the model is pytorch, tensorflow etc?

tmchartrand: there's a separate field for software that I think would cover that.

tmchartrand:
closing in favor of #1166

PR #1116, opened 2024-10-21
Refactor schema version

dbirman:
This PR annotates `schema_version` fields with `SkipValidation` allowing old versions of metadata that still validate against the current schema to be considered valid.

Closes #1115


dyf (approved):
I believe this is backward compatible

PR #1122, opened 2024-10-22
Refactor _unit fields to rely on a validator instead of always being set

dbirman:
This PR adds a `AindModel` validator that checks for fields with the suffix `_unit`.

If there is a matching variable `name` for a unit named `name_unit` then the validator ensures that if `name` is set, the unit also needs to be set.

If there are multiple variables matching the pattern `rootname_*` (where * can be anything) for a unit named `rootname_unit` then the validator ensures that if any of the variables are set, the unit is also set.


dbirman:
This turns out not to be a breaking change since the warnings will never appear in old versions since all units are required

CONTRIBUTING.md
```diff
@@ -80,10 +80,29 @@ There are several libraries used to run linters and check documentation. We've i
...
+When you have a single `variable` with a unit, you should add the `_unit` suffix on the name of the unit:
+
+```python
+variable: type = Field(...)
+variable_unit: XUnit = Field(...)
+```
+
+If the variable is `Optional[]` the unit should also be marked as optional.
+
+If you have multiple variables that map onto a single unit type, start each `variable` with the same prefix, i.e. The prefix should be unique within the class and any class that is inherited.
```
saskiad, line 95: it might be helpful to give a specific example here

dbirman: I have this example below that

```python
variable_first_example: type = Field(...)
variable_second_example: type = Field(...)
variable_unit: type = Field(...)
```

Are you thinking we should have a real example instead of this fake one?

saskiad: yeah, I think a real example could be helpful here. I think the example you have is fine for showing the syntax, but a real example unpacks why we sometimes have this. Like `wavelength_cut_on`, `wavelength_cut_off`, `wavelength_unit` or some other real example.

dbirman: Updated it

dbirman:
Todo: some tests need updated fields to pass

dbirman:
We're going to drop python 3.8/3.9, so even though I don't know why it's failing for those I think we can just ignore that?

dbirman:
@jtyoung84 do you know what's going on here? I can't reproduce the python 3.8 test failures on my local machine. We discussed dropping 3.8/3.9 support anyways so maybe this is the chance?

src/aind_data_schema/base.py
```diff
@@ -94,6 +94,37 @@ class AindModel(BaseModel, Generic[AindGenericType]):
...
+                var_name = unit_name.rsplit("_unit", 1)[0]
+
+                # Go through all the values again, if any value matches the variable name
+                # and is set, then the unit needs to be set as well
+                for variable_name, variable_value in values:
+                    if variable_name == var_name:
+                        if variable_value:
+                            raise ValueError(f"Unit {unit_name} is required when {variable_name} is set.")
+                        # if we found our variable and it was None, we can move on
+                        continue
```
jtyoung84, line 117: This line seems to be skipped when I run the tests locally on python 3.8. It seems like you can refactor it to:
```python
                for variable_name, variable_value in values:
                    if variable_name == var_name and variable_value:
                        raise ValueError(f"Unit {unit_name} is required when {variable_name} is set.")
```

src/aind_data_schema/base.py
```diff
@@ -94,6 +94,34 @@ class AindModel(BaseModel, Generic[AindGenericType]):
...
+    def unit_validator(cls, values):
+        """Ensure that all fields matching the pattern variable_unit are set if
+        they have a matching variable that is set (!= None)
+
+        This also checks the multi-variable condition, i.e. variable_unit is set
+        if any of variable_* are set
+        """
+        # Accumulate a dictionary mapping variable : unit/unit_value
+        for unit_name, unit_value in values:
+            if "_unit" in unit_name and not unit_value:
```
jtyoung84, line 107: Pydantic is best when validating against types. It would be ideal if we can add the validator on the "Unit" type instead of whether the field name contains a substring. If that's not feasible, then we can move forward with this implementation.

dbirman: The XUnit types are Enums, not pydantic classes https://github.com/AllenNeuralDynamics/aind-data-schema-models/blob/main/src/aind_data_schema_models/units.py. In any case this idea doesn't work because field/model validators on children can't access fields in the parent (as far as I know), so the validator still ends up on the root model. If you're thinking of a different solution including an example would be useful for showing how it would work.

dbirman: We want the combined Unit+Value classes to be deprecated and removed in the future https://github.com/AllenNeuralDynamics/aind-data-schema-models/issues/86 so I'm not sure this is possible. The removal is the reason I said we're limited to the enum classes. We'll need to discuss w/ @saskiad

jtyoung84: I'm cool with moving forward with your implementation for now and discussing this for aind-data-schema 2.0.

dbirman: Great it's in the doc

jtyoung84 (approved):
We should also wait for Saskia to approve the changes before merging.

saskiad (approved):

PR #1124, opened 2024-10-22
Support modality and stage filters

dbirman:
This PR changes the `status` property to a function `status(modality, stage)` that can optionally filter the `QCEvaluations` in a `QualityControl` object based on their modality and/or stage.

This is a breaking change, but since nobody actively uses this code I don't think we need to push it off to v2.

### DIFF
--- a/src/aind_data_schema/core/quality_control.py
+++ b/src/aind_data_schema/core/quality_control.py
@@ -165,15 +165,29 @@ class QualityControl(AindCoreModel):
     evaluations: List[QCEvaluation] = Field(..., title="Evaluations")
     notes: Optional[str] = Field(default=None, title="Notes")
 
-    @property
-    def status(self) -> Status:
+    def status(self, modality: str = None, stage: Stage = None) -> Status:
         """Loop through all evaluations and return the overall status
 
         Any FAIL -> FAIL
         If no fails, then any PENDING -> PENDING
         All PASS -> PASS
+
+        Parameters
+        ----------
+        modality : str, optional
+            Modality.ONE_OF to filter by, by default None
+        stage : Stage, optional
+            Stage to filter by, by default None
+
+        Returns
+        -------
+        Status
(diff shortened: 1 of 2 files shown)


dbirman:
Removed the review request, will add `tags` as a filter parameter when https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1131 is merged

dbirman (commented):
Add tag filters
Change filters to lists
Date

dbirman:
- [ ] Tag filter
- [ ] Filters -> lists
- [ ] Status on date
- [ ] Coverage

tests/test_base.py
dbirman, line 1: This was caught by the linter, not related to the PR

src/aind_data_schema/core/quality_control.py
```diff
@@ -168,15 +178,36 @@ class QualityControl(AindCoreModel):
...
         Any FAIL -> FAIL
         If no fails, then any PENDING -> PENDING
         All PASS -> PASS
         """
-        eval_statuses = [evaluation.status for evaluation in self.evaluations]
+        if not modality and not stage and not tag:
+            eval_statuses = [evaluation.status(date=date) for evaluation in self.evaluations]
+        else:
+            if modality and not isinstance(modality, list):
+                modality = [modality]
```
saskiad, line 198: Am I correct, this is to make it so if the input to status() is not a list, it becomes one here? 
Is it not easier to just require the input to status() be a list?

saskiad (approved):
I think this looks good - I had one question, but I'm fine if you want to keep it as is.

src/aind_data_schema/core/quality_control.py
dbirman: I find it ugly when you have just one of something to have to wrap it in a list as a user, easier to just hide it in the function

PR #1134, opened 2024-10-30
feat: validate fieldnames in AindGeneric

### LINKED ISSUE #1133: `is_dict_corrupt` should check nested lists recursively
**Describe the bug**
`is_dict_corrupt` is recursively checking if dict keys contain "$" or "." since those are not allowed in document db fieldnames.
However, the check needs to also parse any nested lists.

**To Reproduce**
1. Create a dictionary that contains a nested list containing a dict with invalid key. Example:
```
example = {
    "a": 1,
    "b": 2,
    "c": {
        "d": 3,
        "e": 4,
        "f": [{
            "g": 5,
            "h.corrupt": 6
        }]
    }
}
```
2. Observe that `is_dict_corrupt` returns False

**Expected behavior**
`is_dict_corrupt` should check nested lists recursively

**Additional context**
This issue is causing errors in the lambda function and indexer for several files in the private bucket:
prefixes = [
    "ecephys_722679_2024-06-29_16-53-42",
    "ecephys_722679_2024-06-30_12-17-35",
    "ecephys_722679_2024-07-01_13-10-54",
    "ecephys_722679_2024-07-01_14-29-35",
    "ecephys_722679_2024-07-02_13-40-57"
]

helen-m-lin:
closes #1133 

- Check fields in `AindGeneric` (including in nested lists or dictionaries) for forbidden characters.

Note: Previously this logic was in `aind-data-access-api` and `aind-data-asset-indexer` separately. There was a bug where nested lists were not checked. This is now fixed.


tests/test_base.py
```diff
@@ -55,6 +55,33 @@ def test_aware_datetime_with_setting(self):
...
+        expected_error = (
+            "1 validation error for AindGeneric\n"
+            "  Value error, Field names cannot contain '.' or '$' "
+        )
+        invalid_params = [
+            {"$foo": "bar"},
+            {"foo": {"foo.name": "bar"}},
+        ]
+        for params in invalid_params:
+            with self.assertRaises(ValueError) as e:
```
jtyoung84, line 78: I think these should be raising ValidationError?

src/aind_data_schema/utils/docdb.py
```diff
@@ -0,0 +1,38 @@
+""" Utility methods for metadata jsons stored in AWS DocumentDB. """
```
jtyoung84, line 1: I don't think we need to put this in it's own module. We can put the is_dict_corrupt method inside the Metadata class directly. We don't need to reference DocumentDB directly anywhere.

helen-m-lin: Moved to method inside base module.

tests/test_base.py
helen-m-lin: Fixed

jtyoung84 (approved):

saskiad (approved):

PR #1149, opened 2024-11-08
1057 brain atlas model for targeted structures

dbirman:


src/aind_data_schema/base.py
```diff
@@ -90,10 +91,23 @@ def validate_fieldnames(self):
...
+    Also performs validation checks / coercion / upgrades where necessary
+    """
 
     model_config = ConfigDict(extra="forbid", use_enum_values=True)
 
+    @model_validator(mode="before")
+    def coerce_targeted_structures(cls, values):
+        """If a user passes a targeted_structure as a str, convert to CCFStructure"""
+        for field_name, value in values.items():
+            if "targeted_structure" in field_name and isinstance(value, str):
```
jtyoung84, line 105: It might be unreliable to do validation on field names like this. We can either add field validators as such fields are added or add a custom TargetedStructure Type with a BeforeValidator.

dbirman: I was trying to avoid field validators because we would need them in every class that used to have targeted structures, vs this way we we can do it all in one place. I don't mind doing it the other way, but it will make the data schema more difficult for users to understand.

saskiad: I think I lean towards having a single validator rather than trying to put it in every class (our attempts to do that with the validator for "other" fields failed). And ... correct me if I'm misunderstanding - this validator is to make this backwards compatible. So at the point of a v2.0 we could remove the validator here and not have to rely on it, correct?

saskiad (commented):
I think I lean towards having a single validator rather than trying to put it in every class (our attempts to do that with the validator for "other" fields failed). And ... correct me if I'm misunderstanding - this validator is to make this backwards compatible. So at the point of a v2.0 we could remove the validator here and not have to rely on it, correct?

src/aind_data_schema/base.py
dbirman: Right in 2.0 we would take the validator out. Also we shouldn't ever need to add more of these it's just a one-time thing

saskiad (approved):

PR #1158, opened 2024-11-13
docs: clarifying QC docs for reference images

dbirman:


docs/source/quality_control.md
```diff
@@ -90,15 +94,35 @@ The QC portal automatically calls ``QualityControl.evaluate_status()`` whenever
...
-- Provide a relative path to a file in this data asset's S3 bucket, i.e. "figures/my_figure.png". The mount/asset name should not be included.
-- Provide a url to a FigURL figure
-- Provide the name of one of the interactive plots, e.g. "ecephys-drift-map"
+You can link to your references in one of three ways:
+
+- Provide a relative path to a file in the data asset's S3 bucket, i.e. "figures/my_figure.png". The mount/asset name should not be included.
+- Provide a url to a publicly accessible file, i.e. "https://mywebsite.com/myfile.png"
+- Provide a path to any public S3 bucket, i.e. "s3://bucket/key"
 
 **Q: I saw fancy things like dropdowns in the QC Portal, how do I do that?**
```
saskiad, line 113: I just wonder if this should be in documentation for the QC portal rather than for the metadata schema? I can argue both ways so happy to discuss

saskiad (commented):
see comment

docs/source/quality_control.md
dbirman: Agreed if we had QC portal documentation that was separate I would put it there... but right now there's no qc portal specific documentation and I'm hesitant to add it since I'm finding onboarding people is already difficult even with the documentation all in one place. 

Maybe separating the documentation can be part of a separate ticket?

dbirman:
Just added alternate workflow information for reference images, now that I've tested the FigURL/kachery-cloud workflow

dbirman:
Noticed that a few sections were out-of-date, updated those

saskiad (approved):

PR #1166, opened 2024-11-25
Model schema (updated to use DataProcess)

tmchartrand:
addresses #1078, replacing PR #1109


tmchartrand:
note that one point under discussion here was whether to have separate fields under ModelTraining for hyperparameter search. I had this in an earlier draft, but realized the more straightforward way to represent that might be to have a separate ModelTraining object for hyperparameter optimization and for regular training in cases when both are occurring, with the resulting hyperparameter values stored in the outputs field.

tmchartrand:
tests are dependent on https://github.com/AllenNeuralDynamics/aind-data-schema-models/pull/107

tmchartrand:
Just showed this to Nick and he's on board with the updates @saskiad

src/aind_data_schema/core/model.py
```diff
@@ -0,0 +1,72 @@
...
+
+class ModelTraining(DataProcess):
+    """Description of model training"""
+
+    name: ProcessName = Field(ProcessName.MODEL_TRAINING, title="Process name")
+    train_performance: List[PerformanceMetric] = Field(
+        ..., title="Training performance", description="Performance on training set"
+    )
+    test_performance: Optional[List[PerformanceMetric]] = Field(
+        default=None, title="Test performance", description="Performance on untrained data, evaluated during training"
```
saskiad, line 47: maybe description should be "Performance on test_data..." to contextualize the test_data field

saskiad (approved):
Looks good - I had one tiny edit to the description

PR #1201, opened 2025-01-02
feat: adding a max file size warning

dbirman:
This PR adds a call to `logging.warning` when the file size generated by `write_standard_file` exceeds 500 KB

### DIFF
--- a/src/aind_data_schema/base.py
+++ b/src/aind_data_schema/base.py
@@ -184,5 +184,6 @@ class AindCoreModel(AindModel):
         with open(filename, "w") as f:
             f.write(self.model_dump_json(indent=3))
 
-        if os.path.getsize(filename) > MAX_FILE_SIZE:
+        # Check that size doesn't exceed the maximum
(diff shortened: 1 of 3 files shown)


jtyoung84 (approved):
I'll approve it as it satisfies the wording of the issue, but it'd be more helpful to have a method that estimates the expected file size from the json contents because we don't always write to files. We also wouldn't need to modify the `test_examples.py` file if we checked the size of the json object.

dbirman:
> I'll approve it as it satisfies the wording of the issue, but it'd be more helpful to have a method that estimates the expected file size from the json contents because we don't always write to files. We also wouldn't need to modify the `test_examples.py` file if we checked the size of the json object.

Seems like we should just do that then?

jtyoung84:
> > I'll approve it as it satisfies the wording of the issue, but it'd be more helpful to have a method that estimates the expected file size from the json contents because we don't always write to files. We also wouldn't need to modify the `test_examples.py` file if we checked the size of the json object.
> 
> Seems like we should just do that then?

Sounds good if you want to modify the branch. You can re-request a review from me if you want.

PR #1204, opened 2025-01-02
fix: pull datetime at runtime instead of at package import time

dbirman:
This PR fixes a bug in the `QualityControl.status()` function where the default `date` parameter was being set at package import time instead of at runtime. This causes an issue in downstream code where if a server doesn't get restarted fairly often then it flags QC data with dates in the future (from it's perspective) as being invalid.


dbirman:
Closes https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1204

helen-m-lin (approved):

PR #1220, opened 2025-01-08
objects in arena

### LINKED ISSUE #1127: need a way to specify "object in the arena" in the Session
most likely an optional list of object names.

saskiad:
closes #1127

### DIFF
--- a/src/aind_data_schema/core/session.py
+++ b/src/aind_data_schema/core/session.py
@@ -47,6 +47,7 @@ class StimulusModality(str, Enum):
     """Types of stimulus modalities"""
 
     AUDITORY = "Auditory"
+    FREE_MOVING = "Free moving"
     OLFACTORY = "Olfactory"
     OPTOGENETICS = "Optogenetics"
     NONE = "None"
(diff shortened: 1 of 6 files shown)


src/aind_data_schema/core/session.py
```diff
@@ -537,7 +540,7 @@ class Session(AindCoreModel):
 
     _DESCRIBED_BY_URL = AindCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/session.py"
     describedBy: str = Field(default=_DESCRIBED_BY_URL, json_schema_extra={"const": _DESCRIBED_BY_URL})
-    schema_version: SkipValidation[Literal["1.1.1"]] = Field(default="1.1.1")
+    schema_version: SkipValidation[Literal["1.0.3"]] = Field(default="1.0.3")
```
dbirman, line 543: Something weird happened here and the version got downgraded -- should definitely be 1.1.1

src/aind_data_schema/core/session.py
```diff
@@ -63,6 +64,7 @@ class FiberConnectionConfig(AindModel):
     patch_cord_output_power: Decimal = Field(..., title="Output power (uW)")
     output_power_unit: PowerUnit = Field(default=PowerUnit.UW, title="Output power unit")
     fiber_name: str = Field(..., title="Fiber name (must match procedure)")
+    channels: Optional[List[Channel]] = Field(default=None, title="Channels")
```
dbirman, line 67: Is this part of the objects in arena PR, might have snuck in by accident?

dbirman (changes requested):
There's one line of code that seems like it wasn't meant for this PR, not sure though. The version also got downgraded by accident looks like -- if you revert that and re-generate examples this looks good to go.

src/aind_data_schema/core/session.py
saskiad: yup, thanks!

dbirman (approved):

PR #1223, opened 2025-01-09
feat: replace experimenter_full_name with Investigator class

dbirman:
This PR replaces all usage of `*_full_name: str` with `investigator: List[Investigator]` and introduces a new class `Investigator`. An `Investigator` can be defined either with a first and last name (and optional orcid id) or with an anonymous ID string.


src/aind_data_schema/components/identifiers.py
```diff
@@ -0,0 +1,23 @@
...
+
+from typing import Optional
+from pydantic import BaseModel, Field, model_validator
+from aind_data_schema_models.registries import Registry, _Orcid
+
+
+class Experimenter(BaseModel):
+    """Experimenter identifier"""
+
+    first_name: Optional[str] = Field(default=None, title="Experimenter first name")
```
saskiad, line 11: We want to avoid separate fields for first/last name. The key thing being that there are cultural reasons that make this complicated. This is why we settled on a single field for full name. I'm also against having separate fields that are both optional because we'll end up with half a name which is useless.

src/aind_data_schema/components/identifiers.py
```diff
@@ -0,0 +1,23 @@
...
+    last_name: Optional[str] = Field(default=None, title="Experimenter last name")
+
+    registry: _Orcid = Field(default_factory=lambda: Registry.ORCID, title="Registry")
+    registry_identifier: Optional[str] = Field(default=None, title="ORCID ID")
+    anonymous_id: Optional[str] = Field(default=None, title="Anonymous ID")
```
saskiad, line 16: I think I don't want to have this be a separate field. If I want to separate data based on experimenter, some of which have names and some of which have IDs, I'll have to use multiple fields to access this information. I'd prefer to just have one field that can be used in different ways

saskiad (commented):
I'm just fundamentally torn on this. Is there a reason not to use the PIDName the way we do for Investigators in the data_description? If we do want a class for people, let's use a more generic term rather than experimenter.

src/aind_data_schema/components/identifiers.py
dbirman: The validator would have prevented half names but sure we can store everything in a single field

dbirman:
> I'm just fundamentally torn on this. Is there a reason not to use the PIDName the way we do for Investigators in the data_description? If we do want a class for people, let's use a more generic term rather than experimenter.

Oh I missed investigators, but also didn't we agree months ago to deprecate PIDName in favor of this exact kind of change where the registry is pre-populated for users?

dbirman:
Few more places where *_full_name pattern needs to be replaced

src/aind_data_schema/components/identifiers.py
dbirman, line 1: person

src/aind_data_schema/core/acquisition.py
dbirman, line 1: experimenters (+ session)

src/aind_data_schema/core/model.py
dbirman, line 1: developers

src/aind_data_schema/core/data_description.py
```diff
@@ -41,7 +41,7 @@ class DataDescription(DataCoreModel):
     _DESCRIBED_BY_URL = DataCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/data_description.py"
     describedBy: str = Field(default=_DESCRIBED_BY_URL, json_schema_extra={"const": _DESCRIBED_BY_URL})
     schema_version: SkipValidation[Literal["1.0.4"]] = Field(default="1.0.4")
-    license: Literal["CC-BY-4.0"] = Field("CC-BY-4.0", title="License")
+    license: Literal["CC-BY-4.0"] = Field(default="CC-BY-4.0", title="License")
```
saskiad, line 44: Does it make sense to have a default for a literal? I thought the point of a literal is to have a fixed value. (This is a question, not a review ... )

saskiad (approved):

src/aind_data_schema/core/data_description.py
dbirman: Not related to the literal (and this probably shouldn't be in this PR, I'll remove it). Some linters seem to struggle with interpreting the default value as the default unless you set it explicitly, so when you try to instantiate a DataDescription object the linter complains that the license field is required

saskiad: that's odd. we should see if Jon has thoughts on this. I'm not opposed to having the default, I just was surprised by it.

PR #1227, opened 2025-01-10
refactor: rename AindX classes to show their actual purposes

dbirman:
This PR renames several models in aind-data-schema to reflect their functionality instead of their association to AIND


dyf (approved):
I'm okay with this. I would also be okay with `DataModel` instead of `MetadataModel` to match the name of the repo.

PR #1238, opened 2025-01-14
Breaking: merge instrument and rig, List[Union] pattern, aind-data-schema-models 2.0

dbirman:
This is the first major PR for v2.

This PR is mostly focused on merging the `instrument.py` and `rig.py` files together and sets up `Instrument` to accept a `List[Union[]]` of components instead of specifying parameters to accept. Validators and tests have been overhauled to match the new design pattern.

To allow the `Union` to work properly the `Assembly` classes needed to have a `device_type` field added, even though they don't inherit from `Device` itself. The code throws warnings if you try to add a new option to the union that doesn't include the discriminator field, so there's no need to add a validator. Note that a future v2 PR is going to replace all the type fields with `data_type`, so the specifics here don't matter much -- just note that eventually all *_type fields will become data_type. 

A lot of tests changed because during the refactor I found that a huge number of tests were mis-specified and didn't actually test what they seemed to have been designed for. I've fixed broken tests and brought together all tests related to instrument in one place (rather than being sprinkled in other files like `test_imaging`). Hopefully the tests are more robust now.

For some reason the AindModel->DataModel ended up in this PR, not sure how that happened. 

Unfortunately because this PR is also the point where I had to merge aind-data-schema-models 2.0 it also does some other stuff:
- Deprecates platform
- Upgrades to use external validation for mouse_anatomy
- Changes file requirement validation errors to warnings
- Bumped the minimum python version to 3.10


src/aind_data_schema/components/devices.py
```diff
@@ -264,7 +264,7 @@ class MyomatrixArrayType(str, Enum):
 class Device(DataModel):
     """Generic device"""
 
-    device_type: str = Field(..., title="Device type")  # Needs to be set by child classes that inherits
+    device_type: Literal["device"] = "device"
```
saskiad, line 267: 1) Use sentence casing in the type name: "Device"
2) The behavior here will be that if I'm using the Device class as a generic device, then the type will be "Device" but if I'm using, e.g. a Camera which inherits Device, then the device type will be "Camera"? Or will every device, even from subclasses, have "Device" as their type? That would be bad.

src/aind_data_schema/components/devices.py
```diff
@@ -449,7 +449,8 @@ class CameraAssembly(DataModel):
 
     # required fields
     name: str = Field(..., title="Camera assembly name")
-    camera_target: CameraTarget = Field(..., title="Camera target")
+    device_type: Literal["Camera assembly"] = "Camera assembly"
+    camera_target: Union[CameraTarget, str] = Field(..., title="Camera target")
```
saskiad, line 453: I don't want this to be a string. I want people to use a controlled vocabulary here.

saskiad: if we need to move it to the models repo, we should do that. But I really really want to avoid people having ten different ways of describing the same target, three of which are typos.

src/aind_data_schema/components/devices.py
```diff
@@ -855,7 +859,7 @@ def validate_other(self):
         return self
 
 
-class RewardDelivery(DataModel):
+class RewardDelivery(Device):
```
saskiad, line 862: This shouldn't inherit Device - that will add a bunch of fields that aren't relevant for Reward Delivery. It's like an assembly.

src/aind_data_schema/components/devices.py
```diff
@@ -949,6 +954,3 @@ class MyomatrixArray(Device):
 
     device_type: Literal["Myomatrix Array"] = "Myomatrix Array"
     array_type: MyomatrixArrayType = Field(..., title="Array type")
-
-
-LIGHT_SOURCES = Annotated[Union[Laser, LightEmittingDiode, Lamp], Field(discriminator="device_type")]
```
saskiad, line 954: what's the rationale for getting rid of this? I assume it's just moving somewhere else?

src/aind_data_schema/components/devices.py
dbirman: `device_type` will be replaced by `data_type` https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1245. All subclasses of `DataModel` will have a `data_type` literal set by their class name directly, you can see in that branch how it will work but basically device would convert to "device" and something like EphysAssembly becomes "ephys_assembly". They will all need to be unique.

src/aind_data_schema/components/devices.py
```diff
@@ -449,7 +449,8 @@ class CameraAssembly(DataModel):
 
     # required fields
     name: str = Field(..., title="Camera assembly name")
-    camera_target: CameraTarget = Field(..., title="Camera target")
+    device_type: Literal["Camera assembly"] = "Camera assembly"
```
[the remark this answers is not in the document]
dbirman: Everything that goes into the `List[Union[]]` pattern is required to share a discriminator. In the future this will become `data_type`. For now I needed to patch it in by using device_type.

src/aind_data_schema/components/devices.py
dbirman: It got removed because it gets merged into the `components` List[Union[]] pattern. There's no longer a specific `light_sources` field so there's no need to have a Union for them anymore.

src/aind_data_schema/components/devices.py
dbirman: Typo, fixed

src/aind_data_schema/components/devices.py
dbirman: It would be really nice to get rid of this kind of super specific validation logic, or move it down into the class where the issue lives. I'll revert this change for now but I think we need to have a discussion about changing this pattern

dbirman: This is reverted

src/aind_data_schema/core/instrument.py
```diff
@@ -1,27 +1,66 @@
...
+    Speaker,
 )
 
+# Define the mapping of modalities to their required device types
+# The list of list pattern is used to allow for multiple options within a group, so e.g.
+# FIB requires a light (one of the options) plus a detector and a patch cord
+DEVICES_REQUIRED = {
+    Modality.ECEPHYS.abbreviation: [EphysAssembly],
+    Modality.FIB.abbreviation: [[Laser, LightEmittingDiode, Lamp], [Detector], [Patch]],
+    Modality.POPHYS.abbreviation: [[Laser, LightEmittingDiode, Lamp], [Detector], [Objective]],
```
saskiad, line 55: This should probably only be Laser rather than any light source

src/aind_data_schema/core/instrument.py
```diff
@@ -1,27 +1,66 @@
...
+    Modality.SLAP.abbreviation: [[Laser, LightEmittingDiode, Lamp], [Detector], [Objective]],
```
saskiad, line 56: Ditto

src/aind_data_schema/core/instrument.py
```diff
@@ -1,27 +1,66 @@
...
+    Modality.BEHAVIOR_VIDEOS.abbreviation: [CameraAssembly],
+    Modality.BEHAVIOR.abbreviation: [[Olfactometer, RewardDelivery, Speaker, Monitor]],
```
saskiad, line 58: This is probably fine for now. I might just reduce this to RewardDelivery...

src/aind_data_schema/core/instrument.py
```diff
@@ -30,70 +69,111 @@ class Com(DataModel):
...
 
 
+class Connection(DataModel):
+    """Connection between two devices"""
+
+    device_names: List[str] = Field(..., title="Names of connected devices")
+    inputs: Optional[List[bool]] = Field(default=None, title="Input status")
+    outputs: Optional[List[bool]] = Field(default=None, title="Output status")
+    channels: Optional[List[int]] = Field(default=None, title="Connection channels")
+
```
saskiad, line 80: So this would look like:
`device_names: ["DAQ Board", "Camera"],
inputs: True,
outputs: False,
channels: 3`

what do we mean by Input/Output status?

src/aind_data_schema/core/instrument.py
```diff
@@ -30,70 +69,111 @@ class Com(DataModel):
...
+
 class Instrument(DataCoreModel):
-    """Description of an instrument, which is a collection of devices"""
+    """Description of an instrument"""
 
-    _DESCRIBED_BY_URL = DataCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/instrument.py"
+    # metametadata
+    _DESCRIBED_BY_URL = DataCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/inst.py"
```
saskiad, line 86: Let's keep the file as instrument.py

src/aind_data_schema/core/instrument.py
```diff
@@ -30,70 +69,111 @@ class Com(DataModel):
...
-        default=None,
-        description="Unique instrument identifier, name convention: <room>-<apparatus name>-<date modified YYYYMMDD>",
+    # instrument definition
+    instrument_id: str = Field(
+        ...,
+        description="Unique instrument identifier, name convention: <room>_<apparatus name>_<date modified YYYYMMDD>",
         title="Instrument ID",
+        pattern=instrument_id_PATTERN,
     )
+    mouse_platform: Optional[MOUSE_PLATFORMS] = Field(default=None, title="Mouse platform")
```
saskiad, line 97: why is mouse_platform separate from the other components?

src/aind_data_schema/core/instrument.py
```diff
@@ -30,70 +69,111 @@ class Com(DataModel):
...
+        title="CCF coordinate transform",
+        description="Path to file that details the CCF-to-lab coordinate transform",
     )
+    origin: Optional[Origin] = Field(default=None, title="Origin point for instrument position transforms")
+    instrument_axes: Optional[List[Axis]] = Field(default=None, title="Instrument axes", min_length=3, max_length=3)
+    modalities: List[Modality.ONE_OF] = Field(..., title="Modalities")
     com_ports: List[Com] = Field(default=[], title="COM ports")
-    daqs: List[DAQDevice] = Field(default=[], title="DAQ")
-    notes: Optional[str] = Field(default=None, validate_default=True)
+    instrument_type: Optional[ImagingInstrumentType] = Field(default=None, title="Instrument type")
```
saskiad, line 109: I think this ends up being redundant with modality. Hmm... no, not exactly. But I think we want to remove this
Maybe we want a description field that might be useful for saying "ExaSPIM" or "Ephys rig for dynamic foraging task" where people can endow some meaning. But the `ImagingInstrumentType` is a weird list that conflates platform and modality that I think is outdated.

src/aind_data_schema/core/instrument.py
```diff
@@ -30,70 +69,111 @@ class Com(DataModel):
...
+    manufacturer: Optional[Organization.ONE_OF] = Field(default=None, title="Instrument manufacturer")
```
saskiad, line 110: I'm also thinking we want to remove this. Not sure. 
For SmartSPIM we need it and for confocals we need it. Everything else it doesn't make sense because they are custom. 
Maybe we need a concept of a "microscope" as a device that can be used for SmartSPIM and confocals so we can get that manufacturer information? Even the 2P rigs probably we want to say this is built on a Thorlabs MesoScope...

src/aind_data_schema/core/instrument.py
```diff
@@ -30,70 +69,111 @@ class Com(DataModel):
...
+                PockelsCell,
+                HarpDevice,
+                NeuropixelsBasestation,
+                OpenEphysAcquisitionBoard,
+                OpticalTable,
+                MotorizedStage,
+                ScanningStage,
+                AdditionalImagingDevice,
+                DAQDevice,
+                Device,  # note that order matters in the Union, DAQDevice and Device should go last
```
saskiad, line 151: Does order matter for the others? It's organized roughly conceptually right now, but it might be better to be alphabetical - for searchability purposes? I don't feel super strongly, more thinking out loud.

src/aind_data_schema/core/instrument.py
```diff
@@ -30,70 +69,111 @@ class Com(DataModel):
...
+            ],
+            Field(discriminator="device_type"),
+        ]
+    ] = Field(
+        default=[],
```
saskiad, line 156: we might have this be required since a instrument with no components doesn't make sense. `default=...`

src/aind_data_schema/core/processing.py
```diff
@@ -123,7 +123,7 @@ class Processing(DataCoreModel):
 
     _DESCRIBED_BY_URL: str = DataCoreModel._DESCRIBED_BY_BASE_URL.default + "aind_data_schema/core/processing.py"
     describedBy: str = Field(default=_DESCRIBED_BY_URL, json_schema_extra={"const": _DESCRIBED_BY_URL})
-    schema_version: SkipValidation[Literal["1.1.6"]] = Field(default="1.1.6")
+    schema_version: SkipValidation[Literal["2.0.0"]] = Field(default="2.0.0")
```
saskiad, line 126: Hmm ... I haven't been paying attention. I assume if we touch all of the classes in the schema, all versions will be up to 2.0 ... but I think it's valuable having the classes have their own version number, so when dealing with old files we know which classes are different and which aren't.

saskiad (changes requested):
I haven't gone through quite everything, but the main ones.

src/aind_data_schema/core/instrument.py
dbirman: We'll build this out in the DAQDevice PR instead of here, but the idea we discussed was that if users know which direction signals are going from they can put that into the inputs/outputs lists. I suppose I need to have two ints for the channel list so that you can store that information on both ends.  Originally I wanted to do `input_channel: List[int]` and similar with output_channel.

src/aind_data_schema/core/instrument.py
dbirman: This was typo, resolved

src/aind_data_schema/components/devices.py
saskiad: yeah, that's probably true. I believe there's some validation logic that might look for any of them, but that's not hard to work out.

src/aind_data_schema/components/devices.py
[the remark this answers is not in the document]
saskiad: what's `data_type`?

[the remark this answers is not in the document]
saskiad: ah, I think we discussed this before... it's coming back

src/aind_data_schema/core/instrument.py
dbirman: There's validation logic that specifically applies to mouse_platform in the session/instrument compatibility checker that made it difficult to separate these. I can try to merge it into components.

src/aind_data_schema/core/instrument.py
dbirman: If you want to change this I suggest we do it in a separate ticket (unless you just want to remove it). I think this would be better tracked by creating InstrumentAssembly classes that have some set of default components attached to them for different types of instruments

src/aind_data_schema/core/instrument.py
dbirman: https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1273, resolving this for now

src/aind_data_schema/core/instrument.py
dbirman: Until we have a universal discriminator field the order matters. Once we have `data_type` merged we can make these lists alphabetical.

src/aind_data_schema/core/instrument.py
dbirman: Added required field

src/aind_data_schema/core/processing.py
dbirman: As far as I know all files will eventually be affected by changes in some way

src/aind_data_schema/core/instrument.py
dbirman: Merged, resolved

src/aind_data_schema/core/processing.py
dbirman: I suppose the version bump doesn't need to happen now until we change things but I'd also rather not forget later

src/aind_data_schema/core/instrument.py
saskiad: oh the inputs is a list of booleans. I missed that before. So it would be True/False for which device_name are inputs and which are outputs? Wouldn’t it just be easier to have a inputs be a list of device_names that are inputs (same for outputs)? 
But also, I agree that we can solve this is in the DAQ PR.

src/aind_data_schema/core/instrument.py
saskiad: Let’s make a ticket. I’m not sure just removing it is the answer, but I do want something different.

dbirman: https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1274 feel free to add more detail there, or we'll revisit when I get to it

src/aind_data_schema/core/instrument.py
dbirman: I replaced this with a better temporary placeholder, but yes I'd like to punt it to the DAQ PR to keep things simple here. I haven't actually gone through and done any of the connection mappings or written tests using connection so there's a lot to be done there.

examples/bergamo_ophys_session.json
```diff
@@ -15,8 +15,8 @@
    "session_start_time": "2022-07-12T07:00:00Z",
    "session_end_time": "2022-07-12T07:00:00Z",
    "session_type": "BCI Photometry",
+   "instrument_id": "ophys_inst",
```
saskiad, line 18: can we just try to make this comply with our naming convention:
322_bergamo_20220705 (or whatever combination on room number and dates you want)

examples/bergamo_ophys_session.py
```diff
@@ -28,8 +28,8 @@
     session_end_time=t,
     subject_id="652567",
     session_type="BCI Photometry",
+    instrument_id="ophys_inst",
```
saskiad, line 31: oh, this is the place to fix that.

examples/exaspim_instrument.py
```diff
@@ -5,211 +5,242 @@
...
+        detector_type="Camera",
+        data_interface="Coax",
+        cooling="Air",
+        manufacturer=Organization.VIEWORKS,
+        model="VNP-604MX",
+        serial_number="MB151BAY001",
+    ),
+]
+
+light_sources = [
```
saskiad, line 46: for clarity, let’s make this lasers = […] rather than light_sources since these are just lasers.

examples/fip_behavior_instrument.py
```diff
@@ -0,0 +1,365 @@
...
+)
+from aind_data_schema.core.instrument import Instrument
+
+camera1 = CameraAssembly(
+    name="BehaviorVideography_FaceSide",
+    camera_target=CameraTarget.FACE_SIDE_LEFT,
+    camera=Camera(
+        name="Side face camera",
+        detector_type="Camera",
+        serial_number="TBD",
```
saskiad, line 41: just remove this  - serial number isn’t a required field and I really hate the precedent of fake data that examples suggest are okay.

examples/fip_behavior_instrument.py
```diff
@@ -0,0 +1,365 @@
...
+        sensor_height=480,
+        chroma="Color",
+        cooling="Air",
+        bin_mode="Additive",
+        recording_software=Software(name="Bonsai", version="2.5"),
+    ),
+    lens=Lens(
+        name="Xenocam 1",
+        model="XC0922LENS",
+        serial_number="unknown",
```
saskiad, line 59: ditto

examples/fip_behavior_instrument.py
```diff
@@ -0,0 +1,365 @@
...
+    ),
+)
+
+camera2 = CameraAssembly(
+    name="BehaviorVideography_FaceBottom",
+    camera_target=CameraTarget.FACE_BOTTOM,
+    camera=Camera(
+        name="Bottom face Camera",
+        detector_type="Camera",
+        serial_number="TBD",
```
saskiad, line 72: ditto

examples/fip_behavior_instrument.py
```diff
@@ -0,0 +1,365 @@
...
+        sensor_height=480,
+        chroma="Color",
+        cooling="Air",
+        bin_mode="Additive",
+        recording_software=Software(name="Bonsai", version="2.5"),
+    ),
+    lens=Lens(
+        name="Xenocam 2",
+        model="XC0922LENS",
+        serial_number="unknown",
```
saskiad, line 90: ditto

examples/fip_behavior_instrument.py
```diff
@@ -0,0 +1,365 @@
...
+            lick_sensor=Device(
+                name="Janelia_Lick_Detector Right",
+                manufacturer=Organization.JANELIA,
+            ),
+            lick_sensor_type=LickSensorType("Capacitive"),
+        ),
+    ],
+    stage_type=MotorizedStage(
+        name="NewScaleMotor for LickSpouts",
+        serial_number="xxxx",  # grabbing from GUI/SettingFiles
```
saskiad, line 140: ditto

examples/fip_ophys_instrument.py
```diff
@@ -0,0 +1,332 @@
...
+        sensor_height=480,
+        chroma="Color",
+        cooling="Air",
+        bin_mode="Additive",
+        recording_software=d.Software(name="Bonsai", version="2.5"),
+    ),
+    lens=d.Lens(
+        name="Xenocam 1",
+        model="XC0922LENS",
+        serial_number="unknown",
```
saskiad, line 35: blergh - same here as the last file, but do what you can.

src/aind_data_schema/core/instrument.py
```diff
@@ -1,27 +1,71 @@
...
+# The list of list pattern is used to allow for multiple options within a group, so e.g.
+# FIB requires a light (one of the options) plus a detector and a patch cord
+DEVICES_REQUIRED = {
+    Modality.ECEPHYS.abbreviation: [EphysAssembly],
+    Modality.FIB.abbreviation: [[Laser, LightEmittingDiode, Lamp], [Detector], [Patch]],
+    Modality.POPHYS.abbreviation: [[Laser], [Detector], [Objective]],
+    Modality.SLAP.abbreviation: [[Laser], [Detector], [Objective]],
+    Modality.BEHAVIOR_VIDEOS.abbreviation: [CameraAssembly],
+    Modality.BEHAVIOR.abbreviation: [[RewardDelivery]],
+    Modality.SPIM.abbreviation: [Objective],
```
saskiad, line 64: SPIM probably also needs a Laser and a scanning stage? 
SLAP needs a DMD?
I don't want this to be restrictive, but we should try to make them actually serve as a check

saskiad (approved):
I left a few comments that are all very minor. Feel free to ignore or make small annoying changes. Otherwise I think this looks good.

PR #1240, opened 2025-01-16
feat: add LightAssembly class

dbirman:
This PR adds a LightAssembly class to a allow lights to be combined with a lens. It's possible other devices (filter?) should be allowed in a light assembly as well?


src/aind_data_schema/components/devices.py
dbirman, line 1: Add filter

saskiad (approved):

PR #1243, opened 2025-01-16
refactor: move ground wire information into its own procedures

dbirman:
These PR refactors ground wire information out of the protective material replacement procedure and fixes a mistake in the `ProtectiveMaterialReplacement.procedure_type` field.

I also added a structure target so that this procedure can be used alongside, e.g. the EMG muscle implant procedure instead of embedding that information in that same procedure. This does imply that procedures might need to reference each other, which suggests we need a name field in `DataModel` potentially...


src/aind_data_schema/core/procedures.py
dbirman, line 1: Embed in EMG

dbirman: ground_wire_hole add description of what this is

src/aind_data_schema/core/procedures.py
```diff
@@ -361,20 +361,18 @@ class Headframe(DataModel):
...
-    """Description of a protective material replacement procedure in preparation for ephys recording"""
+class GroundWireImplant(DataModel):
+    """Ground wire implant procedure"""
 
-    procedure_type: Literal["Ground wire"] = "Ground wire"
-    protocol_id: str = Field(..., title="Protocol ID", description="DOI for protocols.io")
-    protective_material: ProtectiveMaterial = Field(
-        ..., title="Protective material", description="New material being applied"
+    procedure_type: Literal["Ground wire implant"] = "Ground wire implant"
+    ground_electrode_location: MouseAnatomicalStructure.ONE_OF = Field(..., title="Location of ground electrode")
```
saskiad, line 368: Would it be valuable to have a sublist of structures that we anticipate using for the ground electrode. We currently use BODY_PARTS for the EMG ground wire position, could we make a list called something ~ GROUND_WIRE_PARTS or GROUND_WIRE_LOCATIONS that has all of those plus brain? The full list of anatomical structures is super long and for people using the GUI it will be impossible to navigate

saskiad (changes requested):
I think we need a subselection of body parts used for ground wires.

src/aind_data_schema/core/procedures.py
dbirman: This is in the models repo as a PR now https://github.com/AllenNeuralDynamics/aind-data-schema-models/pull/147

saskiad (approved):

PR #1250, opened 2025-01-22
Add extensions

dbirman:
This PR adds a field `extensions` in `DataCoreModel` that accepts the generic type.


src/aind_data_schema/base.py
dbirman, line 1: ExtensionModel with a name field?

PR #1260, opened 2025-02-04
fix: fix incorrect use of model_validate_json

dbirman:
This PR fixes a bug where dictionaries were being passed to `model_validate_json` (which accepts a string) instead of `model_validate`.

Because some of the core files have model validators that run in "after" mode and depend on the model being an object and not a dictionary, this bug was causing attempts to validate a (valid) file using `Metadata.model_validate()` to fail. Using the core file directly, i.e. `Rig.model_validate()` always worked.

The bug was causing user warnings which no longer exist, so I had to remove some test assertions, I checked w/ Helen that these aren't important.

I also added two print statements so that these kinds of errors can be caught by users/developers. It was very confusing to debug this because the errors were being hidden.

### DIFF
--- a/src/aind_data_schema/core/metadata.py
+++ b/src/aind_data_schema/core/metadata.py
@@ -147,7 +147,7 @@ class Metadata(AindCoreModel):
         # If the input is a json object, we will try to create the field
         if isinstance(value, dict):
             try:
-                core_model = field_class.model_validate_json(value)
+                core_model = field_class.model_validate(value)
             # If a validation error is raised,
             # we will construct the field without validation.
             except ValidationError:
@@ -203,7 +203,8 @@ class Metadata(AindCoreModel):
                 model_contents = model.model_dump()
                 try:
                     model_class(**model_contents)
-                except ValidationError:
+                except ValidationError as e:
+                    print(f"Error in {field_name}: {e}")
                     metadata_status = MetadataStatus.INVALID
         # For certain required fields, like subject, if they are not present,
         # mark the metadata record as missing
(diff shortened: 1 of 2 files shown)


src/aind_data_schema/core/metadata.py
```diff
@@ -144,13 +144,11 @@ def validate_core_fields(cls, value, info: ValidationInfo):
...
-        # If the input is a json object, we will try to create the field
         if isinstance(value, dict):
             try:
-                core_model = field_class.model_validate_json(value)
-            # If a validation error is raised,
-            # we will construct the field without validation.
-            except ValidationError:
+                core_model = field_class.model_validate(value)
+            except ValidationError as e:
+                print(f"Error in validating {field_name}: {e}")
```
jtyoung84, line 151: `logging.warning` might be better than `print` since it will allow downstream users more control of whether to display the message.

src/aind_data_schema/core/metadata.py
```diff
@@ -144,13 +144,11 @@ def validate_core_fields(cls, value, info: ValidationInfo):
...
                 core_model = field_class.model_construct(**value)
```
jtyoung84, line 152: This might be outside the scope of the ticket, but the transfer service just attaches the dictionary directly instead of using model_construct. There are pros and cons to both, but something we should probably revisit.

dbirman: Lets revisit this as part of a larger conversation about whether we're going to start pushing for valid metadata in the future, especially with the 2.0 rollout coming

jtyoung84: Sounds good

jtyoung84 (approved):

PR #1261, opened 2025-02-04
refactor: replace ValueError in AindGeneric with warning for $/. char…

dbirman:
This PR lowers the error for field names containing '$' and '.' to a warning, to tell users that these fields cause issues in MongoDB queries but still allow their use.

### DIFF
--- a/src/aind_data_schema/base.py
+++ b/src/aind_data_schema/base.py
@@ -87,7 +87,7 @@ class AindGeneric(BaseModel, extra="allow"):
     @model_validator(mode="after")
     def validate_fieldnames(self):
         """Warn users when field names contain forbidden characters
-        
+
         These characters will cause issues with MongoDB queries
         """
         model_dict = json.loads(self.model_dump_json(by_alias=True))
--- a/src/aind_data_schema/base.py
+++ b/src/aind_data_schema/base.py
@@ -3,6 +3,7 @@
 import json
 import re
 import logging
+import warnings
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


src/aind_data_schema/base.py
```diff
@@ -85,10 +86,13 @@ class AindGeneric(BaseModel, extra="allow"):
...
     def validate_fieldnames(self):
-        """Ensure that field names do not contain forbidden characters"""
+        """Warn users when field names contain forbidden characters
+
+        These characters will cause issues with MongoDB queries
+        """
         model_dict = json.loads(self.model_dump_json(by_alias=True))
         if is_dict_corrupt(model_dict):
-            raise ValueError("Field names cannot contain '.' or '$'")
+            warnings.warn("MongoDB queries will not work for fields that contain '.' or '$'")
```
helen-m-lin, line 95: Maybe change this to "may not work as expected" instead of "will not work"? There are a few cases where it actually does work, e.g. $foo when not at top level.

helen-m-lin (changes requested):

helen-m-lin:
@dbirman can you also remove the check for `is_dict_corrupt` here: https://github.com/AllenNeuralDynamics/aind-data-schema/blob/e347569f00eeb54fa2389cbb6afa08d1fa1d3023/src/aind_data_schema/core/metadata.py#L322?

We don't need to ignore those anymore during metadata creation.

helen-m-lin (approved):
Looks great, thanks!

PR #1266, opened 2025-02-10
BREAKING CHANGE: All models inherit data_type from DataModel

dbirman:
This PR adds a field `DataModel.data_type` inherited by all classes in `aind-data-schema`. `data_type` is automatically set to to the value `Literal[snake_case(class.__name__)]`, so e.g. `DataModel.data_type == Literal["data_model"]`, etc. It does not need to be defined in subclasses.

This allows us to use the `data_type` field as a generic discriminator for all discriminated Unions. The fields `procedure_type`, `stimulus_type`, `device_type` have all been replaced by `data_type`. 

Because the type is set at the class level there is also a model validator running in "before" mode which coerces the value to match the class name -- this is needed so that if you try to deserialize a class as a subclass (or vice versa) you don't get validation errors on the data_type field. You can see this happening in `test_aind_generic.py`.


examples/quality_control.json
```diff
@@ -1,8 +1,10 @@
 {
+   "data_type": "Quality control",
    "describedBy": "https://raw.githubusercontent.com/AllenNeuralDynamics/aind-data-schema/main/src/aind_data_schema/core/quality_control.py",
    "schema_version": "2.0.0",
    "evaluations": [
       {
+         "data_type": "Q c evaluation",
```
saskiad, line 7: blergh.... How do we get this to be either "QC evaluation"  or "Quality control evaluation"?

examples/processing.json
```diff
@@ -18,6 +21,7 @@
             "outputs": {},
             "notes": null,
             "resources": {
+               "data_type": "Resource usage",
```
saskiad, line 24: so this is now adding a type field to a class where we didn't have one before. So every time we use a class we're adding data_type fields? Is that correct?

dbirman: Yes exactly, all classes that inherit from DataModel (prev AindModel) will now have a data_type

examples/quality_control.json
dbirman: I can make a special case where if we use multiple capital letters in a row it doesn't put spaces between them

saskiad (approved):

PR #1269, opened 2025-02-12
Release v1.4.0

dbirman:
PR merges change for v1.4 release, see note below though I think we made an error merging some changes into dev.


dbirman:
Now that I'm looking at this -- the device changes are required fields. Doesn't that make this backward incompatible?

dyf (approved):

PR #1272, opened 2025-02-16
BREAKING CHANGE: remove Pipeline/AnalysisProcess and Registration, me…

dbirman:
…rge into Processing.data_processes

### DIFF
--- a/src/aind_data_schema/core/processing.py
+++ b/src/aind_data_schema/core/processing.py
@@ -5,6 +5,7 @@ from typing import List, Literal, Optional, Union
 
 from aind_data_schema_models.process_names import ProcessName
 from aind_data_schema_models.units import MemoryUnit, UnitlessUnit
+from aind_data_schema.core.quality_control import Stage
 from pydantic import Field, SkipValidation, ValidationInfo, field_validator
 
 from aind_data_schema.base import (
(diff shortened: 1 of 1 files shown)


src/aind_data_schema/core/processing.py
```diff
@@ -55,7 +61,17 @@ class DataProcess(DataModel):
...
     name: ProcessName = Field(..., title="Name")
+    stage: ProcessStage = Field(..., title="Processing stage")
+    experimenters: List[Person] = Field(..., title="experimenters", description="People responsible for processing")
     code: Code = Field(..., title="Code used for processing")
+    pipeline_steps: Optional[List[str]] = Field(
+        default=None,
+        title="Pipeline steps",
+        description=(
+            "For pipeline processes, these should be the names of the",
+            " DataProcess objects that are part of the pipeline.",
```
saskiad, line 72: We might need to just add more to indicate that those steps are then listed separately? I think it will not be obvious to people how to specify these.

saskiad (approved):
I think just some info in the description will help people understand expected behavior, but otherwise this is good.

src/aind_data_schema/core/processing.py
dbirman: This is definitely a place where we will want to update the documentation with an example. There's one example that uses the pattern already so we just need to move it onto one of the documentation pages

PR #1281, opened 2025-02-24
Breaking change: merging all uses of code_url/code_version into Software class

dbirman:
This PR merges lingering uses of `code_url` or `code_version` into the `Software` and `Code` classes. It also moves the software class from `devices` and into `components.identifiers`.

The `Software` class is intended to be used for software packages, e.g. Bonsai or SpikeInterface and don't require a specific URL link.

The `Code` class is intended to be used for one-off scripts, they require a URL and can optionally define the software they are running on as well as programming language, etc.

I also fixed a minor error where the `Person` identifier was a `BaseModel` instead of `DataModel`.


src/aind_data_schema/components/identifiers.py
```diff
@@ -1,14 +1,32 @@
...
+
+
+class Software(DataModel):
+    """Software identifier"""
+
+    url: str = Field(..., title="Software URL", description="Path to code repository")
+    name: Optional[str] = Field(default=None, title="Software name", description="Name of the software used")
+    version: Optional[str] = Field(default=None, title="Software version", description="Version of the software used")
+
+    language: Optional[str] = Field(default=None, title="Software language", description="Programming language used")
```
saskiad, line 25: should this maybe be a list?

saskiad: if so, then we need to match language and language version together...

src/aind_data_schema/core/session.py
```diff
@@ -37,8 +37,8 @@
     Scale3dTransform,
     Translation3dTransform,
 )
-from aind_data_schema.components.devices import Calibration, Maintenance, RelativePosition, Scanner, Software, SpoutSide
-from aind_data_schema.components.identifiers import Person
+from aind_data_schema.components.devices import Calibration, Maintenance, RelativePosition, Scanner, SpoutSide
+from aind_data_schema.components.identifiers import Person, Software
 from aind_data_schema.components.stimulus import (
     AuditoryStimulation,
     OlfactoryStimulation,
```
saskiad, line 44: Look at both stream and stimulus epoch to make sure it's being used right there. In stream I think we really only want the language/package + version (e.g. I expect something like ScanImage 3.4 but not with a commit hash. 
For epoch we want to remove either software or script. 
(It won't let me comment in that section of the file, so I'm just tacking this up here)

src/aind_data_schema/components/identifiers.py
saskiad: yeah, this is my big question - we often have software as a list, but I can't tell if we want to continue having software as a list or have the language field inside this be a list.

saskiad (commented):
I'm unsure about the list question (see above)

src/aind_data_schema/components/identifiers.py
dbirman: Sorry I don't think I understood this issue so didn't address it in my modification. Why do we want the software language to be a list?

src/aind_data_schema/core/session.py
dbirman: I think the new version deals with this properly, I used software where we only care vaguely about the package name and code where we care specifically about the code itself.

src/aind_data_schema/core/acquisition.py
```diff
@@ -78,7 +78,7 @@ class Acquisition(DataCoreModel):
         title="Processing steps",
         description="List of downstream processing steps planned for each channel",
     )
-    software: Optional[List[Software]] = Field(default=[], title="Acquisition software version data")
+    software: Optional[List[Software]] = Field(default=[], title="Acquisition software")
```
saskiad, line 81: I think this one wants to be Code

saskiad (approved):
I think this looks good. I want to sanity check the changes to Processing with Alessio and Camillo and Arielle to make sure that this fits with the information they are tracking and we aren't forgetting something they need.

PR #1282, opened 2025-02-26
Docs: minor change to docstrings to improve LLM understanding of created/last_modified fields

dbirman:
See title


saskiad:
is the `created` field here when the *asset* was created or when the metadata record was created? I.e. do we expect it to match the timestamp of the data asset name?

dbirman:
> is the `created` field here when the _asset_ was created or when the metadata record was created? I.e. do we expect it to match the timestamp of the data asset name?

Given the default_factory in `created` I assumed it was the creation time of the record, but this is why we need these description fields!

Issue #1273, opened 2025-02-17
[2.0] Microscope


dbirman:
**Is your feature request related to a problem? Please describe.**
Add a Microscope class that includes manufacturer information  for confocal

dbirman:
Look for use of `OtherImagingType?`

dbirman:
Transform SlapChannel into a SlapMicroscopeConfig

class SlapChannel(Channel):
    """Description of a channel for Slap"""

    dilation: Optional[int] = Field(default=None, title="Dilation (pixels)")
    dilation_unit: Optional[SizeUnit] = Field(default=None, title="Dilation unit")
    description: Optional[str] = Field(default=None, title="Description")

dbirman:
Move magnification from FieldOfView to Microscope

dbirman:
Pre-emptive closure, this is now in the Channel refactor

PR #1284, opened 2025-02-27
BREAKING CHANGE: Combine session and acquisition into acquiession.json

dbirman:
This PR merges the session and acquisition files and tries to simplify the new file. Some of the key changes:

- Standardization of conventions: both StimulusEpoch and DataStream now have a list of `active_devices` and a list of `configurations`. The configurations all inherit from a parent class `DeviceConfig` that requires them to have a `device_name` field -- this is the field that needs to be matched against a corresponding device in the instrument.
- All subject-specific details from `Session` are now stored in a `SubjectDetails` object. This isn't ideal but I think it's better than putting them at the top-level of acquisition. I wanted to have a corresponding `SpecimenDetails`, but there wasn't anything to put in it. A validator ensures that users provide these details.


src/aind_data_schema/core/acquisition.py
```diff
@@ -1,60 +1,212 @@
...
+    )
+    animal_weight_post: Optional[Decimal] = Field(
+        default=None,
+        title="Animal weight (g)",
+        description="Animal weight after procedure",
+    )
+    weight_unit: MassUnit = Field(default=MassUnit.G, title="Weight unit")
+    anaesthesia: Optional[Anaesthetic] = Field(default=None, title="Anaesthesia")
+    mouse_platform_name: str = Field(..., title="Mouse platform")
+    active_mouse_platform: bool = Field(
```
saskiad, line 80: we need to change this name because it is being completely misused.

src/aind_data_schema/components/configs.py
```diff
@@ -63,14 +51,19 @@ class StimulusModality(str, Enum):
...
+
 # Ophys components
-class FiberConnectionConfig(DataModel):
+class FiberConnectionConfig(DeviceConfig):
     """Description for a fiber photometry configuration"""
 
     patch_cord_name: str = Field(..., title="Patch cord name (must match rig)")
     patch_cord_output_power: Decimal = Field(..., title="Output power (uW)")
     output_power_unit: PowerUnit = Field(default=PowerUnit.UW, title="Output power unit")
-    fiber_name: str = Field(..., title="Fiber name (must match procedure)")
```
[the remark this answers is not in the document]
saskiad: I think perhaps the key is that this isn't a FiberConfig but a FiberConnectionConfig - it's not about the fiber, it's about the pairing.

src/aind_data_schema/components/configs.py
```diff
@@ -183,7 +174,7 @@ class SlapSessionType(str, Enum):
 class SlapFieldOfView(FieldOfView):
     """Description of a Slap2 scan"""
 
-    session_type: SlapSessionType = Field(..., title="Session type")
+    acquisition_type: SlapSessionType = Field(..., title="Session type")
```
saskiad, line 177: have the title match the field name.

src/aind_data_schema/components/configs.py
```diff
@@ -192,11 +183,16 @@ class SlapFieldOfView(FieldOfView):
...
+
+    objects_in_area: List[str] = Field(..., title="Objects in area")
+
+
 # Ephys Components
-class DomeModule(DataModel):
+class DomeModule(DeviceConfig):
     """Movable module that is mounted on the ephys dome insertion system"""
 
-    assembly_name: str = Field(..., title="Assembly name")
```
saskiad, line 199: I do think this needs to be specifically for an assembly and not a generic device.

src/aind_data_schema/components/configs.py
```diff
@@ -327,15 +315,17 @@ class SubjectPosition(str, Enum):
     SUPINE = "Supine"
 
 
-class MRIScan(DataModel):
+class MRIScan(DeviceConfig):
```
saskiad, line 318: I'm confused about when you use DeviceConfig and when you use DataModel. This serves the same purpose as the FOV or SlapFOV classes. Why is this a device config?

src/aind_data_schema/components/configs.py
```diff
@@ -327,15 +315,17 @@ class SubjectPosition(str, Enum):
...
     """Description of a 3D scan"""
 
+    headframe_registration: Optional[Affine3dTransform] = Field(
```
saskiad, line 321: this is not part of the mri acquisition - it is part of ephys acquisition

src/aind_data_schema/components/configs.py
```diff
@@ -381,215 +371,67 @@ def validate_primary_scan(self):
...
-            message = "\n     ".join(errors)
-            raise ValueError(message)
-        return value
+    medium: ImmersionMedium = Field(..., title="Immersion medium")
+    refractive_index: Decimal = Field(..., title="Index of refraction")
 
 
-class StimulusEpoch(DataModel):
-    """Description of stimulus used during session"""
+class ProcessingSteps(DataModel):
```
saskiad, line 381: not for this PR, but I'd like to remove this from acquisition. We need to talk with the team about this - they want to communicate to downstream processors what the intention is, but I don't think this is the right way to go about it.

src/aind_data_schema/components/configs.py
```diff
@@ -381,215 +371,67 @@ def validate_primary_scan(self):
...
-    trials_total: Optional[int] = Field(default=None, title="Total trials")
-    trials_finished: Optional[int] = Field(default=None, title="Finished trials")
-    trials_rewarded: Optional[int] = Field(default=None, title="Rewarded trials")
-    notes: Optional[str] = Field(default=None, title="Notes")
+    ] = Field(...)
 
 
-class Session(DataCoreModel):
-    """Description of a physiology and/or behavior session"""
+class ImagingConfig(DataModel):
```
saskiad, line 402: InVitroImagingConfig

saskiad: or something in that vein

src/aind_data_schema/components/configs.py
```diff
@@ -381,215 +371,67 @@ def validate_primary_scan(self):
...
-        default=[],
-        title="experimenter(s)",
-    )
-    session_start_time: AwareDatetimeWithDefault = Field(..., title="Session start time")
-    session_end_time: Optional[AwareDatetimeWithDefault] = Field(default=None, title="Session end time")
-    session_type: str = Field(..., title="Session type")
-    instrument_id: str = Field(..., title="Instrument ID")
-    ethics_review_id: Optional[str] = Field(default=None, title="Ethics review ID")
-    calibrations: List[Calibration] = Field(
+    tiles: List[AcquisitionTile] = Field(..., title="Acquisition tiles")
```
saskiad, line 405: I think this goes in Stream rather than config - we discussed earlier

src/aind_data_schema/core/acquisition.py
```diff
@@ -1,60 +1,212 @@
...
+class DataStream(DataModel):
+    """Data streams with a start and stop time"""
+
+    stream_start_time: AwareDatetimeWithDefault = Field(..., title="Stream start time")
+    stream_end_time: AwareDatetimeWithDefault = Field(..., title="Stream stop time")
+    modalities: List[Modality.ONE_OF] = Field(..., title="Modalities")
+    software: Optional[List[Software]] = Field(default=[], title="Software packages")
+    notes: Optional[str] = Field(default=None, title="Notes")
+
+    active_devices: List[str] = Field(..., title="Active devices")
```
saskiad, line 97: add a description indicating that these are the names of the specific devices

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
dbirman: So it's possible that this can get fixed downstream by replacing this class with the upcoming generic `Connection` class. I would prefer that since it's one more step toward standardization! Do you think that class will cover all the features / maybe we should design it with that intention?

src/aind_data_schema/components/configs.py
dbirman: DeviceConfig causes the class to inherit the extra `device_name` field, just a way of making that easier without typing it out each time.

src/aind_data_schema/core/acquisition.py
dbirman: I moved this into a config object `MousePlatformConfig` which I think makes more sense for how we *want* it to be used!

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
dbirman: My plan is that Connection is something that can be passed into the configurations list in a DataStream, which I think captures this situation

src/aind_data_schema/components/configs.py
dbirman: Fixed this and a lot of other places I accidentally left session

src/aind_data_schema/components/configs.py
dbirman: This is a DeviceConfig so it inherits the `device_name` field now

src/aind_data_schema/components/configs.py
dbirman: But then why does the description say "MRI transform matrix for headframe"?

src/aind_data_schema/core/acquisition.py
```diff
@@ -63,43 +215,55 @@ class Acquisition(DataCoreModel):
...
-    tiles: List[AcquisitionTile] = Field(..., title="Acquisition tiles")
-    axes: List[ImageAxis] = Field(..., title="Acquisition axes")
-    chamber_immersion: Immersion = Field(..., title="Acquisition chamber immersion data")
-    sample_immersion: Optional[Immersion] = Field(default=None, title="Acquisition sample immersion data")
-    active_objectives: Optional[List[str]] = Field(default=None, title="List of objectives used in this acquisition.")
+
+    # Information about the acquisition
+    acquisition_start_time: AwareDatetimeWithDefault = Field(..., title="Acquisition start time")
+    acquisition_end_time: AwareDatetimeWithDefault = Field(..., title="Acquisition end time")
+    acquisition_type: Optional[str] = Field(default=None, title="Acquisition type")
```
saskiad, line 219: we should probably make this a required field. It's pretty important for physiology/behavior data, and wouldn't hurt for in vitro imaging.

src/aind_data_schema/core/acquisition.py
```diff
@@ -63,43 +215,55 @@ class Acquisition(DataCoreModel):
...
+    @model_validator(mode="after")
+    def subject_details_if_not_specimen(self):
+        """Check that subject details are present if no specimen ID"""
+        if not self.specimen_id and not self.subject_details:
+            raise ValueError("Subject details are required for in vivo experiments")
+
+        return self
+
+    @model_validator(mode="after")
+    def check_subject_specimen_id(self):
```
saskiad, line 246: This is specific to AIND, so this might be prohibitive for non-AIND users?

saskiad (changes requested):
mostly stuff we talked about

src/aind_data_schema/components/configs.py
dbirman: The whole InVitroImagingConfig ends up inside the `DataStream.config` list (that's what we discussed). Did you mean for tiles specifically to go somewhere else?

src/aind_data_schema/core/acquisition.py
dbirman: I weakened the requirement to just be that the specimen_id appear inside the subject_id, but that doesn't solve this. I'm not sure we should remove this though... people put junk in the data_description too often for me to want to risk not getting this information.

Maybe eventually we need regex fields for these so that people can fork the repo and re-define the regex patterns.

src/aind_data_schema/components/configs.py
dbirman: Oh okay I think I get it, this is the transform matrix for the headframe, determined from the MRI, for a particular headframe setup?

src/aind_data_schema/components/configs.py
```diff
@@ -1,53 +1,41 @@
-""" Schemas for Physiology and/or Behavior Sessions """
+""" Configurations for devices, software, and other components """
```
saskiad, line 1: maybe add "during acqusition" to this description. 
My thinking is to help people understand the difference between the `device` file and the `configs` file.

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
saskiad: I wonder if maybe the thing to do is make the class something like `PatchCordConfig` and tether it to the patch device. Then the device_name is the patch cord name. 

my worry with Connection is (1) that we'd need to get the power info in there and (2) I think that users wouldn't understand that this was the fiber photometry data stream as opposed to a daq connection...

src/aind_data_schema/components/configs.py
saskiad: right, that's my concern. We want it to point to an assembly.

Perhaps the thing to do is have a validator on the type of devices that can be used here, because I think the key thing is it needs to be an assembly that has a defined manipulator. 

Or, perhaps we can get rid of the EphysAssembly and FiberAssembly classes by adding a manipulator to this DomeModuleConfig? But no, because each of those assemblies takes a list of probes/fibers. So the device_name here would have to point to that list, which is the assembly. 

Okay, so back to validator on the device_type of the device that this config is used for. Does that work?

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
saskiad: The place where this might be messy is the FiberModule which also uses this to modify the FiberAssembly.

src/aind_data_schema/components/configs.py
saskiad: I don't think that's my confusion here. The MRIScan isn't just describing configuration of the scanner. I see it as akin to the FOV class, which isn't a a DeviceConfig. If this is about the configuration of the scanner, it would only be a subset of the fields that are in here.  And we'd need a different class for the other fields.

dbirman: I think it makes sense, a scan is very literally a configuration of a scanner. Maybe it makes more sense in context, I was also thinking of changing FOV/SlapFOV to be device configurations as well once I add Microscope https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1273

Which fields do you think don't make sense to be here if this is a configuration of the scanner?

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
dbirman: I changed this to PatchCordConfig, I think it makes sense now.

src/aind_data_schema/components/configs.py
saskiad: Hmm ... I thought we'd talked about separating the immersion fields from the tiles/axes ... but I can't remember why so I'm not sure we need to.

src/aind_data_schema/components/configs.py
```diff
@@ -192,11 +183,24 @@ class SlapFieldOfView(FieldOfView):
     path_to_array_of_frame_rates: str = Field(..., title="Array of frame rates")
 
 
+class MousePlatformConfig(DeviceConfig):
+    """Configuration for mouse platforms"""
+
+    objects_in_arena: List[str] = Field(..., title="Objects in area")
```
saskiad, line 188: (a) objects_in_arena needs to be optional field

(b) I'm wrestling with whether these need to be two separate configs? ArenaConfig with objects_in_arena and TreadmillConfig with active_control?

src/aind_data_schema/core/acquisition.py
```diff
@@ -1,60 +1,201 @@
...
+                ManipulatorModule,
+                DomeModule,
+                DetectorConfig,
+                FiberConnectionConfig,
+                FiberModule,
+                FieldOfView,
+                SlapFieldOfView,
+                Stack,
+                MRIScan,
+                InVitroImagingConfig,
```
saskiad, line 115: MousePlatformConfig too?

src/aind_data_schema/core/acquisition.py
```diff
@@ -1,60 +1,201 @@
...
+    stimulus_end_time: AwareDatetimeWithDefault = Field(
+        ...,
+        title="Stimulus end time",
+        description="When a specific stimulus ends. This might be the same as the acquisition end time.",
+    )
+    stimulus_name: str = Field(..., title="Stimulus name")
+    code: Optional[Code] = Field(
+        default=None,
+        title="Code or script",
+        description="Custom code or script used to control the behavior/stimulus",
```
saskiad, line 153: "Code used to control the behavior/stimulus along with parameters" ? - I want something to indicate that this is where the stimulus parameters live.

src/aind_data_schema/core/acquisition.py
```diff
@@ -1,60 +1,201 @@
...
+    )
+    stimulus_modalities: List[StimulusModality] = Field(..., title="Stimulus modalities")
+    output_parameters: GenericModelType = Field(default=GenericModel(), title="Performance metrics")
+    reward_consumed_during_epoch: Optional[Decimal] = Field(default=None, title="Reward consumed during training (uL)")
+    reward_consumed_unit: VolumeUnit = Field(default=VolumeUnit.UL, title="Reward consumed unit")
+    trials_total: Optional[int] = Field(default=None, title="Total trials")
+    trials_finished: Optional[int] = Field(default=None, title="Finished trials")
+    trials_rewarded: Optional[int] = Field(default=None, title="Rewarded trials")
```
saskiad, line 161: there's a part of me that thinks it might be good to group output_parameters, reward, trials all into some behavior performance class. Kind of the way we have a SubjectDetails. But I don't think it helps anything, right?

src/aind_data_schema/core/acquisition.py
```diff
@@ -63,43 +204,55 @@ class Acquisition(DataCoreModel):
...
-    tiles: List[AcquisitionTile] = Field(..., title="Acquisition tiles")
-    axes: List[ImageAxis] = Field(..., title="Acquisition axes")
-    chamber_immersion: Immersion = Field(..., title="Acquisition chamber immersion data")
-    sample_immersion: Optional[Immersion] = Field(default=None, title="Acquisition sample immersion data")
-    active_objectives: Optional[List[str]] = Field(default=None, title="List of objectives used in this acquisition.")
+
+    # Information about the acquisition
+    acquisition_start_time: AwareDatetimeWithDefault = Field(..., title="Acquisition start time")
+    acquisition_end_time: AwareDatetimeWithDefault = Field(..., title="Acquisition end time")
+    acquisition_type: str = Field(default=None, title="Acquisition type")
```
saskiad, line 218: ugh. I want this field to be session_type. For whatever reason I see acquisition_type to be akin to modality where we want something that is indicating experimental goals. Is there a different name?

saskiad: experiment_type? or even experiment?

src/aind_data_schema/core/acquisition.py
```diff
@@ -63,43 +204,55 @@ class Acquisition(DataCoreModel):
...
     local_storage_directory: Optional[str] = Field(default=None, title="Local storage directory")
     external_storage_directory: Optional[str] = Field(default=None, title="External storage directory")
-    processing_steps: List[ProcessingSteps] = Field(
```
saskiad, line 76: I think we should remove the two storage_directory fields. I don't think they are being used and I think they will end up confusing things given that this record is associated with a file with a location...

src/aind_data_schema/core/acquisition.py
saskiad: Sounds good. I agree that we want something - we can revisit it if it becomes a problem with external users.

saskiad (changes requested):
More comments - looking good!

src/aind_data_schema/components/configs.py
dbirman: My logic is that all mouse platforms can be under active control, just a generic statement of whether the platform is also a stimulus basically?

src/aind_data_schema/core/acquisition.py
dbirman: I kind of really like that. We're under my 3 depth limit here since it's `acquisition.data_streams.summary`!

src/aind_data_schema/core/acquisition.py
dbirman: I like `experiment_type`

src/aind_data_schema/components/configs.py
dbirman: In the end I just put this back into Acquisition itself. Punting this into the future

src/aind_data_schema/core/acquisition.py
```diff
@@ -1,60 +1,218 @@
...
+        description="When a specific stimulus ends. This might be the same as the acquisition end time.",
+    )
+    stimulus_name: str = Field(..., title="Stimulus name")
+    code: Optional[Code] = Field(
+        default=None,
+        title="Code or script",
+        description="Custom code/script used to control the behavior/stimulus and parameters",
+    )
+    stimulus_modalities: List[StimulusModality] = Field(..., title="Stimulus modalities")
+    summary: Optional[EpochSummary] = Field(default=None, title="Summary")
```
saskiad, line 170: I think a different field name might be clearer - I worry this is interpreted in the same vein as "session_type" - like a description of the epoch. E.g. for people searching for this information. EpochOutput? PerformanceMetrics? (that's our title for output_parameters, but that can change)

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
saskiad: You need to add the `fiber_name` back since the config uses the patch_cord name. Need to know what it's connected to.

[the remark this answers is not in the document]
dbirman: Do these change day-to-day? If not, this should go into a `Connection` object when I get to that PR

src/aind_data_schema/components/configs.py
saskiad: I don't think FOV/SlapFOV can work that way with the microscope. There's barely a microscope device to those rigs.
Regardless, I can move on. There are a handful of fields in here that are about mouse position and scale/rotation factors that aren't configurations of the scanner. But it's fine.

saskiad (commented):
I can't tell where it is right now because I think you've been editing while I've been reviewing?

src/aind_data_schema/components/configs.py
dbirman: It doesn't need to be called "Microscope", "ImagingDevice" or something else is fine, the point is that the pattern of 1:1 device (in instrument): config (in acquisition) should be the same across all of these

saskiad: My point isn't about the name microscope - there isn't the same singular concept of an imaging device for those rigs. The fields in the FOV class aren't about a single device.

saskiad (approved):

PR #1292, opened 2025-03-02
feat: Composability

dbirman:
This PR adds the ability to do `metadata = metadata1 + metadata2` for the core files, where it's possible to do this.

- [x] quality_control
- [ ] acquisition
- [ ] instrument
- [ ] procedures


saskiad (commented):
looks good to me

src/aind_data_schema/core/procedures.py
```diff
@@ -703,3 +704,19 @@ class Procedures(DataCoreModel):
...
+
+    def __add__(self, other: "Procedures") -> "Procedures":
+        """Combine two Procedures objects"""
+
+        if not self.schema_version == other.schema_version:
+            raise ValueError("Schema versions must match to combine Procedures")
+
+        if not self.subject_id == other.subject_id:
+            raise ValueError("Subject IDs must match to combine Procedures objects.")
+
```
saskiad, line 716: probably want a similar check on specimen ids.

dbirman: There's no specimen_id in procedures. If you mean to check individual specimen procedures against each other then that doesn't make sense to me, since multiple specimens can be generated from one subject.

saskiad: But data can only be collected from one specimen. So the specimen_id in the specimen procedures need to match.

saskiad (approved):

PR #1296, opened 2025-03-03
BREAKING CHANGE: InjectionDynamics

dbirman:
This PR introduces a new concept to the `Injection` class. `InjectionDynamics` wrap the `volume`, `rate`, and `duration` together into a single class that is shared across all injections. The dynamics comes in because the field `Injection.dynamics` accepts a `List[InjectionDynamics]`, allowing people to describe in as much detail as they want the pattern of an injection -- if it's important. All existing injections can probably be ported simply as bolus injections.

In addition I added a concept `InjectionProfile` which provides an english-language description of the type of injection. I gave it three options bolus/continuous/pulsed for different types of injection I've seen. There are probably more. A validator prevents users from putting multiple injection events in while describing it as a bolus injection. 

I also fixed several tests that turned out to not be working at all.


src/aind_data_schema/core/procedures.py
```diff
@@ -426,6 +434,17 @@ class NonViralMaterial(Reagent):
     )
 
 
+class InjectionDynamics(DataModel):
+    """Description of the volume and rate of an injection"""
+
+    volume: Decimal = Field(..., title="Injection volume")
```
saskiad, line 440: I think volume has to be optional. Currently we use volume for most injection types except iontophoresis which is defined by current. 
So also add current as a field here.

src/aind_data_schema/core/procedures.py
```diff
@@ -479,23 +502,16 @@ class BrainInjection(Injection):
...
 
-    injection_volume: List[Decimal] = Field(
-        ...,
-        title="Injection volume (nL)",
-        description="Injection volume, one value per location",
-    )
-    injection_volume_unit: VolumeUnit = Field(VolumeUnit.NL, title="Injection volume unit")
-
-    @field_validator("injection_volume")
-    def check_dv_and_vol_list_lengths(cls, v, info: ValidationInfo):
```
saskiad, line 490: what's our plan for multiple depths?

src/aind_data_schema/core/procedures.py
```diff
@@ -434,27 +453,31 @@ class Injection(DataModel):
     ] = Field(..., title="Injection material", min_length=1)
     recovery_time: Optional[Decimal] = Field(default=None, title="Recovery time")
     recovery_time_unit: Optional[TimeUnit] = Field(default=None, title="Recovery time unit")
-    injection_duration: Optional[Decimal] = Field(default=None, title="Injection duration")
-    injection_duration_unit: Optional[TimeUnit] = Field(default=None, title="Injection duration unit")
+    dynamics: List[InjectionDynamics] = Field(
+        ..., title="Injection dynamics", description="List of injection events, pattern should match the profile"
+    )
+    profile: InjectionProfile = Field(..., title="Injection profile")
```
saskiad, line 459: would it make sense for profile to be a field inside dynamics?

src/aind_data_schema/core/procedures.py
saskiad: ah, this is why injection profile is a list?

saskiad (changes requested):
a few small things

src/aind_data_schema/core/procedures.py
```diff
@@ -434,27 +453,31 @@ class Injection(DataModel):
...
 
-    injection_volume: Decimal = Field(..., title="Injection volume (uL)")
-    injection_volume_unit: VolumeUnit = Field(default=VolumeUnit.UL, title="Injection volume unit")
     injection_eye: Side = Field(..., title="Injection eye")
 
 
 class IntraperitonealInjection(Injection):
     """Description of an intraperitoneal injection procedure"""
 
-    time: Optional[AwareDatetimeWithDefault] = Field(default=None, title="Injection time")
```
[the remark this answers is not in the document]
dbirman: Added back

src/aind_data_schema/core/procedures.py
dbirman: moved

saskiad (approved):

PR #1299, opened 2025-03-04
BREAKING CHANGE: Validate that species matches background_strain, use new StrainModel classes

dbirman:
This PR updates `Subject` to use the new `Strain` class from aind-data-schema-models and adds a validator to ensure that the `Subject.species.name` field matches what is in the `Subject.background_strain.species`

I also had to fix a field_validator that was suddenly getting `None` somehow, changing it to a model_validator fixed the issue. I'm not sure how validators interact but it seems to have been caused by something in that interaction.


src/aind_data_schema/core/metadata.py
```diff
@@ -332,6 +332,8 @@ def create_metadata_json(
 ) -> dict:
     """Creates a Metadata dict from dictionary of core schema fields."""
     # Extract basic parameters and non-corrupt core schema fields
+
+    print("\n\n\n\n\n")
```
dyf, line 336: stray print

dyf (changes requested):

src/aind_data_schema/core/metadata.py
dbirman: fixed

dyf (approved):

PR #1300, opened 2025-03-05
BREAKING CHANGE: Validate subject/specimen ID values in Procedures

dbirman:
This PR adds two validators to enforce that that the SpecimenProcedure.specimen_id field needs to both be an extension of the Procedure.subject_id and needs to be identical across all procedures. Test coverage added.

This wasn't true in one of the examples, I fixed it there.

I also refactored how this works so that a shared function is used to check whether a specimen_id matches with a subject_id, since we were already doing this in `Acquisition` and I suspect we might change the rules later. With the shared function future maintenance on this will be a bit smoother.

### DIFF
--- a/examples/ophys_procedures.py
+++ b/examples/ophys_procedures.py
@@ -116,7 +116,7 @@ p = Procedures(
     specimen_procedures=[
         SpecimenProcedure(
             procedure_type="Immunolabeling",
-            specimen_id="625100-672640",
+            specimen_id="625100_001",
             start_date="2023-06-09",
             end_date="2023-06-12",
             experimenters=[Person(name="Scientist Smith")],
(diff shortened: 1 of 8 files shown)


examples/ophys_procedures.py
```diff
@@ -116,7 +116,7 @@
     specimen_procedures=[
         SpecimenProcedure(
             procedure_type="Immunolabeling",
-            specimen_id="672640",
+            specimen_id="625100-672640",
```
saskiad, line 119: can we make this comply with our system: "625100_001"

examples/ophys_procedures.py
```diff
@@ -139,7 +139,7 @@
         ),
         SpecimenProcedure(
             procedure_type="Immunolabeling",
-            specimen_id="672640",
+            specimen_id="625100-672640",
```
saskiad, line 142: ditto

saskiad (approved):
Looks good but will you just have the example follow our naming convention.

PR #1311, opened 2025-03-11
BREAKING CHANGE: Refactoring to CoordinateSystem, Coordinate, and RelativePosition

dbirman:
This PR refactors the `coordinates.py` file to use a standardized system for supporting coordinate systems. It also adds support for defining relative positions in space, for devices that don't need specific positions.

New concepts:

- CoordinateSystem -- an origin coordinate and a set of axes pointing in defined directions
- CoordinateSystemLibrary -- a set of pre-defined coordinate systems covering most use cases, e.g. BREGMA_ARI for Bregma origin with the X axis pointing anterior, Y axis pointing right, and Z axis pointing inferior. 
- RelativePosition -- a vague position in space, e.g. anterior, left, superior. For use with some devices when specific positions are not available
- Transform -- an exact position in space, composed of a Translation, Rotation, Scale, or an AffineTransformMatrix
- CoordinateTransform -- a transformation from one coordinate system to another, this is not currently used but will reappear after I refactor our ephys definitions

Some useful notes: units are generally attached to the systems, not the coordinates in a system. AxisNames have been generally simplified to a single concept defined in coordinates.py. Origins have been merged into a single origin concept in coordinates.py. 

Models that used to define separate coordinates (ccf_coordinate_ap, etc) now group them together into a Coordinate or SurfaceCoordinate.

Models that used to define positions of things in space now use the Transform or RelativePosition classes.

Models can define their own coordinate systems implicitly, see for example how `ManipulatorConfig.manipulator_coordinates` get defined in `test_config.py`.

### Additional cleanup
Some additional cleanup happened while I was doing this refactor, I took notes out of a few procedures that shouldn't have had them. I took some debug print statements that snuck in a previous PR out of the schema_version_bump code because I was tired of all the unnecessary prints happening while testing.


examples/mri_acquisition.py
```diff
@@ -47,10 +47,15 @@
     echo_time=Decimal(5.33333333333333),
     effective_echo_time=Decimal("10.6666666666666998253276688046753406524658203125"),
     repetition_time=Decimal("500.0"),
-    vc_orientation=Rotation3dTransform(rotation=[1.0, 0.0, 0.0, 0.0, 0.0, -1.0, 0.0, 1.0, 0.0]),
```
saskiad, line 50: We should take a second look at how this works for mri and make sure that we can adapt the mapper to conform to this. I think these values came straight from the scanner, so we need to make sure we know how to turn a rotation array shape 9 into this array shape 3.

src/aind_data_schema/components/configs.py
```diff
@@ -201,43 +205,59 @@ class DomeModule(DeviceConfig):
...
-        title="Targeted CCF coordinates",
+
+    atlas_coordinates: Optional[List[Coordinate]] = Field(
+        default=None,
+        title="Targeted coordinates in the Acquisition Atlas",
     )
-    manipulator_coordinates: Coordinates3d = Field(
-        ...,
-        title="Manipulator coordinates",
+    manipulator_coordinates: Optional[List[SurfaceCoordinate]] = Field(
```
saskiad, line 229: Which coordinates are required? Do we allow recordings without coordinates?

src/aind_data_schema/components/configs.py
```diff
@@ -201,43 +205,59 @@ class DomeModule(DeviceConfig):
...
+        title="Manipulator local axis positions",
     )
-    surface_z: Optional[Decimal] = Field(default=None, title="Surface z")
-    surface_z_unit: Optional[SizeUnit] = Field(default=None, title="Surface z unit")
     dye: Optional[str] = Field(default=None, title="Dye")
     implant_hole_number: Optional[int] = Field(default=None, title="Implant hole number")
 
+    @model_validator(mode="after")
+    def validate_len_coordinates(self):
+        """Check that the length of atlas_ maipulator_ and manipulator_axis_ coordinates are the same"""
```
saskiad, line 242: fix typo

src/aind_data_schema/components/coordinates.py
```diff
@@ -1,171 +1,477 @@
...
-    """Image axis name"""
+    """Axis name"""
 
     X = "X"
     Y = "Y"
     Z = "Z"
+    AP = "AP"
+    ML = "ML"
+    SI = "SI"
+    DEPTH = "Depth"
```
saskiad, line 47: The "depth" option intrigues me. What's the use case for this? Should height and width be in here too?

src/aind_data_schema/components/coordinates.py
```diff
@@ -1,171 +1,477 @@
...
+class RelativePosition(DataModel):
+    """Relative position in a coordinate system"""
+
+    position: List[AnatomicalRelative] = Field(..., title="Relative position")
+
+
+class Coordinate(DataModel):
+    """A coordinate in a brain (CoordinateSpace) or atlas (AtlasSpace)
+
+    Angles can be optionally provided
```
saskiad, line 314: might help to explain what the angles communicate - it's actually a bit unclear to me because I assume coordinate is for a point, and so not sure what the angle would communicate for that point. This is for when the coordinate is being employed to describe the position of an object, right? But should that be smooshed into coordinate (conceptually) or should that the position be a class that combines coordinate and angle?

src/aind_data_schema/components/coordinates.py
```diff
@@ -1,171 +1,477 @@
...
+    transforms: List[
+        Annotated[
+            Union[Translation, Rotation, Scale, AffineTransformMatrix, NonlinearTransform],
+            Field(discriminator="object_type"),
+        ]
+    ] = Field(..., title="Transform")
+
+
+class RelativePosition(DataModel):
+    """Relative position in a coordinate system"""
```
saskiad, line 306: I think this needs a different description because this isn't the position in a coordinate system, this is a relative description of position without a defined coordinate system.

src/aind_data_schema/components/coordinates.py
```diff
@@ -1,171 +1,477 @@
...
     )
-    unit: SizeUnit = Field(SizeUnit.UM, title="Axis physical units")
 
 
-class RelativePosition(DataModel):
-    """Position and rotation of a device in a rig or instrument"""
+class CoordinateSystemLibrary:
+    """Library of common coordinate systems"""
+
+    BREGMA_ARI = CoordinateSystem(
```
saskiad, line 342: what does ARI stand for?

src/aind_data_schema/components/coordinates.py
```diff
@@ -1,171 +1,477 @@
...
+        origin=Origin.ORIGIN,
+        axis_unit=SizeUnit.UM,
+        axes=[
+            Axis(name=AxisName.X, direction=Direction.LR),
+            Axis(name=AxisName.Y, direction=Direction.AP),
+            Axis(name=AxisName.Z, direction=Direction.SI),
+        ],
+    )
+
+    SPIM_YXZ = CoordinateSystem(
```
saskiad, line 451: This is the same name as above, and the different is that x and z are inverted? Is that right? If so, it's that useful to have as a different coordinate system?

src/aind_data_schema/components/coordinates.py
```diff
@@ -1,171 +1,477 @@
...
+    MEDIAL = "Medial"
+    LATERAL = "Lateral"
+
+
+class Axis(DataModel):
+    """Linked direction and axis"""
+
+    name: AxisName = Field(..., title="Axis")
+    direction: Direction = Field(
+        ..., title="Direction", description="Use Other for device-defined axes, such as on manipulators"
```
saskiad, line 84: it helps to indicate that it's the direction for increased values

examples/mri_acquisition.py
saskiad: I assume we can use a variant of the transform function in the coordinates file for this?

src/aind_data_schema/components/coordinates.py
```diff
@@ -1,171 +1,477 @@
...
+        for i, angle in enumerate(self.angles):
+            # Get the angle, convert if needed
+            if self.angles_unit == AngleUnit.DEG:
+                angle = math.radians(angle)
 
-    translation: List[Decimal] = Field(..., title="3D translation parameters", min_length=3, max_length=3)
+            angles.append(angle)
 
+            # Get the axis order
+            axes += "xyz"[self.order[i]]
```
saskiad, line 176: I don't understand what this is doing ... this isn't used in an example is it? Maybe you can just unpack it for me quickly.

examples/mri_acquisition.py
dbirman: I'm pretty sure it translates 3 elements at a time into a 3x3 matrix (i.e. first 3 values = first row, etc) which would make that 1x9 array a 90 degree rotation around the X axis, so that makes sense. But yes whatever mapper handles MRI data will need to handle the mapping correctly in the future.

I'm not sure if these vc_ fields are generic across scan types though, it would be good to confirm that we're making a scan configuration that will work for other types of scans in the future.

There's also a typo here I didn't copy the rotation matrix values properly into the new version.

src/aind_data_schema/components/configs.py
dbirman: I set them all as optional because I'm not sure. I think it would make sense to at least make atlas_coordinates and manipulator_coordinates required.

Knowing the axis positions on the manipulator isn't that useful, I'm not sure why they asked for that originally -- it's basically just a rotation of the manipulator_coordinates.

src/aind_data_schema/components/coordinates.py
dbirman: It's still a position in a coordinate system since you need to know the Origin

src/aind_data_schema/components/coordinates.py
dbirman: Maybe you've seen "LAS" before? Shorthands for the axis directions. Anterior/Right/Inferior here

src/aind_data_schema/components/coordinates.py
dbirman: These are taken from conversation with Mike and Camilo, they said these are the two standard brain orientations in which SPIM images are taken (i.e. for YXZ the brain is rotated so that in the image the Y axis is AP, X is ML, Z is SI). Another way to name this would be SPIM_RPI for right/posterior/inferior, maybe that's less confusing?

src/aind_data_schema/components/coordinates.py
dbirman: This maps the order the user provides to the string the scipy `from_euler` function uses to order rotations. So e.g. order=[0, 1, 2] maps to "xyz", while [2, 1, 0] maps to "zyx"

saskiad: yup, that seems pretty obvious now.

src/aind_data_schema/components/coordinates.py
dbirman: Added

saskiad (commented):
see comments

src/aind_data_schema/components/coordinates.py
```diff
@@ -1,171 +1,479 @@
...
+        origin=Origin.FRONT_CENTER,
+        axis_unit=SizeUnit.MM,
+        axes=[
+            Axis(name=AxisName.X, direction=Direction.LR),
+            Axis(name=AxisName.Y, direction=Direction.BT),
+            Axis(name=AxisName.Z, direction=Direction.BF),
+        ],
+    )
+
+    MONITOR_SIPE = CoordinateSystem(
```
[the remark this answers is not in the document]
saskiad: or SIPE_MM SIPE_UM. I think naming them after specific devices is not a great idea.

src/aind_data_schema/components/configs.py
dbirman: I set manipulator coordinates required and kept the other two optional

src/aind_data_schema/components/coordinates.py
dbirman: Added

src/aind_data_schema/components/coordinates.py
dbirman: Whoops yeah that's a typo. Changed these to RPI/LPS

src/aind_data_schema/components/coordinates.py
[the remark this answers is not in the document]
dbirman: Simplified down to just a single DEVICE_SIPE with mm units

docs/source/example_workflow/example_workflow.py
```diff
@@ -109,11 +113,17 @@
...
-                        # multiple injection volumes at different depths are allowed, but that's not happening here
-                        injection_coordinate_depth=[float(coords[2])],
+                        coordinates=[
+                            SurfaceCoordinate(
+                                system_name="BREGMA_ARI",
+                                position=[coords[1], coords[0]],
+                                depth=float(coords[2]),
+                                angles=Rotation(
+                                    angles=[coords[3]],
+                                    order=[0],
```
dyf, line 123: having to specify a trivial order here is weird

examples/aibs_smartspim_instrument.py
```diff
@@ -180,6 +183,7 @@
 inst = Instrument(
     instrument_id="440_SmartSPIM2_20231004",
     modification_date=datetime.date(2023, 10, 4),
+    coordinate_system=CoordinateSystemLibrary.DEFAULT,
```
dyf, line 186: let's come up with a better name than "DEFAULT"

examples/ephys_acquisition.py
```diff
@@ -12,17 +12,113 @@
...
+            position=[8150, 3250, 7800],
+        ),
+    ],
+    device_name="Ephys_assemblyA",
+    coordinate_transform="behavior/calibration_info_np2_2023_04_24.py",
+    primary_targeted_structure=CCFStructure.LGD,
+    manipulator_axis_positions=[
+        Coordinate(
+            system_name="BREGMA_ARI",
+            position=[8422, 4205, 11087.5],
```
dyf, line 42: should we have "extent" validators to ensure that coordinates match axis ranges for a given coordinate system?

examples/ephys_acquisition.py
```diff
@@ -12,17 +12,113 @@
...
+        ),
+    ],
+    manipulator_coordinates=[
+        SurfaceCoordinate(
+            system_name="BREGMA_ARI",
+            position=[5000, 5000, 0],
```
dyf, line 48: why is depth not an axis?

examples/ephys_acquisition.py
```diff
@@ -12,17 +12,113 @@
...
+    rotation_angle=0,
+    device_name="Stick_assembly_1",
+    arc_angle=-180,
+    module_angle=-180,
+    notes="did not record angles, did not calibrate.",
+)
+stick_config_2 = DomeModule(
+    rotation_angle=0,
+    device_name="Stick_assembly_2",
+    arc_angle=-180,
```
dyf, line 103: no coordinate system here?

examples/exaspim_acquisition.py
```diff
@@ -20,6 +20,52 @@
 # script will be used as default
 t = datetime(2022, 11, 22, 8, 43, 00, tzinfo=timezone.utc)
 
+tile_scale = Scale(
+    scale=[0.748, 0.748, 1],
+)
+transform = Transform(system_name="SPIM_RPI", transforms=[tile_scale])
```
dyf, line 26: shouldn't this be a CoordinateTransform? the one that defines input and output system

examples/fip_ophys_instrument.py
```diff
@@ -8,12 +8,16 @@
...
+from aind_data_schema.components.coordinates import RelativePosition, AnatomicalRelative, CoordinateSystemLibrary
 
 bonsai_software = Software(name="Bonsai", version="2.5")
 
 camera_assembly_1 = d.CameraAssembly(
     name="BehaviorVideography_FaceSide",
-    camera_target=d.CameraTarget.FACE_SIDE_LEFT,
+    target=d.CameraTarget.FACE,
+    position=RelativePosition(
+        position=[AnatomicalRelative.LEFT],
```
dyf, line 19: I'm not clear on what this is communicating. I would guess we are trying to communicate something like this: this camera assembly is XYZ relative to the instrument coordinate system. It could be that the `position` field name is misleading?

examples/mri_acquisition.py
```diff
@@ -47,10 +47,14 @@
...
-    vc_position=Translation3dTransform(translation=[-6.1, -7.0, 7.9]),
+    vc_orientation=AffineTransformMatrix(
```
dyf, line 50: I think we can ditch the word `Matrix` , just `Affine` is fine, to be consistent with `Scale` `Rotation` etc

examples/procedures.py
```diff
@@ -16,71 +16,97 @@
...
+            recovery_time=0,
+            instrument_id=None,
+            coordinates=[
+                SurfaceCoordinate(
+                    system_name="SurgerySystem",
+                    position=[-0.85, -3.8, 0],
+                    depth=3.3,
+                    angles=Rotation(
+                        angles=[0, 10, 0],
+                        order=[0, 1, 2],
```
dyf, line 84: still don't think we want to expose `order`

src/aind_data_schema/components/configs.py
```diff
@@ -335,11 +358,11 @@ class MRIScan(DeviceConfig):
     repetition_time: Decimal = Field(..., title="Repetition time (ms)")
     repetition_time_unit: TimeUnit = Field(default=TimeUnit.MS, title="Repetition time unit")
     # fields required to get correct orientation
-    vc_orientation: Optional[Rotation3dTransform] = Field(default=None, title="Scan orientation")
-    vc_position: Optional[Translation3dTransform] = Field(default=None, title="Scan position")
+    vc_orientation: Optional[AffineTransformMatrix] = Field(default=None, title="Scan orientation")
```
dyf, line 361: why did this switch from a rotation to affine?

dyf: seems like we probably want to combine `vc_orientation` and `vc_position` into `vc_transform`?

src/aind_data_schema/components/tile.py
```diff
@@ -39,17 +33,7 @@ class Channel(DataModel):
...
-            Union[
-                Scale3dTransform,
-                Translation3dTransform,
-                Rotation3dTransform,
-                Affine3dTransform,
-            ],
-            Field(discriminator="object_type"),
-        ]
-    ] = Field(..., title="Tile coordinate transformations")
+    transforms: Transform = Field(..., title="Tile coordinate transformations")
```
dyf, line 36: let's use CoordinateTransform for this (the one w/ input and output)

src/aind_data_schema/core/procedures.py
```diff
@@ -259,8 +257,10 @@ class Sectioning(DataModel):
     section_thickness_unit: SizeUnit = Field(default=SizeUnit.MM, title="Section thickness unit")
     section_distance_from_reference: Decimal = Field(..., title="Section distance from reference")
     section_distance_unit: SizeUnit = Field(default=SizeUnit.MM, title="Distance unit")
-    reference_location: CoordinateReferenceLocation = Field(..., title="Reference location for distance measurement")
     section_strategy: SectionStrategy = Field(..., title="Slice strategy")
+
+    reference: Origin = Field(..., title="Reference origin")
```
dyf, line 262: should we use CoordinateSystem for this?

tests/resources/ephys_instrument.py
```diff
@@ -149,28 +154,32 @@
 
 microscope_1 = CameraAssembly(
     name="Stick_assembly_1",
-    camera_target=CameraTarget.BRAIN_SURFACE,  # NEEDS TO BE FILLED OUT
+    target=CameraTarget.BRAIN,
+    position=RelativePosition(position=[AnatomicalRelative.SUPERIOR]),
```
dyf, line 158: still confused about this

dyf (changes requested):

examples/exaspim_acquisition.py
dyf: acquisition needs a coordinatesystem

examples/fip_ophys_instrument.py
dyf: let's call the inside one `relative_position`

src/aind_data_schema/components/tile.py
dbirman: Done

examples/ephys_acquisition.py
dbirman: Will do this separately in the ephys refactor

examples/ephys_acquisition.py
dbirman: We can't do extent validators for the coordinate systems because they don't have a size -- but we can do them for Atlas. But it's complicated to add them because the code needs to pull the extent as part of this "recurse all fields" thing. If we really want this lets open a separate ticket for it, I want this merged so I can move on to the all the small (easy) wins that are left for 2.0

src/aind_data_schema/components/coordinates.py
saskiad: oh I meant in the docstring, but this work too.

src/aind_data_schema/components/devices.py
```diff
@@ -154,13 +152,9 @@ class CameraTarget(str, Enum):
...
-    BOTTOM = "Bottom"
-    BRAIN_SURFACE = "Brain surface"
+    BRAIN = "Brain"
     EYE = "Eye"
-    FACE_BOTTOM = "Face bottom"
-    FACE_FORWARD = "Face forward"
-    FACE_SIDE_LEFT = "Face side left"
-    FACE_SIDE_RIGHT = "Face side right"
+    FACE = "Face"
     SIDE = "Side"
```
saskiad, line 158: I presume SIDE should be remove for the same reason as BOTTOM

src/aind_data_schema/components/devices.py
```diff
@@ -782,7 +785,9 @@ class RewardSpout(Device):
     side: SpoutSide = Field(..., title="Spout side", description="If Other use notes")
     spout_diameter: Decimal = Field(..., title="Spout diameter (mm)")
     spout_diameter_unit: SizeUnit = Field(default=SizeUnit.MM, title="Spout diameter unit")
-    spout_position: Optional[RelativePosition] = Field(default=None, title="Spout stage position")
+    position: Optional[Annotated[Union[Transform, RelativePosition], Field(discriminator="object_type")]] = Field(
```
saskiad, line 788: The annotated union allows us to have both a specific transform and a relative position, right? Because I think ideally we want both for things like camera and monitor. One to say the camera targets the eye and it is "right front" and the other to provide precise position information so we can use the eye tracking data. Ditto for the monitor here. But my understanding is that we can do that here, yes?

dbirman: The Union is equivalent to an OR operator so you can only do one here. My rationale was that if you have precise position data you don't need the relative position, you can always infer it downstream.

src/aind_data_schema/components/devices.py
dbirman: Thanks, missed that. Removed

src/aind_data_schema/components/tile.py
```diff
@@ -39,17 +33,7 @@ class Channel(DataModel):
...
-            Union[
-                Scale3dTransform,
-                Translation3dTransform,
-                Rotation3dTransform,
-                Affine3dTransform,
-            ],
-            Field(discriminator="object_type"),
-        ]
-    ] = Field(..., title="Tile coordinate transformations")
+    coordinate_transform: CoordinateTransform = Field(..., title="Tile coordinate transformations")
```
saskiad, line 36: I'm confused by this. What are the input and output fields for a tile?

src/aind_data_schema/core/instrument.py
```diff
@@ -105,15 +105,13 @@ class Instrument(DataCoreModel):
...
-        title="CCF coordinate transform",
-        description="Path to file that details the CCF-to-lab coordinate transform",
-    )
-    origin: Optional[Origin] = Field(default=None, title="Origin point for instrument position transforms")
-    instrument_axes: Optional[List[Axis]] = Field(default=None, title="Instrument axes", min_length=3, max_length=3)
     modalities: List[Modality.ONE_OF] = Field(..., title="Modalities")
+    calibrations: Optional[List[Calibration]] = Field(default=None, title="Full calibration of devices")
+
+    # coordinate system
+    coordinate_system: CoordinateSystem = Field(..., title="Coordinate system")
```
saskiad, line 112: What is the coordinate system for a SPIM? 
We define the coordinate system for the InVitroImagingConfig, but I'm not sure we have a coordinate system for the microscope itself.

saskiad: it'll need to be required through validators

tests/resources/ephys_instrument.py
```diff
@@ -1,11 +1,16 @@
-"""Generates an example JSON file for an ephys instrument"""
+"""Generates an example JSON file for an ephys rig"""
```
saskiad, line 1: why are we changing this to rig?

tests/resources/spim_instrument.py
```diff
@@ -245,6 +246,7 @@
     instrument_id="440_SmartSPIM1_20231004",
     manufacturer=Organization.LIFECANVAS,
     modification_date=datetime.date(2023, 10, 4),
+    coordinate_system=CoordinateSystemLibrary.BREGMA_ARI,
```
saskiad, line 249: yeah, ditto the comment on this above. I don't think Bregma makes sense as an origin for in vitro imaging. I don't think we can have a coordinate system for the SPIMs.

examples/aibs_smartspim_instrument.py
```diff
@@ -180,6 +183,7 @@
 inst = Instrument(
     instrument_id="440_SmartSPIM2_20231004",
     modification_date=datetime.date(2023, 10, 4),
+    coordinate_system=CoordinateSystemLibrary.BREGMA_ARI,
```
saskiad, line 186: see comments below

src/aind_data_schema/core/instrument.py
saskiad: ahh SPIM_RPI. And the Origin as "Origin" means just whatever the origin of the sample is? 
I think that's the only thing we can use for SPIM - the BREGMA ones in the SPIM examples don't make sense.

examples/ophys_procedures.py
```diff
@@ -92,11 +94,10 @@
...
                             ),
                             targeted_structure=CCFStructure.VTA,
-                            angle=0,
-                            stereotactic_coordinate_ap=-3.05,
-                            stereotactic_coordinate_ml=-0.6,
-                            stereotactic_coordinate_dv=-4,
-                            stereotactic_coordinate_reference="Bregma",
+                            coordinate=Coordinate(
+                                system_name="BREGMA_ARID",
+                                position=[-600, -3050, 0, -4.2],
```
saskiad, line 99: -4.2 seems like the wrong unit.

saskiad (changes requested):
Really close - I think there are a couple questions to make sure we're aligned on, the rest are just minor

src/aind_data_schema/components/tile.py
saskiad: I see the use of this in the example. I guess I get it. I don't love it.

src/aind_data_schema/components/devices.py
saskiad: We need to be able to have both. your argument for taking the side of the eye (e.g. ) out of the target for the camera is so people can easily parse what each camera is doing, but the precise location is also needed to be able to use the data. I get the use case of when we don't have precise location, but when we do have that, we'll want both.

saskiad: If we made this List[Annotated[Union]] then we could have both - is that what we need?

saskiad: Also, it shouldn't be optional for things like spout and camera since the RelativePosition is super flexible.

src/aind_data_schema/components/tile.py
saskiad: @dyf why? I don't understand this.

src/aind_data_schema/components/devices.py
dbirman: Refactored

tests/resources/ephys_instrument.py
dbirman: Fixed

src/aind_data_schema/core/instrument.py
dbirman: Fixed

tests/resources/spim_instrument.py
dbirman: Fixed

examples/aibs_smartspim_instrument.py
dbirman: Fixed

examples/ophys_procedures.py
dbirman: Fixed

src/aind_data_schema/components/devices.py
dbirman: Split to position/relative_position

src/aind_data_schema/components/devices.py
```diff
@@ -779,25 +780,19 @@ class Monitor(Device):
 class RewardSpout(Device):
     """Description of a reward spout"""
 
-    side: SpoutSide = Field(..., title="Spout side", description="If Other use notes")
     spout_diameter: Decimal = Field(..., title="Spout diameter (mm)")
     spout_diameter_unit: SizeUnit = Field(default=SizeUnit.MM, title="Spout diameter unit")
-    spout_position: Optional[RelativePosition] = Field(default=None, title="Spout stage position")
+
+    relative_position: Optional[List[AnatomicalRelative]] = Field(default=None, title="Relative position")
```
saskiad, line 786: let's move these to the config instead of the device

tests/resources/ephys_instrument.py
saskiad: ? still says rig, but we can fix later

saskiad (approved):
small things. Otherwise I think this is good.

PR #1313, opened 2025-03-15
BREAKING CHANGE: remove processingsteps class

dbirman:
This PR removes ProcessingSteps. We will follow up with users about whether this information still needs to be stored somehow. If it does, we will move it into a generic setup that works for all users rather than have it attached to a Config class.


saskiad (approved):
We need to make sure we talk with the imaging teams about how they can communicate this outside of the metadata.

PR #1320, opened 2025-03-18
feat: add model version

### LINKED ISSUE #1319: model schema tweaks
Going over model schema usage with @camilolaiton in preparation for a huggingface release, we released we forgot to add a version field (model weights, unlike data assets, are expected to be versioned).
I'll submit an urgent PR for this today.

tmchartrand:
closes #1319, adds optional model version field.

### DIFF
--- a/src/aind_data_schema/core/model.py
+++ b/src/aind_data_schema/core/model.py
@@ -60,6 +60,7 @@ class Model(AindCoreModel):
 
     name: str = Field(..., title="Name")
     license: str = Field(..., title="License")
+    version: Optional[str] = Field(default=None, title="Version")
     developer_full_name: Optional[List[str]] = Field(default=None, title="Name of developer")
     developer_institution: Optional[Organization.ONE_OF] = Field(default=None, title="Institute where developed")
     modality: List[Modality.ONE_OF] = Field(..., title="Modality")


tmchartrand:
seems like the test environment needs an update of some kind...

saskiad (approved):
I think we had relegated this into the ModelArchitecture fields in the software class. But I agree that there should be an explicit top level version field.

saskiad:
> seems like the test environment needs an update of some kind...

I ran into this as well and have no idea what's going on. For me it wasn't worth figuring it out as we should probably be pushing to the 2.0 version rather than to dev at this stage...

tmchartrand:
@saskiad yeah i can submit a separate PR to 2.0, but camilo wanted to use it ASAP from dev also. looks like we can ignore the checks and still merge?

saskiad:
I don't think we're planning a release prior to 2.0

saskiad:
all of the example rig files have versions ahead of the actual rig version.

camilolaiton:
@tmchartrand it is fine that we use the current schemas and add a changelog where I could specify this is v1.0 or include it also in the notes. When 2.0 is released with this, I can include it for future models

saskiad:
@tmchartrand should we close this PR or do you want to point it to v2.0?

tmchartrand:
closing in favor of a broader model PR incorporating other changes too

PR #1323, opened 2025-03-20
1189 20 add connection class and remove daq

dbirman:
This PR refactors DAQChannel and COM into a `Connection` class with `ConnectionData` attached to it for each connection. I think it's pretty intuitive!


dbirman:
@dbirman ping Adam Glaser, check about exaspim_instrument example com ports

saskiad (approved):

PR #1325, opened 2025-03-21
BREAKING CHANGE: Support more standard craniotomy types

dbirman:
This PR refactors `Craniotomy` to better match the new coordinate system approach to and to allow for circular/square craniotomies that have a position on the brain. It still supports the existing SHIELD implants.

### DIFF
--- a/src/aind_data_schema/core/procedures.py
+++ b/src/aind_data_schema/core/procedures.py
@@ -102,8 +102,8 @@ class CraniotomyType(str, Enum):
     """Name of craniotomy Type"""
 
     DHC = "Dual hemisphere craniotomy"
-    THREE_MM = "3 mm"
-    FIVE_MM = "5 mm"
+    CIRCLE = "Circle"
+    SQUARE = "Square"
     VISCTX = "Visual Cortex"
     WHC = "Whole hemisphere craniotomy"
     OTHER = "Other"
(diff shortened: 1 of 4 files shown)


src/aind_data_schema/core/procedures.py
```diff
@@ -339,12 +339,14 @@ class Craniotomy(DataModel):
 
     protocol_id: str = Field(..., title="Protocol ID", description="DOI for protocols.io")
     craniotomy_type: CraniotomyType = Field(..., title="Craniotomy type")
-    craniotomy_hemisphere: Optional[Side] = Field(default=None, title="Craniotomy hemisphere")
+    position: Union[Coordinate, List[AnatomicalRelative]] = Field(..., title="Craniotomy position")
```
saskiad, line 342: what's the position for dual hemisphere craniotomy?

src/aind_data_schema/core/procedures.py
```diff
@@ -102,8 +102,8 @@ class CraniotomyType(str, Enum):
     """Name of craniotomy Type"""
 
     DHC = "Dual hemisphere craniotomy"
-    THREE_MM = "3 mm"
-    FIVE_MM = "5 mm"
+    CIRCLE = "Circle"
+    SQUARE = "Square"
```
saskiad, line 106: VISCTX should be removed. It should be a CIRCLE with a position and size.

saskiad (commented):
I think this looks good, but I need to understand how we use this for dual hemisphere craniotomy.

src/aind_data_schema/core/procedures.py
dbirman: I'll make it optional with a validator

src/aind_data_schema/core/procedures.py
dbirman: Removed

src/aind_data_schema/core/procedures.py
```diff
@@ -339,12 +338,24 @@ class Craniotomy(DataModel):
 
     protocol_id: str = Field(..., title="Protocol ID", description="DOI for protocols.io")
     craniotomy_type: CraniotomyType = Field(..., title="Craniotomy type")
-    craniotomy_hemisphere: Optional[Side] = Field(default=None, title="Craniotomy hemisphere")
+
+    position: Optional[Union[Coordinate, List[AnatomicalRelative]]] = Field(default=None, title="Craniotomy position")
+
+    size: Optional[float] = Field(default=None, title="Craniotomy size", description="Diameter or side length")
```
saskiad, line 344: let's add a validator requiring a size for a circle/square craniotomy.

saskiad (approved):
I do think we want a validator for the size field for circle/square. But otherwise I think this is good!

src/aind_data_schema/core/procedures.py
dbirman: Added

PR #1329, opened 2025-03-24
1135 add exposure and compression string to camera detector

dbirman:
This PR adds an optional compression field that takes a `Code` object as input. Users can specify their specific compression parameters in the generic parameters field.


saskiad (approved):
I'm okay with it, but does it really make sense to have this separate from the other fields in Detector?

PR #1332, opened 2025-03-24
refactor: LickSpoutConfig, AirPuffDevice, and AirPuffConfig

dbirman:
This PR removes the class `RewardSpout` and `RewardSpoutconfig` and replaces them with device-specific configurations for recording the valence and qualities of rewards. New classes:

- LickSpout
- LickSpoutConfig
- AirPuffDevice
- AirPuffConfig

### DIFF
--- a/src/aind_data_schema/components/configs.py
+++ b/src/aind_data_schema/components/configs.py
@@ -50,6 +50,15 @@ class StimulusModality(str, Enum):
     WHEEL_FRICTION = "Wheel friction"
 
 
+class Valence(str, Enum):
+    """Valence of a stimulus"""
+
+    POSITIVE = "Positive"
+    NEGATIVE = "Negative"
+    NEUTRAL = "Neutral"
+    UNKNOWN = "Unknown"
+
+
 class DeviceConfig(DataModel):
     """Parent class for all configurations"""
 
@@ -272,15 +281,18 @@ class LaserConfig(DeviceConfig):
 
 
 # Behavior components
-class RewardSolution(str, Enum):
-    """Reward solution name"""
+class Liquid(str, Enum):
+    """Solution names"""
 
     WATER = "Water"
     OTHER = "Other"
 
 
-class RewardSpoutConfig(DataModel):
-    """Reward spout acquisition information"""
+class LickSpoutConfig(DataModel):
+    """Lick spout acquisition information"""
+
+    solution: Liquid = Field(..., title="Solution")
+    solution_valence:  Valence = Field(..., title="Valence")
 
     relative_position: List[AnatomicalRelative] = Field(..., title="Initial relative position")
     position: Optional[Transform] = Field(default=None, title="Initial position")
@@ -290,23 +302,25 @@ class RewardSpoutConfig(DataModel):
         description="True if spout position changes during acquisition as tracked in data",
     )
 
-
-class RewardDeliveryConfig(DataModel):
-    """Description of reward delivery configuration"""
-
-    reward_solution: RewardSolution = Field(..., title="Reward solution", description="If Other use notes")
-    reward_spouts: List[RewardSpoutConfig] = Field(..., title="Reward spouts")
     notes: Optional[str] = Field(default=None, title="Notes", validate_default=True)
 
     @field_validator("notes", mode="after")
     def validate_other(cls, value: Optional[str], info: ValidationInfo) -> Optional[str]:
         """Validator for other/notes"""
 
-        if info.data.get("reward_solution") == RewardSolution.OTHER and not value:
+        if info.data.get("reward_solution") == Liquid.OTHER and not value:
             raise ValueError(
                 "Notes cannot be empty if reward_solution is Other. Describe the reward_solution in the notes field."
             )
         return value
+
+
+class AirPuffConfig(DataModel):
+    """ Air puff device configuration """
+
+    valence: Valence = Field(default=Valence.NEGATIVE, title="Valence")
+    relative_position: List[AnatomicalRelative] = Field(..., title="Initial relative position")
+    position: Optional[Transform] = Field(default=None, title="Initial position")
 
 
 class SpeakerConfig(DeviceConfig):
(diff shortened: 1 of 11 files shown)


src/aind_data_schema/components/configs.py
```diff
@@ -272,15 +281,18 @@ class LaserConfig(DeviceConfig):
 
 
 # Behavior components
-class RewardSolution(str, Enum):
-    """Reward solution name"""
+class Liquid(str, Enum):
+    """Solution names"""
```
saskiad, line 285: should we go ahead and add quinine if that's something they are planning on doing, and to have a clear example of a non-positive liquid

src/aind_data_schema/components/devices.py
```diff
@@ -787,11 +787,18 @@ class RewardSpout(Device):
...
 
 
-class RewardDelivery(DataModel):
-    """Description of reward delivery system"""
+class LickSpoutAssembly(DataModel):
+    """Description of multiple lick spouts, possibly mounted on a stage"""
 
-    stage_type: Optional[MotorizedStage] = Field(default=None, title="Motorized stage")
-    reward_spouts: List[RewardSpout] = Field(..., title="Water spouts")
+    reward_spouts: List[LickSpout] = Field(..., title="Water spouts")
```
saskiad, line 793: this should be lick_spouts for consistency, right? And title should be lick spouts. (never should have been water spouts)

src/aind_data_schema/components/configs.py
```diff
@@ -290,23 +302,26 @@ class RewardSpoutConfig(DataModel):
...
-                "Notes cannot be empty if reward_solution is Other. Describe the reward_solution in the notes field."
+                "Notes cannot be empty if LickSpoutConfig.reward_solution is Other."
+                "Describe the reward_solution in the notes field."
             )
-        return value
+        return values
+
+
+class AirPuffConfig(DataModel):
+    """Air puff device configuration"""
```
saskiad, line 320: per the teams thread, we need air pressure and duration as fields

src/aind_data_schema/components/devices.py
```diff
@@ -787,11 +787,18 @@ class RewardSpout(Device):
...
+    motorized_stage: Optional[MotorizedStage] = Field(default=None, title="Motorized stage")
+
+
+class AirPuffDevice(Device):
+    """Description of an air puff device"""
+
+    size: float = Field(..., title="Spout diameter")
```
saskiad, line 800: I think this is better as diameter than size

saskiad: we should maybe ask Kenta and Bruno whether we need more info for the Air Puff device. E.g. the equivalent of the solenoid. I don't want unnecessary things, but I also want to avoid needing to add a bunch of stuff later.

saskiad (changes requested):
see comments

saskiad:
I'm tempted to remove the `variable_position` boolean for the lick spout config. But not sure. There's value in knowing something isn't fixed and that it ends up as a piece of data, but it also feels weird in this way. any strong feelings?

src/aind_data_schema/components/configs.py
dougollerenshaw: That seems right. We should anticipate different reward solutions to avoid having people select "other". Is anyone using any other liquid rewards beyond water?

saskiad: Not to my knowledge .... but I didn't know about the air puffs either.

dbirman: Added a few I'm aware of

src/aind_data_schema/components/devices.py
dbirman: fixed

src/aind_data_schema/components/devices.py
dbirman: Swapped to diameter, for now I think we're okay here

src/aind_data_schema/components/configs.py
dbirman: Added

dbirman:
> I'm tempted to remove the `variable_position` boolean for the lick spout config. But not sure. There's value in knowing something isn't fixed and that it ends up as a piece of data, but it also feels weird in this way. any strong feelings?

I dropped it, I agree that it's probably not important to have in the metadata

src/aind_data_schema/components/devices.py
saskiad: still says `reward_spouts` and "Water spouts"

src/aind_data_schema/core/instrument.py
```diff
@@ -137,7 +137,7 @@ class Instrument(DataCoreModel):
             Union[
                 Monitor,
                 Olfactometer,
-                RewardDelivery,
+                LickSpoutAssembly,
```
saskiad, line 140: we also want to add AirPuffDevice here

saskiad: and for the imports

src/aind_data_schema/core/acquisition.py
```diff
@@ -32,7 +32,6 @@
     MousePlatformConfig,
     Stack,
     MRIScan,
-    RewardDeliveryConfig,
```
saskiad, line 35: don't we need the LickSpoutConfig and AirPuffConfig in the acquisition?

saskiad (changes requested):
The logic is all good - I think we need the configs in the acquisition and the airpuff device in the instrument. Or I'm not reading these files right.

src/aind_data_schema/components/devices.py
dbirman: Okay, actually fixed...

src/aind_data_schema/core/instrument.py
dbirman: Whoops yeah, added

src/aind_data_schema/core/acquisition.py
dbirman: Added

saskiad (approved):

PR #1334, opened 2025-03-24
refactor: Calibration and Subject.subject_details

dbirman:
This PR refactors some of the classes that were originally slated to become extensions.

- There is now a specific `WaterCalibration` class
- Subject details are now stored in `subject_details`, accepting either `MouseSubject` or `HumanSubject` detail objects.


src/aind_data_schema/components/subjects.py
```diff
@@ -0,0 +1,149 @@
...
+            if value.species.name != value.strain.species:
+                raise ValueError("The animal species and it's strain's species do not match")
+
+        return value
+
+
+class HumanSubject(DataModel):
+    """Description of a human subject"""
+
+    year_of_birth: int = Field(..., title="Year of birth")
```
saskiad, line 149: human subject should have sex too

saskiad: and also probably a source - where we got the tissue from

src/aind_data_schema/components/calibrations.py
```diff
@@ -0,0 +1,27 @@
...
+    description: str = Field(..., title="Description", description="Brief description of what is being calibrated")
+    input: GenericModelType = Field(GenericModel(), description="Calibration input", title="inputs")
+    output: GenericModelType = Field(GenericModel(), description="Calibration output", title="outputs")
+    notes: Optional[str] = Field(default=None, title="Notes")
+
+
+class WaterCalibration(Calibration):
+    """Calibration of a water delivery device"""
+
+    input: List[float] = Field(..., title="Input times", description="Length of time solenoid is open")
```
saskiad, line 24: Might be worth adding a fixed description in this class

src/aind_data_schema/components/calibrations.py
```diff
@@ -0,0 +1,27 @@
...
+
+    calibration_date: AwareDatetimeWithDefault = Field(..., title="Date and time of calibration")
+    device_name: str = Field(..., title="Device name", description="Must match a device name in instrument")
+    description: str = Field(..., title="Description", description="Brief description of what is being calibrated")
+    input: GenericModelType = Field(GenericModel(), description="Calibration input", title="inputs")
+    output: GenericModelType = Field(GenericModel(), description="Calibration output", title="outputs")
+    notes: Optional[str] = Field(default=None, title="Notes")
+
+
+class WaterCalibration(Calibration):
```
saskiad, line 21: Let's stick with the liquid instead of water (in a couple of places ehre)

saskiad (commented):
Couple quick comments.

src/aind_data_schema/components/subjects.py
dbirman: Whoops I didn't even consider that these were specimens

src/aind_data_schema/components/calibrations.py
dbirman: Done

src/aind_data_schema/components/calibrations.py
dbirman: Done

src/aind_data_schema/components/measurements.py
```diff
@@ -0,0 +1,48 @@
...
+    description: str = Field(..., title="Description", description="Brief description of what is being calibrated")
+    input: GenericModelType = Field(GenericModel(), description="Calibration input", title="inputs")
+    output: GenericModelType = Field(GenericModel(), description="Calibration output", title="outputs")
+    notes: Optional[str] = Field(default=None, title="Notes")
+
+
+class LiquidCalibration(Calibration):
+    """Calibration of a liquid delivery device"""
+
+    input: List[float] = Field(..., title="Input times", description="Length of time solenoid is open")
```
saskiad, line 25: I still think it would be useful to have a fixed description for the LiquidCalibration class

src/aind_data_schema/components/subjects.py
```diff
@@ -0,0 +1,155 @@
...
+        return v
+
+    @model_validator(mode="after")
+    def validate_genotype(value):
+        """Validator for mice genotype"""
+
+        if not hasattr(value, "species"):  # pragma: no cover
+            return value
+
+        if value.species is Species.MUS_MUSCULUS and value.genotype is None:
```
saskiad, line 130: We can make genotype a required field for `MouseSubject` and eliminate this validator

saskiad (approved):
Overall good - see a couple comments - key one being about the genotype for mouse subject - make that field required and then eliminate the validator.

src/aind_data_schema/components/measurements.py
dbirman: Added

src/aind_data_schema/components/subjects.py
dbirman: Done

src/aind_data_schema/components/measurements.py
```diff
@@ -0,0 +1,48 @@
...
+    input_unit: TimeUnit = Field(..., title="Input unit")
+    output: List[float] = Field(..., title="Output", description="Liquid output")
+    output_unit: VolumeUnit = Field(..., title="Output unit")
+
+
```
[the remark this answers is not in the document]
dbirman: Opened separate ticket

src/aind_data_schema/components/measurements.py
saskiad: just make it "liquid" instead of water for consistency.

src/aind_data_schema/components/measurements.py
```diff
@@ -0,0 +1,93 @@
...
+    UnitlessUnit,
+]
+
+
+class Calibration(DeviceConfig):
+    """Generic calibration class"""
+
+    calibration_date: AwareDatetimeWithDefault = Field(..., title="Date and time of calibration")
+    description: str = Field(..., title="Description", description="Brief description of what is being calibrated")
+    input: List[float | str] = Field(..., description="Calibration input", title="inputs")
```
saskiad, line 52: so this is a list of either float or str? (I might misunderstand this syntax)
So at no point can the user say what the input/output is. They can provide the units, but they aren't saying what it is. I think we need them to specify what they are.

saskiad (commented):
see comment

src/aind_data_schema/components/measurements.py
dbirman: Messaged on teams but copying here:
- x | y is equivalent to Union[x, y]
- Either they put it in the description or we split description into input_description, output_description. I'm fine either way but I think this isn't really a big concern. For most devices the units will probably be sufficient to know what was calibrated

PR #1335, opened 2025-03-25
refactor: remove recovery and instrument_id and ICM/ICV injections

dbirman:
See title


saskiad (approved):
Great! Let's remember to add lateral ventricle in our options in models

PR #1338, opened 2025-03-25
testing oligo probe stuff

### LINKED ISSUE #715: Procedure model for BARSeq
We don't yet have a BARSeq model

we need to add a model for BARSeq processes. I'm working with Yoh to specify exactly what we need and will update this ticket

saskiad:
closes #715


src/aind_data_schema/core/procedures.py
```diff
@@ -168,6 +168,27 @@ class InjectionProfile(str, Enum):
...
+
+
+class GeneProbes(DataModel):
+    '''Description of a set of oligonucleotide probes targeting a specific gene'''
+
+    gene: PIDName = Field(..., title="Gene name")
+    probes: List[OligoProbe] = Field(..., title="Probes")
+
+
+class OligoProbeSet(Reagent):
```
dbirman, line 185: Is this supposed to inherit from Reagent? I'm not sure I follow why that is. It seems like an OligoProbe might be a reagant itself, but the set of them seems strange to me.

src/aind_data_schema/core/procedures.py
```diff
@@ -168,6 +168,27 @@ class InjectionProfile(str, Enum):
...
+    '''set of probes used in BarSEQ'''
+
+    gene_probes: List[GeneProbes] = Field(..., title="Gene probes")
```
dbirman, line 188: If this isn't a reagent and it's just a list of these objects then I think it should be added like Antibody into the new `procedure_details` List as just GeneProbes, without a wrapper class.

src/aind_data_schema/core/procedures.py
```diff
@@ -183,16 +204,15 @@ class HCRReadout(Readout):
     initiator_name: str = Field(..., title="Initiator name")
 
 
-class OligoProbe(Reagent):
-    """Description of an oligonucleotide probe"""
+class OligoProbeWithReadout(Reagent):       # probably want a different name
```
dbirman, line 207: Maybe we can take Oligo out since it's already short for something and call this `GeneticStain` or something? Or `GeneTargetedStain`

src/aind_data_schema/core/procedures.py
saskiad: This barseq probe set is made as a set by a company and has a batch number that we want to associate with it.

src/aind_data_schema/core/procedures.py
saskiad: see above - we do need the reagent fields

dbirman:
@saskiad I fixed this branch so that it's up-to-date with the release-v2.0.0. If you want to merge it in time for 2.0 take a look at what still needs to be done?

src/aind_data_schema/core/procedures.py
saskiad: It's not a stain. 
Oligo is short for Oligonucleotide. Gene implies a full gene, while this is a short nucleotide sequence. E.g. we use ~40 different oligoprobes that all target the same gene in one OligoProbeSet
I think we want to keep OligoProbe - it's about conveying that it has a readout attached...

saskiad:
> @saskiad I fixed this branch so that it's up-to-date with the release-v2.0.0. If you want to merge it in time for 2.0 take a look at what still needs to be done?

It seems like the merge lost some things? I might just be lost, but I'm not seeing some of these pieces.

dbirman:
Again merged release branch in

src/aind_data_schema/core/procedures.py
dbirman: I left it as is

dbirman (approved):

dbirman:
OligoProbeStain (vs OligoProbeReadout)

dbirman:
or GeneticProbeReadout?

PR #1340, opened 2025-03-25
refactor: Move SpecimenProcedure details in procedure_details field

dbirman:
This PR refactors SpecimenProcedure to add a procedure_details field, and moves some otherwise Optional fields into it.

### DIFF
--- a/src/aind_data_schema/core/procedures.py
+++ b/src/aind_data_schema/core/procedures.py
@@ -288,7 +288,7 @@ class SpecimenProcedure(DataModel):
 
     procedure_details: List[Annotated[Union[
         HCRSeries,
-        List[Antibody],
+        Antibody,
         Sectioning,
     ], Field(discriminator="object_type")]] = Field(
         default=[],
@@ -305,8 +305,7 @@ class SpecimenProcedure(DataModel):
             isinstance(detail, HCRSeries) for detail in self.procedure_details
         )
         has_antibodies = any(
-            isinstance(detail, list) and all(isinstance(antibody, Antibody) for antibody in detail)
-            for detail in self.procedure_details
+            isinstance(detail, Antibody) for detail in self.procedure_details
         )
 
         if self.procedure_type == SpecimenProcedureType.OTHER and not self.notes:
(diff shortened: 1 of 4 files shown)


src/aind_data_schema/core/procedures.py
```diff
@@ -285,22 +285,37 @@ class SpecimenProcedure(DataModel):
     )
     protocol_id: List[str] = Field(..., title="Protocol ID", description="DOI for protocols.io")
     reagents: List[Reagent] = Field(default=[], title="Reagents")
-    hcr_series: Optional[HCRSeries] = Field(default=None, title="HCR Series")
-    antibodies: Optional[List[Antibody]] = Field(default=None, title="Immunolabeling")
-    sectioning: Optional[Sectioning] = Field(default=None, title="Sectioning")
+
+    procedure_details: List[
```
saskiad, line 289: I'm not sure this needs to be a List. A single procedure can't have these combined.
Oh, I guess there can be multiple antibodies used at one time, so that needs to be a list.
But we might need a validator that makes sure that we aren't mixing and matching these items. It's not like surgery where they can happen together.

src/aind_data_schema/core/procedures.py
```diff
@@ -285,22 +285,37 @@ class SpecimenProcedure(DataModel):
...
+        Annotated[
+            Union[
+                HCRSeries,
+                Antibody,
+                Sectioning,
+            ],
```
saskiad, line 295: I wonder if we do want to add Reagent to this Union. It's a generic concept for me, but might as well live in the details section.

src/aind_data_schema/core/procedures.py
```diff
@@ -285,22 +285,37 @@ class SpecimenProcedure(DataModel):
...
+
     notes: Optional[str] = Field(default=None, title="Notes")
 
     @model_validator(mode="after")
     def validate_procedure_type(self):
         """Adds a validation check on procedure_type"""
 
+        has_hcr_series = any(isinstance(detail, HCRSeries) for detail in self.procedure_details)
+        has_antibodies = any(isinstance(detail, Antibody) for detail in self.procedure_details)
+
```
saskiad, line 311: can you add same validator for sectioning?

saskiad (commented):
super small things

src/aind_data_schema/core/procedures.py
dbirman: Validator added

src/aind_data_schema/core/procedures.py
dbirman: Validator added

src/aind_data_schema/core/procedures.py
dbirman: Moved

saskiad (approved):

PR #1342, opened 2025-03-25
refactor: use Path and soft-validation

dbirman:
This PR refactors a few places in the schema where we had paths to local files to use the `PurePosixPath` object from pathlib. A validator gets called with `write_standard_file` that does soft validation to check if paths exist in the filesystem and warns the user if they don't exist.

Pretty benign right now but gives us an easy way to do real validation in the future, and makes it easier for downstream code to do validation. Hopefully this will also help people understand that we want relative paths.


tmchartrand (changes requested):
I think you've got all the actual paths here as far as I can tell. A few comments:

1. I think this should subclass Path just for clarity even if nothing is added (ie in case a field wants to use Path with different validation in the future for an inaccessible URI or something). AssetPath(Path) maybe?
2. Could add actual validation that the paths are relative and written as posix (convert absolute to relative?), while keeping the existence check as a delayed soft validation.
3. Might as well change this field to not use the word "Path" while we're at it: https://github.com/AllenNeuralDynamics/aind-data-schema/blob/fc493f4d0d9e35a0c829faa97233fa344771acd6/src/aind_data_schema/components/identifiers.py#L30

src/aind_data_schema/utils/validators.py
```diff
@@ -106,3 +108,23 @@ def recursive_get_all_names(obj: Any) -> List[str]:
...
+
+
+def recursive_check_paths(obj: Any, directory: Optional[Path] = None):
+    """
+    Recursively checks if any attribute in an object is a pathlib.Path and logs if it doesn't exist.
+    """
+    if isinstance(obj, Enum):
+        return
+
+    if isinstance(obj, Path):
```
tmchartrand, line 120: shouldn't this be AssetPath?

tmchartrand: maybe add a test_standard_path with a regular Path to confirm it is not checked.

src/aind_data_schema/components/identifiers.py
```diff
@@ -1,13 +1,24 @@
...
 
+class AssetPath(type(Path())):
+    """Relative path to a file"""
+
+    @classmethod
+    def __get_pydantic_core_schema__(cls, source_type, handler: GetCoreSchemaHandler):
+        """Correctly serialize/deserialize for pydantic"""
+        return core_schema.str_schema()
+
+
```
tmchartrand, line 21: not sure i follow why this is all needed rather than just AssetPath(Path), but if you're positive it's all good!

tmchartrand (commented):
looks good, but unless i'm missing something still doesn't enforce posix encoding or relative vs absolute path?

src/aind_data_schema/utils/validators.py
dbirman: Changed to check properly for PurePosixPath and convert

src/aind_data_schema/components/identifiers.py
dbirman: Just a pydantic inheritance thing

tests/test_identifiers.py
```diff
@@ -21,5 +21,14 @@ def test_experimenter_missing_fields(self):
             Person()
 
 
+class TestAssetPath(unittest.TestCase):
+    """Test AssetPath class"""
+
+    def test_asset_path(self):
+        """Test AssetPath with valid path"""
+        asset_path = AssetPath("/path/to/file.txt")
+        self.assertEqual(str(asset_path), "/path/to/file.txt")
```
tmchartrand, line 30: This is looking good - these work also and might be worth documenting
```suggestion
        asset_path = AssetPath(Path("/path/to/file.txt"))
        self.assertEqual(str(asset_path), "/path/to/file.txt")

        asset_path = AssetPath(PureWindowsPath(r"path\to\file.txt"))
        self.assertEqual(str(asset_path), "path/to/file.txt")
```

src/aind_data_schema/utils/validators.py
```diff
@@ -106,3 +108,26 @@ def recursive_get_all_names(obj: Any) -> List[str]:
...
+    """
+    Recursively checks if any attribute in an object is a pathlib.Path and logs if it doesn't exist.
+    """
+    if isinstance(obj, Enum):
+        return
+
+    if isinstance(obj, PurePosixPath):
+        obj = Path(obj)
+
+    if isinstance(obj, Path):
```
tmchartrand, line 123: This is what I thought we want:
```suggestion
    if isinstance(obj, AssetPath):
```

tmchartrand: that'll work fine for `directory / obj` but not for `obj.exists`, so we'll need to remove the directory=None option, which doesn't seem to be used anyway.

src/aind_data_schema/utils/validators.py
```diff
@@ -106,3 +108,26 @@ def recursive_get_all_names(obj: Any) -> List[str]:
...
+        full_path = directory / obj if directory else obj
+        if not full_path.exists():
+            logging.warning(f"AssetPath {full_path} does not exist")
```
tmchartrand, line 126: maybe add soft validation of the relative path aspect too, something like
```suggestion
            logging.warning(f"AssetPath {full_path} does not exist")
        if obj.is_absolute():
            try:
                relative_path = obj.relative_to(directory)
            except ValueError:
                logging.warning(f"AssetPath {obj} is absolute, should be relative")
            else:
                logging.warning(f"AssetPath {obj} is absolute, replace with {relative_path}")
    ```

Comment (tmchartrand) [reply src/aind_data_schema/utils/validators.py:126]: (don't think its possible to actually change the path object in-place)

**Review by tmchartrand — CHANGES_REQUESTED** [review]

File: src/aind_data_schema/utils/validators.py (hunk 4 above)
Comment (dbirman) [reply src/aind_data_schema/utils/validators.py:123]: AssetPath as defined has no exists() function, we have to convert to Path to use the local filesystem.

File: src/aind_data_schema/utils/validators.py (hunk 5 above)
Comment (dbirman) [reply src/aind_data_schema/utils/validators.py:126]: This should probably be a validator, I'll add it as one

File: src/aind_data_schema/utils/validators.py (hunk 4 above)
Comment (tmchartrand) [reply src/aind_data_schema/utils/validators.py:123]: true, what i meant was `directory / obj` should be a Path. explicitly converting is fine too, but it should all be nested under a top `isinstance(obj, AssetPath)`

File: src/aind_data_schema/utils/validators.py (hunk 5 above)
Comment (dbirman) [reply src/aind_data_schema/utils/validators.py:126]: Oh wait right these aren't pydantic models... I guess this is another recursive validator?

File: src/aind_data_schema/utils/validators.py (hunk 4 above)
Comment (dbirman) [reply src/aind_data_schema/utils/validators.py:123]: If we do type checks against AssetPath we have to move AssetPath into its own file (recursive imports), I'm inclined to keep it in identifiers. I think it's fine as-is, unless I'm missing something here you're not changing functionality just how the code is written?

File: src/aind_data_schema/utils/validators.py (hunk 5 above)
Comment (tmchartrand) [reply src/aind_data_schema/utils/validators.py:126]: we could do it as a validator i think by adding it to your __get_pydantic_core_schema__ call, just would lose the ability to check relative to the write directory and suggest a replacement. that's probably fine, although i suppose a separate recursive validator could actually make the replacement if structured differently. don't think i have an opinion.

File: src/aind_data_schema/utils/validators.py (hunk 4 above)
Comment (tmchartrand) [reply src/aind_data_schema/utils/validators.py:123]: It is different functionally if a user inserts a PurePosixPath somewhere random, but I'm not sure if that actually matters. can't you just add the import inline within the function here though?

Comment (dbirman) [reply src/aind_data_schema/utils/validators.py]: Okay now I understand. You couldn't add the import it was circular, but I refactored it so that this works.

Comment (dbirman) [reply src/aind_data_schema/utils/validators.py]: I see now. A schema developer would need to add a PurePosixPath typed field to trigger that issue, but there's no reason to risk it I refactored it so that the import works.

File: src/aind_data_schema/utils/validators.py (hunk 5 above)
Comment (dbirman) [reply src/aind_data_schema/utils/validators.py]: I'm just adding it to the recursive checker. Good enough

File: src/aind_data_schema/utils/validators.py [hunk 6]
```diff
@@ -106,3 +110,29 @@ def recursive_get_all_names(obj: Any) -> List[str]:
             names.extend(recursive_get_all_names(field_value))
 
     return names
+
+
+def recursive_check_paths(obj: Any, directory: Optional[Path] = None):
+    """
+    Recursively checks if any attribute in an object is a pathlib.Path and logs if it doesn't exist.
```
tmchartrand, line 117: update docstring (AssetPath, relative path check, and meaning of `directory` (with default meaning cwd).

tmchartrand (approved):
i think this is set, just update the docstring mentioned!

src/aind_data_schema/utils/validators.py
dbirman: done, thanks

PR #1344, opened 2025-03-26
feat: add optional coordinate system to acquisition, and improve validators

dbirman:
This PR adds an optional CoordinateSystem field to acquisition, for situations where the acquisition coordinate system differs from the instrument coordinate system. It also cleans up how the validators work:

- Added a check for the axis count in systems, so you can't provide a 3 axis coordinate system and then provide only 1 or 2 values in a coordinate
- Improved how nested `coordinate_system` fields are handled. Now if the validator starts with e.g. the instrument's coordinate system but runs into a different system on a particular object it will correctly validate that object's fields against the object's `coordinate_system` value instead of the instrument's.

### DIFF
--- a/src/aind_data_schema/utils/validators.py
+++ b/src/aind_data_schema/utils/validators.py
@@ -34,9 +34,9 @@ def recursive_coord_system_check(data, system_name: str, axis_count: int):
     if not data:
         return
 
-    if hasattr(data, "coordinate_system"):
+    if hasattr(data, "coordinate_system") and data.coordinate_system:
         # If we find a new coordinate_system, allow it to over-write our settings
-        system_name = data.coordinate_system
+        system_name = data.coordinate_system.name
         axis_count = len(data.coordinate_system.axes)
 
     # Check if the object we are looking at has a system_name field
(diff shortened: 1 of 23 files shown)


src/aind_data_schema/utils/validators.py
```diff
@@ -34,15 +34,28 @@ def recursive_coord_system_check(data, system_name: str):
     if not data:
         return
 
+    if hasattr(data, "coordinate_system") and data.coordinate_system:
+        # If we find a new coordinate_system, allow it to over-write our settings
```
saskiad, line 38: The part I'm confused by is how does it know what order to look at things in? 
Which I think is more of a problem of I'm not quite seeing how this works - what is the data that get's put into the validator? Is it just one class or is it the full metadata object?

saskiad (approved):

dbirman:
Changes from our discussion:

- Make coordinate_system required in both acquisition and procedures. Can be simply copied from instrument or can be re-defined, as needed.
- Look for places where we need optional coordinate_systems to over-write those, e.g. Surgery, SectioningProcedure, etc

PR #1345, opened 2025-03-26
Feat 1045 inputdata

### LINKED ISSUE #1045: allow `DataDescription.input_data` to be a list
Right now it accepts a string.  Would be nice to also allow a list of strings and many data assets can contribute to a new one.

Please do this in a backwards compatible way. e.g. `Union[str, List[str]]`

saskiad:
closes #1045


saskiad:
Changed mind on this - see issue

PR #1353, opened 2025-03-27
refactor: remove computer_name, add Computer device

dbirman:
See title


examples/ephys_instrument.py
```diff
@@ -37,8 +38,8 @@
 # Describes an instrument with running wheel, 2 behavior cameras, one Harp Behavior board,
 # one dual-color laser module, one stick microscope, and 2 Neuropixels probes
 
-behavior_computer = "W10DT72941"
-ephys_computer = "W10DT72942"
+behavior_computer = Computer(name="Behavior computer", serial_number="W10DT72941")
```
saskiad, line 41: hmm... these aren't serial numbers. These are what the institute considers to be the name of the computer. But it's much nicer to say behavior computer instead of that long thing. 
should we add a field to the computer device for this name, because we do want this somehow.

dbirman: Are you sure we want to track the institute name for these? I'm not sure I see it being important

tests/resources/ephys_instrument.py
```diff
@@ -78,6 +83,33 @@
...
+        connection_data={
+            "Harp Behavior": ConnectionData(direction=ConnectionDirection.SEND),
+            "Face Camera": ConnectionData(direction=ConnectionDirection.SEND),
+            "Body Camera": ConnectionData(direction=ConnectionDirection.SEND),
+            computer_names["Behavior computer"]: ConnectionData(direction=ConnectionDirection.RECEIVE),
+        },
+    ),
+    Connection(
+        device_names=[
+            computer_names["Behavior computer"],
```
saskiad, line 97: you probably want this as ephys computer. Same for line 105

saskiad (approved):

tests/resources/ephys_instrument.py
dbirman: Fixed

PR #1357, opened 2025-03-27
Processing changes

### LINKED ISSUE #1167: [2.0] clarify processing input/output
In many data processing tasks there are multiple distinct inputs, e.g. generating the smartspim template or training a model. We added minimal support for this in #1166 in the form of a list of input locations, but discussed a plan to add additional context explaining the selection of those inputs.

```python
class CompositeData(AindModel):
    """Description of a group of data assets used together"""
    data_assets: List[str] = Field(..., title="Data assets")
    name: Optional[str]
    shared_metadata: Optional[AindGenericType] = Field(
        default=None,
        title="Shared metadata", 
        description="Common attributes that provide context for this grouping of assets"
    )
    curation_purpose: Optional[str] = Field(
        default=None,
        title="Curation purpose",
        description="Reason for grouping assets together for processing"
    )
```

The most obvious alternative is to create a separate DataProcess stage for "curation" or the like, with no inputs and the list of curated assets as an output parameter. Other contextual fields could be entered as notes or parameters of the process.

**Additional context**
One advantage of a separate object for this is that it would be easier to write a script to run a new process on the same set of data assets as a previous processing result (could be a common use case for model evaluation etc).
Unclear if this would be reused/allowed anywhere other than the input field of DataProcess

This is also related to the discussion https://github.com/AllenNeuralDynamics/aind-data-schema/discussions/1148

tmchartrand:
Major PR here to hopefully resolve all the Processing schema discussion
(closes #1167 )
bits that may merit further review/discussion:

- I tried out an idea of getting rid of the extra nesting of a Provenance object by adding Code.input_data. This is by default an empty list, which is a possible but rare case for processing code (ie simulations) but common for other uses of Code in the schema.
- Instead of having DataProcess.code of type Union[Code, Software], I added an optional Code.name field to enable Code to be used in case of a GUI software process better identified by name (ie manual alignment of data using QuickNii). URL is still required.
- I changed DataProcess.name to DataProcess.type to reflect that this is filled from a controlled list of types of processing in the models repo, and added an additional DataProcess.name which by default simply duplicates the type, but can be set to any string to disambiguate e.g. multiple analysis steps.
- I added a field to directly represent the computation graph links between processes, DataProcess.input_processes. My suggestion is to allow this to be blank by default, interpreted as taking input from the preceding process in the Processing.data_processes list. Processing validates that these at least match existing names in Processing.data_processes (along with the DataProcess.pipeline_processes as Dan had set up).
- Added CombinedData with optional name and external_links, both of which would be filled only when a CO Combined Data asset exists.

Couple other minor resolutions of previous points:

- didn't rename Software (vs SoftwarePackage) or DataProcess.experimenters (considered 'owners')
- changed DataProcess.output_location to output_path, made optional, and used Path type for now


src/aind_data_schema/core/processing.py
```diff
@@ -55,35 +49,61 @@ class ResourceUsage(DataModel):
 class DataProcess(DataModel):
     """Description of a single processing step"""
 
-    name: ProcessName = Field(..., title="Name")
+    type: ProcessName = Field(..., title="Process type")
```
saskiad, line 52: field can't be named "type" - it could probably be process_type

src/aind_data_schema/core/processing.py
```diff
@@ -55,35 +49,61 @@ class ResourceUsage(DataModel):
...
+    name: str = Field(
+        default="",
+        title="Name",
+        description=("Unique name of the processing step.", " If not provided, the type will be used as the name."),
+    )
     stage: ProcessStage = Field(..., title="Processing stage")
-    experimenters: List[Person] = Field(..., title="experimenters", description="People responsible for processing")
-    code: Code = Field(..., title="Code used for processing")
-    pipeline_steps: Optional[List[ProcessName]] = Field(
+    code: Code = Field(..., title="Code", description="Code or script used for processing")
```
saskiad, line 59: minor: remove the "or script" in the description

src/aind_data_schema/core/processing.py
```diff
@@ -55,35 +49,61 @@ class ResourceUsage(DataModel):
...
+            "For pipeline processes (ProcessName.PIPELINE), include all steps providing output.",
+            "If not provided, the preceding DataProcess is assumed.",
+        ),
+    )
+    pipeline_steps: Optional[List[str]] = Field(
         default=None,
         title="Pipeline steps",
         description=(
-            "For pipeline processes (ProcessName.PIPELINE), steps should indicate the DataProcess objects",
+            "For pipeline processes (ProcessName.PIPELINE) only, list names of all DataProcess objects",
```
saskiad, line 74: add a validator that throw's an error if ProcessName is PIPELINE and pipeline_steps is empty

src/aind_data_schema/core/processing.py
tmchartrand: not arguing, but in the process of changing this I noticed Anaesthetic.type, which we may want to change also!

src/aind_data_schema/core/processing.py
```diff
@@ -55,35 +49,61 @@ class ResourceUsage(DataModel):
...
             " that are part of the pipeline. Each object must show up in the data_processes list.",
         ),
     )
     start_date_time: AwareDatetimeWithDefault = Field(..., title="Start date time")
     end_date_time: AwareDatetimeWithDefault = Field(..., title="End date time")
-    # allowing multiple input locations, to be replaced by CompositeData object in future
-    input_location: Union[str, List[str]] = Field(..., description="Path(s) to data inputs", title="Input location")
```
saskiad, line 73: Don't we still need something for input data? Now we only have an option for input processes

src/aind_data_schema/core/processing.py
```diff
@@ -55,35 +49,61 @@ class ResourceUsage(DataModel):
...
-    pipeline_steps: Optional[List[ProcessName]] = Field(
+    code: Code = Field(..., title="Code", description="Code or script used for processing")
+    experimenters: List[Person] = Field(..., title="Experimenters", description="People responsible for processing")
+    input_processes: List[str] = Field(
+        default=[],
+        title="Input processes",
+        description=(
+            "The names of the DataProcess objects that are inputs to this process.",
+            "For pipeline processes (ProcessName.PIPELINE), include all steps providing output.",
+            "If not provided, the preceding DataProcess is assumed.",
```
saskiad, line 67: We don't make assumptions. Remove this line. Rule number 1 of metadata: no cheat codes.

saskiad (commented):
Still some things to iron out

src/aind_data_schema/core/processing.py
tmchartrand: this is actually already in validate_pipeline_steps - I'll add a test case though.

src/aind_data_schema/core/processing.py
tmchartrand: that's moved into the Code.input_data now

src/aind_data_schema/core/processing.py
saskiad: ah, I must have missed it. Apologies.

src/aind_data_schema/components/identifiers.py
```diff
@@ -30,13 +60,22 @@ class Code(DataModel):
...
+    name: Optional[str] = Field(default=None, title="Name", description="Name of the package or script")
     version: Optional[str] = Field(default=None, title="Code version")
-
-    software: Optional[Software] = Field(default=None, title="Software", description="Software package")
+    run_script: Optional[Path] = Field(default=None, title="Run script", description="Path to run script")
 
     language: Optional[str] = Field(default=None, title="Programming language", description="Programming language used")
     language_version: Optional[str] = Field(default=None, title="Programming language version")
 
+    input_data: List[Union[DataAsset, CombinedData]] = Field(
```
saskiad, line 70: label this as optional since the default is [ ]

src/aind_data_schema/core/model.py
```diff
@@ -32,14 +32,14 @@ class PerformanceMetric(DataModel):
 class ModelEvaluation(DataProcess):
     """Description of model evaluation"""
 
-    name: ProcessName = Field(ProcessName.MODEL_EVALUATION, title="Process name")
+    type: ProcessName = Field(ProcessName.MODEL_EVALUATION, title="Process name")
```
saskiad, line 35: ditto

saskiad: why not use name?

src/aind_data_schema/core/processing.py
saskiad: I'm not sure I like this ... not sure I don't...

src/aind_data_schema/components/identifiers.py
tmchartrand: doesn't Optional[List] imply that it might be None, which isn't the case here?

src/aind_data_schema/core/processing.py
tmchartrand: I think we definitely don't want to require the user to specify this always, but I could implement a validator to fill the field with the "assumed" values if not given?

saskiad: we can make it optional (as it is here... and we should label it as such) but we can't assume what it is. If users can't provide this information, we can't guess at what it is.

saskiad: I'm having a hard time imagining a pipeline process step that doesn't know what the input process is

dbirman: So this is making DataProcess into a node in a directed graph, basically reflecting how capsules were run in CodeOcean?

I think if we want to do this we may want to just explicitly represent the graph. My vote would be to store these not in the DataProcess list itself but rather in a dictionary in Processing that stores the DataProcess->DataProcess mappings. So e.g. if you have "Compression" -> "SpikeSorting" -> "QC" the dictionary is:

```
process_map={
  "Compression": "SpikeSorting",
  "SpikeSorting": "QC",
}
```

Storing these within the objects will make it hard to recover the graph and also make the validator a pain

tmchartrand: I think I like this idea, although probably an OrderedDict[str, List[str]]. I'll refactor

tmchartrand: I think this should be resolved, following Dan's idea. I oriented the dict {process: [input processes]} since it seemed more natural.

src/aind_data_schema/components/identifiers.py
tmchartrand: Dan and I discussed, and both agree that a fully optional field with default None means a lack of information. I think we always want input_data information even if that info is that there are no inputs, so None shouldn't be an option here.

The remaining question is if "no inputs" as a default is too risky, but I think it's by far the most common case of Code usage, at least outside of code for a Processing step. If we're uncomfortable having the default in general, maybe the answer is a subclass ProcessingCode that adds the field, and when using that the user must explicitly specify if there are no inputs.

tmchartrand: Dan and I settled on Optional with default None here, meaning Code which is executed in a context without input data. Ideally Code within a DataProcess should explicitly provide an empty list if there is no input, but I don't think we need to validate that.

tmchartrand:
This should be fully good to go if @saskiad or @dyf want to take a look?

src/aind_data_schema/core/processing.py
```diff
@@ -95,41 +107,99 @@ class Processing(DataCoreModel):
     data_processes: List[DataProcess] = Field(..., title="Data processing")
     notes: Optional[str] = Field(default=None, title="Notes")
 
-    @model_validator(mode="before")
-    def validate_pipeline_steps(cls, values):
-        """Validator for pipeline_steps"""
+    process_graph: Dict[str, List[str]] = Field(
```
dbirman, line 110: This is the reverse of what I was thinking. Why are you deciding to do key: process, value: inputs? Why not just do 1:1 mappings of input:process as key:value pairs?

tests/test_composability_merge.py
```diff
@@ -450,13 +448,20 @@ def test_add_processing_objects(self):
...
         with self.assertRaises(ValueError) as e:
             _ = p1 + p3
         self.assertIn("Cannot add Processing objects with different schema versions.", str(e.exception))
 
+        # Test with duplicate processes
+        p_copy = p1.model_copy()
+        with self.assertWarns(Warning) as w:
+            combined = p_copy + p1
+        self.assertIn("Processing objects have repeated processes", str(w.warning))
+        self.assertEqual(combined.data_processes[1].name, "Analysis_1")
```
dbirman, line 463: We need test coverage in more places here:

- process_graph combinations (maybe assert # of key/value pairs is the original amount +1?)
- need to check that duplicate process names inrement properly

src/aind_data_schema/core/processing.py
```diff
@@ -55,35 +50,52 @@ class ResourceUsage(DataModel):
...
-        if info.data.get("name") == ProcessName.OTHER and not value:
-            raise ValueError("Notes cannot be empty if 'name' is Other. Describe the process name in the notes field.")
+        if info.data.get("process_type") == ProcessName.OTHER and not value:
+            raise ValueError(
+                "Notes cannot be empty if 'process_type' is Other. Describe the type of processing in the notes field."
+            )
         return value
 
+    @model_validator(mode="after")
+    def fill_default_name(self) -> "DataProcess":
```
dbirman, line 92: Doesn't this run the risk of creating duplicate process names in the same Processing object? Then if you were to combine that with another Processing object the names would get re-named unexpectedly.

dbirman (commented):
Questions about processing_graph and the duplicate name handling

src/aind_data_schema/core/processing.py
tmchartrand: Processing.validate_process_graph checks unique names, tested in test_unique_process_names

src/aind_data_schema/core/processing.py
tmchartrand: renaming to dependency_graph for clarity, but we agreed this is the only reasonably natural way to work with the structure.

tmchartrand:
@dbirman this should be ready to merge, models and all.

src/aind_data_schema/components/identifiers.py
```diff
@@ -30,13 +60,22 @@ class Code(DataModel):
...
+    name: Optional[str] = Field(default=None, title="Name")
     version: Optional[str] = Field(default=None, title="Code version")
-
-    software: Optional[Software] = Field(default=None, title="Software", description="Software package")
+    run_script: Optional[Path] = Field(default=None, title="Run script", description="Path to run script")
 
     language: Optional[str] = Field(default=None, title="Programming language", description="Programming language used")
     language_version: Optional[str] = Field(default=None, title="Programming language version")
 
+    input_data: Optional[List[Annotated[Union[DataAsset, CombinedData], Field(discriminator="object_type")]]] = Field(
```
dbirman, line 70: Why did we settle on putting this in Code vs in DataProcess? I just can't remember the discussion... putting it here feels wrong to me.

tmchartrand: mainly to have all provenance-related metadata within a single object without having an extra level of nesting (a Provenance object). I also think it's on par with parameters and they should either both be under code or neither.

dbirman: But most code doesn't have input_data right, and actually specifically code that has no input_data is likely to have parameters (like stimulus code).

I'd rather merge this today so I'm not too worried about it, I think a big point of 2.0 though is to get rid of Optional fields as much as possible because it takes bandwidth for users to look at them and interpret whether they apply to their situation. Since Code will only ever use input_data when it's nested in DataProcess the field should be moved there. Anyway we can drop this point and revisit in 3.0 if it turns out I'm right that Code outside of DataProcess doesn't use this field.

Issue #1180, opened 2024-12-06
[2.0] Re-organize acquisition around channels 


dbirman:
**Is your feature request related to a problem? Please describe.**
We want to re-organize acquisition to be organized by imaging channels instead of tiles

**Describe the solution you'd like**

- [ ] Channel is the main feature of an acquisition
- [ ] Tile deprecated
- [ ] Upgrader from v1.X tiles to v2.0 channels

**Describe alternatives you've considered**
@saskiad additional details probably needed before someone tackles this

**Additional context**
Note, depends on https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1173

https://alleninstitute.sharepoint.com/:w:/r/sites/NeuralDynamics/_layouts/15/Doc.aspx?sourcedoc=%7BE27B889D-27A1-4CC0-948C-10ADBE605654%7D&file=Data+Schema+2.0.docx&action=default&mobileredirect=true&DefaultItemOpen=1

saskiad:
Channel:
- emission information needs to be more specific. current version is based on emission filters being in a wheel, so the wheel index tell us the emission wavelength, etc. I think that should be contained here
- Channel is used for in vitro imaging (acquisition) but also in vivo (ophys) imaging, so need to make sure it works for both of those
- Might need to be a base Channel class and then a few child classes. (e.g. the dilation only applies to slap, so it could be a slap channel thing). 

Tile/Acquisition Tile:
- Don't deprecate these, but remove the channel field from the acquisition tile

In acquisition:
1. add a `channels` field that is a required list. This uses a Channel object that has the base channel info (above) plus a lit of acquisition tiles.
2. Remove the `tiles` from acquisition class

dbirman:
Pre-emptive closing, this is in PR #1395

PR #1360, opened 2025-03-28
attempt at channel reconfigure

### LINKED ISSUE #1290: Channel class updates
Channel class needs to be updated to accomodate use in the Z1:

the Z1 has channels with more than one light source, and then the emission has 2 filter wheel positions and a dichroic filter. 

This class accommodates a lot of different uses, so I want to both be careful we don't break things for others, but also not make it too kludgy. We might need a base Channel with some SlapChannel or Z1Channel inheritence?

saskiad:
closes #1290


saskiad:
note: I think that the updates to channel in 1.4 are missing from here. We should figure that out.

dbirman:
> note: I think that the updates to channel in 1.4 are missing from here. We should figure that out.

1.4 isn't merged into release right now, the merge conflict is pretty bad so I've been putting it off

src/aind_data_schema/components/tile.py
```diff
@@ -13,21 +13,21 @@ class Channel(DataModel):
     """Description of a channel"""
 
     channel_name: str = Field(..., title="Channel")
-    light_source_name: str = Field(..., title="Light source name", description="Must match device name")
-    filter_names: List[str] = Field(..., title="Filter names", description="Must match device names")
     detector_name: str = Field(..., title="Detector name", description="Must match device name")
-    additional_device_names: List[str] = Field(default=[], title="Additional device names")
+    additional_device_names: Optional[List[str]] = Field(default=[], title="Additional device names")
```
dbirman, line 17: flake8 will force you to change this to default=None to pass the linter checks

dbirman (commented):
This all makes sense to me, I'm a little unhappy though with having light_sources/filters which are presumably pointing to devices in instrument without any kind of validation check though.

A different way to do this would be to have a validator that requires you to list light source and filter in `active_devices` when you're doing slap and then link the configs together inside channel?

saskiad:
I think we should think about #1180 at the same time because it might help with this a little bit.

src/aind_data_schema/components/tile.py
saskiad: fixed

dbirman:
> note: I think that the updates to channel in 1.4 are missing from here. We should figure that out.

 I opened a PR with the v1.4 changes, best to merge it before making any more changes here (and then merge release into this branch): https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1386

src/aind_data_schema/components/configs.py
```diff
@@ -85,19 +85,58 @@ class DetectorConfig(DeviceConfig):
...
+    light_source_configurations: List[
+        Annotated[
+            Union[
+                LaserConfig,
+                LightEmittingDiodeConfig,
+            ],
+            Field(discriminator="object_type"),
+        ]
+    ] = Field(default=[], title="Light source configurations")
+    excitation_filters: Optional[List[str]] = Field(default=None, title="Excitation filters")
```
dbirman, line 122: I know it seems like overkill but I think we should create a `FilterConfig` class and use it here and on the next line. It can just be an empty class definition `class FilterConfig(DeviceConfig)`. The advantage of doing this is it looks like the pattern everywhere else and it triggers the `device_name` validator which will make sure people give us the names of real devices.

src/aind_data_schema/components/configs.py
```diff
@@ -85,19 +85,58 @@ class DetectorConfig(DeviceConfig):
...
+    # emission
+    emission_filters: Optional[List[str]] = Field(default=None, title="Emission filter names")
+
+
+class SlapChannel(Channel):
+    """Description of a channel for Slap"""
+
+    dilation: Optional[int] = Field(default=None, title="Dilation (pixels)")
+    dilation_unit: Optional[SizeUnit] = Field(default=None, title="Dilation unit")
+    description: Optional[str] = Field(default=None, title="Description")
```
dbirman, line 132: Can you add a description for this field to say what is supposed to go here

src/aind_data_schema/components/configs.py
```diff
@@ -85,19 +85,58 @@ class DetectorConfig(DeviceConfig):
...
 
 
+class Channel(DataModel):
+    """Description of a channel"""
+
+    channel_name: str = Field(..., title="Channel")
+    intended_measurement: Optional[str] = Field(default=None, title="Intended measurement",
+                                                description="What signal is this channel measuring")
+    detector_configuration: DetectorConfig = Field(..., title="Detector configuration")
+    additional_device_names: Optional[List[str]] = Field(default=None, title="Additional device names")
```
dbirman, line 111: What is supposed to go in the additional_device_names field?

src/aind_data_schema/components/configs.py
```diff
@@ -410,10 +438,28 @@ class Immersion(DataModel):
     refractive_index: Decimal = Field(..., title="Index of refraction")
 
 
+class Tile(DataModel):
```
dbirman, line 441: Can we rename Tile to Image?

src/aind_data_schema/components/configs.py
```diff
@@ -410,10 +438,28 @@ class Immersion(DataModel):
...
+    """Description of an image tile"""
+
+    coordinate_transform: CoordinateTransform = Field(..., title="Tile coordinate transformations")
+    file_name: Optional[str] = Field(default=None, title="File name")
+    imaging_angle: int = Field(default=0, title="Imaging angle")
```
dbirman, line 446: Can you add the description I had in the other PR: `description="Angle of the detector relative to the image plane, 0 indicates perpendicular"`

src/aind_data_schema/components/configs.py
```diff
@@ -410,10 +438,28 @@ class Immersion(DataModel):
...
+    imaging_angle_unit: AngleUnit = Field(default=AngleUnit.DEG, title="Imaging angle unit")
+    tile_start_time: Optional[AwareDatetimeWithDefault] = Field(default=None, title="Tile acquisition start time")
```
dbirman, line 448: Rename to imaging_start_time and imaging_end_time for consistency with angle

src/aind_data_schema/components/configs.py
```diff
@@ -410,10 +438,28 @@ class Immersion(DataModel):
...
+    tile_end_time: Optional[AwareDatetimeWithDefault] = Field(default=None, title="Tile acquisition end time")
+    notes: Optional[str] = Field(default=None, title="Notes")
```
dbirman, line 450: I think this should be removed. Tile/Images are auto-generated by code right? Is there even a mechanism by which someone could add notes to a tile?

src/aind_data_schema/components/configs.py
```diff
@@ -410,10 +438,28 @@ class Immersion(DataModel):
...
+
+
+class SPIMChannel(Channel):
+    """Description of a light sheet channel"""
+
+    tiles: List[Tile] = Field(..., title="Tiles")
```
dbirman, line 456: Anywhere we have a List[Tile] we need to define the CoordinateSystem next to it, so the coordinate_system field from InVitroImagingConfig should get moved here

dbirman (commented):
This captures most of what I had in the other PR, but I think we should lean into defining `*Config(DeviceConfig)` classes rather than explicitly list the names of devices.

src/aind_data_schema/components/configs.py
dbirman: List[DeviceConfig]

src/aind_data_schema/components/configs.py
dbirman: Need to check if these are used (other than specifying power unit)

src/aind_data_schema/components/configs.py
dyf: Yes

src/aind_data_schema/components/configs.py
```diff
@@ -62,6 +61,7 @@ class PatchCordConfig(DeviceConfig):
     output_power: Decimal = Field(..., title="Output power (uW)")
     output_power_unit: PowerUnit = Field(default=PowerUnit.UW, title="Output power unit")
     fiber_name: str = Field(..., title="Fiber name (must match procedure)")
+    channel: Channel = Field(..., title="Channel")
```
dbirman, line 64: Request from Doug to support fiber with temporal multiplexing: make this a List[Channel], otherwise we need 3 PatchCordConfigs per fiber. Either option is ultimately fine but the list might make more sense stream?

src/aind_data_schema/components/configs.py
```diff
@@ -85,19 +85,58 @@ class DetectorConfig(DeviceConfig):
...
+            Union[
+                LaserConfig,
+                LightEmittingDiodeConfig,
+            ],
+            Field(discriminator="object_type"),
+        ]
+    ] = Field(default=[], title="Light source configurations")
+    excitation_filters: Optional[List[str]] = Field(default=None, title="Excitation filters")
+    # emission
+    emission_filters: Optional[List[str]] = Field(default=None, title="Emission filter names")
```
dbirman, line 124: If someone doesn't have a filter in the emission path wouldn't we not know what their emission wavelength is?

src/aind_data_schema/components/configs.py
saskiad: what if we just use DeviceConfig rather than making an empty FilterConfig class?

src/aind_data_schema/components/configs.py
saskiad: let me chat with Doug - I don't think this is how this works.

src/aind_data_schema/components/configs.py
saskiad: I think this is resolved by listing images in the InVitroImagingConfig directly.

src/aind_data_schema/components/configs.py
saskiad: nope.

src/aind_data_schema/components/configs.py
saskiad: not sure. E.g. the emission of the flourophore is a feature of the fluorophore. I don't know that the emission pathway always has a filter to restrict it. But we won't have a field in the metadata that defines it. So the question is do we need a field for it?

src/aind_data_schema/components/configs.py
dbirman: I'm pretty sure that's fine? Trying to think of whether the object_type matters for some reason but I don't think it does.

src/aind_data_schema/components/configs.py
saskiad: figured this out.

dbirman:
I blew up this PR by accidentally merging a branch into it that my mediocre knowledge of git was insufficient to properly revert. I made a new branch off of the last changes you had before I messed things up. Please use this branch/PR instead: https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1395

Issue #1330, opened 2025-03-24
[2.0] Go through devices and remove unecessary fields


saskiad:
eg. detectors and filters have some fields that I think aren't needed and are better documented by the model number. There also might be some fields in daq devices or other things that are no longer needed (e.g. disc output?)

saskiad:
some thoughts:
`Device`
- remove port_index? What is this for?
- does path_to_cad make sense? I don’t see it getting used.

`Filter`
Remove sizes (diameter, thickness, etc)
We don’t want to rely on filter_wheel_index
Keep wavelengths
Remove description (users can use notes if they need to add info)

`Lens`
Do we need any of these fields?

`DAQChannel` - we’re removing this right?

`DAQDevice` 
- remove computer name because we’re making a computer device (correct?)
- remove channels?

`HarpDevice` feels bulkier than we want. 

`Laser`
- wavelength should be optional for variable wavelength lasers.
- remove maximum power
- is coupling used?
- What is `item_number`?

`PatchCord` and `FiberProbe` 
I feel like these warrant asking Kenta if these fields still make sense. He defined them, but we should make sure it’s not just adding part number attributes.

`Enclosure` internal material can be optional probably

`MousePlatform` 
- I don’t know that we need this. Maybe surface material can be added to Disc, but I don’t know how important it is. It’s something we add to the manufactured part so we probably want it documented. 
Remove the date_surface_replaced. Nobody’s using it and it should be a maintenance object if so.
- we should disambiguate disc and wheel for people. Disc mouse runs on the flat side, Wheel mouse runs on the edge.

`OpticalTable` 
Karel has commented on this being ridiculous. I don’t have strong feelings - it’s NEVER the thing that trips users up, but it is a pretty unimportant item.

dbirman:
I opened a branch and removed a first pass, but there's some follow-up that needs to be done.

PR #1365, opened 2025-03-28
1356 20 more last minute refactors

dbirman:
PR closes the following small changes:

1.  Not sure why compression is only on Camera, not Detector. Also compression is probably an acquisition thing, not an Instrument thing. And it should be "during acquisition" not "after acquisition".
2. MotorizedStage.firmware should be a Software
3. HarpDevice.manufacturer is always a DAQ manufacturer?
4. Why is Manipulator the only Device with a CoordinateSystem? Shouldn't all Devices have an optional CS?
5. rename PatchCord -> FiberPatchCord?
6. Headstage -> it's a synonym for device - remove and use Device in EphysProbe.
7. rename MousePlatform -> Platform? Or maybe SubjectPlatform? I don't think RatPlatform will be a thing.
8. Monitor.position -> should rather be a Coordinate?
9. CameraAssembly.position should rather be a Coordinate?
10. Speaker.position should rather be a Coordinate?
11. rename ChannelType ->OlfactometerChannelType


dyf (approved):
probably should have @saskiad take a look

src/aind_data_schema/components/devices.py
```diff
@@ -662,14 +651,14 @@ class Enclosure(Device):
     air_filtration: bool = Field(..., title="Air filtration")
 
 
-class MousePlatform(Device):
-    """Description of a mouse platform"""
+class SubjectPlatform(Device):
```
saskiad, line 654: I think MousePlatform should be removed completely. See #1330

saskiad (commented):
Only thing I think we should change is to remove the MousePlatform (or SubjectPlatform) class altogether.

PR #1368, opened 2025-03-28
model schema changes - draft PR for review

tmchartrand:


tmchartrand:
This should be finalized now. It consists of:

- simple changes to remove Model fields redundant with DataDescription
- a few changes to Metadata regarding core/required models to allow a Model/DataDescription pair
- making DataDescription.license refer to a License enum in the models repo
- adding a required Model.version!

and some minor Model refactors that are independent of the above:

- renaming ModelBackbone to ModelArchitecture, and moving the nested fields up to Model (except for redundant notes, and layers which should just go in parameters). ModelArchitecture.software is renamed software_framework for clarity.
- treating pretraining as a PretrainingProcess that goes in the Training field instead of Model.pretraining_url (so training can be required)

tmchartrand:
I'm basing this on the processing changes branch for clarity, and it can't be tested without merging https://github.com/AllenNeuralDynamics/aind-data-schema-models/pull/171 anyway.

src/aind_data_schema/core/metadata.py
```diff
@@ -217,9 +227,15 @@ def validate_metadata(self):
     def validate_expected_files_by_modality(self):
         """Validator warns users if required files are missing"""
 
-        for file in REQUIRED_FILES:
-            if not getattr(self, file):
-                warnings.warn(f"Metadata missing required file: {file}")
+        # if subject is not present, this better be a model
```
saskiad, line 230: it could also be simulated data

saskiad: Do we want a specific data level for model or do we consider it derived?

src/aind_data_schema/core/model.py
```diff
@@ -59,15 +56,17 @@ class Model(DataCoreModel):
...
 
     name: str = Field(..., title="Name")
-    license: str = Field(..., title="License")
-    code: Optional[Code] = Field(default=None, title="Model code")
-    developers: Optional[List[Person]] = Field(default=None, title="Name of developer(s)")
-    developer_institution: Optional[Organization.ONE_OF] = Field(default=None, title="Institute where developed")
-    modalities: List[Modality.ONE_OF] = Field(..., title="Modalities")
-    architecture: ModelArchitecture = Field(..., title="Model architecture")
+    version: str = Field(..., title="Version")
+    # this is optional for now, ultimately it should be required
```
saskiad, line 60: this is not optional. Do you want it to be optional?

src/aind_data_schema/core/model.py
```diff
@@ -59,15 +56,17 @@ class Model(DataCoreModel):
...
+    # with an option to specify a simple command or script instead of full Code
+    code: Optional[Code] = Field(
+        default=None, title="Code", description="Code to run the model, possibly including reference to sample data"
```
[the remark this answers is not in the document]
saskiad: So my understanding is we've removed ModelArchitecture to use code instead. The model architecture was required, do we not want this to be? I guess everything inside architecture was optional, so I guess it doesn't need to be.

src/aind_data_schema/core/model.py
```diff
@@ -59,15 +56,17 @@ class Model(DataCoreModel):
...
+    )
+    architecture: ModelArchitecture = Field(..., title="architecture", description="Model architecture / type of model")
+    software_framework: Optional[Software] = Field(default=None, title="Software framework")
+    parameters: GenericModelType = Field(default=GenericModel(), title="Parameters")
     intended_use: str = Field(..., title="Intended model use", description="Semantic description of intended use")
     limitations: Optional[str] = Field(default=None, title="Model limitations")
-    pretrained_source_url: Optional[str] = Field(default=None, title="Pretrained source URL")
-    training: Optional[List[ModelTraining]] = Field(default=[], title="Training")
-    evaluations: Optional[List[ModelEvaluation]] = Field(default=[], title="Evaluations")
+    training: List[ModelTraining | ModelPretraining] = Field(..., title="Training", min_length=1)
```
saskiad, line 70: is this synatx the same as an Annotated Union?

src/aind_data_schema/core/model.py
```diff
@@ -59,15 +56,17 @@ class Model(DataCoreModel):
...
-    modalities: List[Modality.ONE_OF] = Field(..., title="Modalities")
-    architecture: ModelArchitecture = Field(..., title="Model architecture")
+    version: str = Field(..., title="Version")
+    # this is optional for now, ultimately it should be required
+    # with an option to specify a simple command or script instead of full Code
+    code: Optional[Code] = Field(
+        default=None, title="Code", description="Code to run the model, possibly including reference to sample data"
+    )
+    architecture: ModelArchitecture = Field(..., title="architecture", description="Model architecture / type of model")
+    software_framework: Optional[Software] = Field(default=None, title="Software framework")
```
saskiad, line 66: what is software_framework? How is this different from the Code field?

src/aind_data_schema/core/model.py
```diff
@@ -59,15 +56,17 @@ class Model(DataCoreModel):
...
+    parameters: GenericModelType = Field(default=GenericModel(), title="Parameters")
```
saskiad, line 67: why isn't this also part of the code field?

saskiad (changes requested):
There are a couple redundancies that we should eliminate

src/aind_data_schema/core/model.py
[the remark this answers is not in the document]
tmchartrand: Sample data would be distinct from actual training or evaluation, just a toy example, maybe small enough to include in the code repo even. Essentially to confirm that someone's installation and run command are working.

What my comment in the code was getting at is that ultimately we want some version of "code to run the model" required, but for many cases like mlflow or huggingface models, the full Code object is a bit overkill (ways of saving models that include some embedded code for rerunning the model in just a few lines, so no need to install a full library). Now that I think through it though the only extraneous requirement is adding a url which could be useful as a link to the enabling package (ie mlflow).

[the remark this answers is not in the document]
tmchartrand: I think clarifying the name of this as "run_code" might help, and making it required is probably the right thing.

src/aind_data_schema/core/model.py
tmchartrand: this was referring to Model.code, resolving

src/aind_data_schema/core/metadata.py
tmchartrand: Ah yes forgot about simulated or processed/aggregated (template brain) cases - so I guess we should allow either model/data_description, processing/data_description, or the full required set?

I think what we settled on with David was that just keeping raw/derived for now seemed fine - don't think model needs it's own unless we also added ones for simulation and aggregated processing for instance.

tmchartrand: i think this is resolved in the latest commit.

src/aind_data_schema/core/model.py
tmchartrand: there could be a model developed in one framework (eg Model.software_framework = Pytorch) but saved in a format that's runnable in another framework (which would be given in Model.code/Model.run_code if we rename).

src/aind_data_schema/core/model.py
tmchartrand: parameters in run_code should be only ones that are specified at runtime, while this should probably be renamed architecture_parameters (or maybe properties?) - intended to capture good-to-know properties of the model like number of layers, input shape, etc.

src/aind_data_schema/core/model.py
tmchartrand: It's the same as Union[ModelTraining, ModelPretraining], no annotation here.

src/aind_data_schema/core/model.py
```diff
@@ -51,6 +39,15 @@ class ModelTraining(DataProcess):
     )
 
 
+class ModelPretraining(DataModel):
+    """Description of model pretraining"""
+
+    performance: List[PerformanceMetric] = Field(..., title="Pretraining performance")
```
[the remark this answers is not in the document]
tmchartrand: i think i meant the source_url required and this optional. but this would be something like reporting benchmarks, maybe don't need it at all.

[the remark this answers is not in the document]
dbirman: I vote for keeping public benchmark data out of the metadata, so maybe not needed here

src/aind_data_schema/core/data_description.py
```diff
@@ -12,6 +12,7 @@
     datetime_from_name_string,
     datetime_to_name_string,
 )
+from aind_data_schema_models.licenses import License
```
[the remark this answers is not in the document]
tmchartrand: done

src/aind_data_schema/core/model.py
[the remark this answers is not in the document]
tmchartrand: done

src/aind_data_schema/core/model.py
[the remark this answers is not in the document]
tmchartrand: changed to 
```
    example_run_code: Code = Field(
        title="Example run code", description="Code to run the model, possibly including example parameters/data"
    )
```

src/aind_data_schema/core/model.py
tmchartrand: added discriminator

dbirman:
Flagging that if you're keeping the `test_data` field it should become a Union[AssetPath, str]

```
    test_data: Optional[str] = Field(
        default=None, title="Test data", description="Path or cross-validation/split approach"
    )
```

src/aind_data_schema/core/model.py
tmchartrand: renamed this to Model.properties and clarified the description. Dan votes for removing it entirely but I feel like this was requested by folks in the design discussions so I'll let @saskiad make a final call.

src/aind_data_schema/core/model.py
```diff
@@ -32,25 +20,31 @@ class PerformanceMetric(DataModel):
...
+    process_type: ProcessName = ProcessName.MODEL_TRAINING
     train_performance: List[PerformanceMetric] = Field(
         ..., title="Training performance", description="Performance on training set"
     )
     test_performance: Optional[List[PerformanceMetric]] = Field(
         default=None, title="Test performance", description="Performance on test data, evaluated during training"
     )
-    test_data: Optional[str] = Field(
-        default=None, title="Test data", description="Path or cross-validation/split approach"
+    test_eval_method: Optional[str] = Field(
```
saskiad, line 37: full words for name. test_evaluation_method

src/aind_data_schema/core/model.py
saskiad: Let's go with `architecture_parameters`

saskiad (commented):
Two small things, otherwise good

src/aind_data_schema/core/model.py
tmchartrand: done

PR #1371, opened 2025-03-28
feat: add Affine to compose() function options

dbirman:


dyf (changes requested):
I think make all the Transforms have the same API and have Affine.to_matrix() be a noop

dyf (approved):

PR #1372, opened 2025-03-29
refactor: remove id, created, last_modified

dbirman:
This PR removes the fields:

- Metadata.id
- Metadata.created
- Metadata.last_modified

These will be used internally on DocDB but we'll no longer use them externally to identify assets. The API will continue to return at least the created/last_modified so downstream tools can use them. 

If we want to keep using `id` in downstream tools (like the QC portal) we would need to keep that in the metadata. I think we're safe to go to using `name`, which is better for users anyways, and we should deal with name conflicts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ readme = "README.md"
 dynamic = ["version"]
 
 dependencies = [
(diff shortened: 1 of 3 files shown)


helen-m-lin (approved):
Changes look good to me. Thanks!

PR #1375, opened 2025-03-29
refactor: move ethics_review_id into DataDescription

dbirman:
See title. Assuming we actually want this it's a small change.


saskiad:
I've asked my iacuc people about this, will check in when I hear back.

saskiad:
nope, it doesn't break down this nicely. We should keep ethics_review_id in the separate places.

PR #1376, opened 2025-03-30
refactor: re-organize excitation channels

dbirman:
[Draft]

This PR re-organizes the classes `Tile`, `AcquisitionTile`, and `Channel` to separate the concept of the excitation/emission channel from the images that are acquired. 

On the devices side, there is now an assembly for imaging channels

```
class ImagingChannelAssembly(Assembly):
    """Named assembly of a light source, filters, and detector for generating a specific excitation wavelength"""

    light: Annotated[Union[Laser, LightEmittingDiode, Lamp], Field(discriminator="object_type")]
    excitation_filters: List[Filter] = Field(..., title="Excitation filters")
    detector: Detector = Field(..., title="Detector")
```

This allows users to specify the excitation path as part of their instrument.

Then two new configs help users define the properties of the path during acquisition. First, we define the relevant wavelengths:

```
class ImagingChannelConfig(DeviceConfig):
    """Configuration of an imaging channel assembly"""

    # excitation path settings
    laser_config: LaserConfig = Field(..., title="Laser configuration")

    # emission info
    emission_wavelength: int = Field(..., title="Wavelength", ge=300, le=1000)
    emission_wavelength_unit: SizeUnit = Field(default=SizeUnit.NM, title="Emission wavelength unit")
```

Then separately we define the imaging properties:

```
class Image(DataModel):
    """Description of a 2D image"""

    coordinate_transform: CoordinateTransform = Field(..., title="Tile coordinate transformations", description="Must go from the image system to the acquisition system")
    file_path: Optional[AssetPath] = Field(default=None, title="File path")
    imaging_angle: int = Field(default=0, title="Imaging angle", description="Angle of the detector relative, 0 indicates perpendicular to the sample")
    imaging_angle_unit: AngleUnit = Field(default=AngleUnit.DEG, title="Imaging angle unit")
    imaging_start_time: Optional[AwareDatetimeWithDefault] = Field(default=None, title="Acquisition start time")
    imaging_end_time: Optional[AwareDatetimeWithDefault] = Field(default=None, title="Acquisition end time")


class ImagingConfig(DeviceConfig):
    """Configuration of images acquired on an imaging channel assembly"""

    images: List[Image] = Field(..., title="Acquisition tiles")
    image_coordinate_system: CoordinateSystem = Field(..., title="Image coordinate system")
    chamber_immersion: Immersion = Field(..., title="Acquisition chamber immersion data")
    sample_immersion: Optional[Immersion] = Field(default=None, title="Acquisition sample immersion data")

    # sort out how to deal with this
    # dilation
    dilation: Optional[int] = Field(default=None, title="Dilation (pixels)")
    dilation_unit: Optional[SizeUnit] = Field(default=None, title="Dilation unit")
```

I think this captures the intent of the original tiles but with better separation of purposes. Some notes:

- This is my first time trying nested configs. I think we need this: when an assembly has devices that have their own configurations we should re-use those device configurations, but often assemblies also have their own information that needs to be tracked separately. See the ephys PR for another example of this and why I think it's important to get right.
- I don't know how to deal with dilation. This is really a property of the sample so I feel like the right place to track it is in procedures, not in the acquisition, but for convenience I can appreciate why we're putting it here. I guess the right thing to do is subclass ImagingConfig to have SPIMImagingConfig?


src/aind_data_schema/components/devices.py
```diff
@@ -59,6 +59,12 @@ class Device(DataModel):
     notes: Optional[str] = Field(default=None, title="Notes")
 
 
+class Assembly(DataModel):
```
saskiad, line 62: We have used assembly to uniquely define devices that are physically co-located. Eg. a camera with an attached lens. I think it's important to maintain that concept, rather than devices that are functionally related.

src/aind_data_schema/components/devices.py
```diff
@@ -294,18 +299,26 @@ class Lamp(Device):
...
     # required fields
-    name: str = Field(..., title="Light assembly name")
     light: Annotated[Union[Laser, LightEmittingDiode, Lamp], Field(discriminator="object_type")]
     lens: Lens = Field(..., title="Lens")
 
     # optional fields
     filter: Optional[Filter] = Field(default=None, title="Filter")
 
 
+class ImagingChannelAssembly(Assembly):
```
saskiad, line 313: I don't quite get this - this isn't the full channel - this is the excitation path? But detector isn't part of the excitation path.
And see above for assembly concept. I don't know that breaking a channel apart into excitation and emission is the right approach.

saskiad, line 313: I don't think we want to define a channel at the device level. It's often an operational definition - different lasers can be combined with different emission filters to create different channels. We want the instrument to have the list of lasers and filters and detectors, and then operationally we say which combination we are using.

PR #1377, opened 2025-03-30
refactor: separating manipulator and ephys configs 

dbirman:
This draft PR refactors the DomeModule and ManipulatorModule classes to better reflect how they link to devices.

On the device side, I've separated `EphysProbe` and `FiberProbe` from the assembly versions (which include a manipulator). Assemblies can be used on rigs where the manipulators and probes are attached, while the separate probe classes can be used for chronic recordings.

To match this, there are now separate `ManipulatorConfig` and `ProbeConfig` classes, as well as corresponding `EphysAssemblyConfig` and `FiberAssemblyConfig` classes. The manipulator configuration stores information about the manipulator local axis positions, independent of the acquisition coordinate system. The probe config class stores information about the target coordinate in both the acquisition coordinate system and optionally in an atlas. `ProbeConfig` is now generic enough to be re-used by all kinds of devices that get stuck into brains.

**Up for debate**: how should we handle the MIS system? If we want to keep things simple we simply force users to report the position of each manipulator according to the acquisition coordinate system. Alternatively, and what I've done in this draft, is let them define the arc and module angles explicitly and then also report the actual angles in the ManipulatorConfig objects. 

Notes:
- Previously there were lasers attached to ephys probes, presumably for NeuropixelsOpto, users should now report those as a `Connection`

### DIFF
--- a/examples/ephys_acquisition.py
+++ b/examples/ephys_acquisition.py
@@ -12,16 +12,46 @@ from aind_data_schema.core.acquisition import (
     SubjectDetails,
 )
 from aind_data_schema.components.configs import (
-    DomeModule,
+    EphysAssemblyConfig,
     ManipulatorConfig,
+    ProbeConfig,
     StimulusModality,
+    AtlasCoordinate,
 )
-from aind_data_schema.components.coordinates import Coordinate, CoordinateSystemLibrary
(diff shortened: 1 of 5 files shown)


dougollerenshaw (approved):
This looks good to me. But I'm not capable of thinking through all of the use cases here.

Issue #960, opened 2024-06-03
Enum-like classes cant be used as dictionary keys


bruno-f-cruz:
Not sure if there is a different intended use...

To reproduce:

```python
from aind_data_schema_models.platforms import Platform
from pydantic import BaseModel


class Foo(BaseModel):
    platform: dict[Platform.ONE_OF, str]


foo = Foo(platform={Platform.BEHAVIOR: "foo"})
print(foo)

bar = foo.model_dump()

```

bruno-f-cruz:
Tracking https://github.com/pydantic/pydantic/issues/9453

bruno-f-cruz:
This is actually more serious that I initially gave it credit for. This pattern completely prevents users from using the `model_dump` method:

```python
import examples
from examples.fip_ophys_rig import rig
from aind_data_schema_models.modalities import Modality
rig.modalities = set([Modality.BEHAVIOR])
rig.model_dump()
```

raises the following error:

```shell
raceback (most recent call last):
  File ".\aind-data-schema\deserialization_order.py", line 27, in <module>     
    rig.model_dump()
  File ".\aind-data-schema\.venv\Lib\site-packages\pydantic\main.py", line 314, in model_dump
    return self.__pydantic_serializer__.to_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unhashable type: 'dict'
```

jtyoung84:
It seems like an issue with the pydantic serializer. Even this causes issues:

```python
from typing import Set
from pydantic import BaseModel, ConfigDict
from dataclasses import dataclass

@dataclass(frozen=True)
class Cell:
    value: str

# This works
set([Cell(value="a"), Cell(value="b")])

class WeightedSeeds(BaseModel):
    seeds: Set[Cell]

foo = WeightedSeeds(seeds=set([Cell(value="a"), Cell(value="b")]))

# This raises unhashable type errors
foo.model_dump()
```

bruno-f-cruz:
Possible solutions

1- Simplify everything and go back to native `Enum`. This is the preferred solution as it is rather isomorphic across programming languages and json.
2- write our own de(serializers) for a base class to be inherited from all complex Enum types. This strategy has a few problems:
[...]

PR #1385, opened 2025-04-02
1384 20 replace modalities set with list fix order validator

dbirman:
This PR removes the use of `Set` on ordered lists. Not sure where that came from, but it also causes serialization to fail when you use `model_dump()`.

### DIFF
--- a/src/aind_data_schema/base.py
+++ b/src/aind_data_schema/base.py
@@ -104,7 +104,7 @@ class DataModel(BaseModel, Generic[GenericModelType]):
     model_config = ConfigDict(extra="forbid", use_enum_values=True)
     object_type: ClassVar[str]  # This prevents Pydantic from treating it as a normal field
 
-    def __init_subclass__(cls, **kwargs): 
+    def __init_subclass__(cls, **kwargs):
         """Automatically set the correct `object_type` as a Literal[...]"""
         super().__init_subclass__(**kwargs)
         object_type_value = cls._object_type_from_name()
--- a/src/aind_data_schema/base.py
+++ b/src/aind_data_schema/base.py
@@ -104,7 +104,7 @@ class DataModel(BaseModel, Generic[GenericModelType]):
     model_config = ConfigDict(extra="forbid", use_enum_values=True)
     object_type: ClassVar[str]  # This prevents Pydantic from treating it as a normal field
 
-    def __init_subclass__(cls, **kwargs):
+    def __init_subclass__(cls, **kwargs): 
         """Automatically set the correct `object_type` as a Literal[...]"""
         super().__init_subclass__(**kwargs)
         object_type_value = cls._object_type_from_name()
--- a/src/aind_data_schema/core/instrument.py
+++ b/src/aind_data_schema/core/instrument.py
@@ -2,7 +2,7 @@
 
 from datetime import date
 from enum import Enum
-from typing import Dict, List, Literal, Optional, Set, Union
+from typing import Dict, List, Literal, Optional, Union
 
 from aind_data_schema_models.modalities import Modality
 from aind_data_schema_models.organizations import Organization
@@ -191,7 +191,7 @@ class Instrument(DataCoreModel):
         return names
 
     @field_serializer("modalities", when_used="json")
-    def serialize_modalities(self, modalities: Set[Modality.ONE_OF]):
+    def serialize_modalities(self, modalities: List[Modality.ONE_OF]):
         """Dynamically serialize modalities based on their type."""
         return sorted(modalities, key=lambda x: x.get("name") if isinstance(x, dict) else x.name)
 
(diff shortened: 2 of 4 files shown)


dyf (approved):
`Set` is how pydantic recommended handling a uniqueness requirement. https://docs.pydantic.dev/2.3/usage/types/set_types/

dbirman:
> `Set` is how pydantic recommended handling a uniqueness requirement. https://docs.pydantic.dev/2.3/usage/types/set_types/

Please try running this, you'll see the error:

```
from pydantic import BaseModel, ConfigDict
from typing import List, Set, Union

class ModalityModel(BaseModel):
    """Base model for modality"""
    model_config = ConfigDict(frozen=True)
    name: str
    abbreviation: str

class ParentModel(BaseModel):
    modalities: Union[List[ModalityModel], Set[ModalityModel]]

# Example usage
modality1 = ModalityModel(name="Magnetic Resonance Imaging", abbreviation="MRI")
modality2 = ModalityModel(name="Computed Tomography", abbreviation="CT")

# Using a list
parent_list = ParentModel(modalities=[modality1, modality2])
print("List Example:")
print(parent_list.model_dump())

# Using a set
parent_set = ParentModel(modalities={modality1, modality2})
print("\nSet Example:")
print(parent_set.model_dump())
```

```
(.venv) dan@db-laptop aind-data-schema % python test.py 
List Example:
{'modalities': [{'name': 'Magnetic Resonance Imaging', 'abbreviation': 'MRI'}, {'name': 'Computed Tomography', 'abbreviation': 'CT'}]}

Set Example:
Traceback (most recent call last):
  File "/Users/dan/proj/aind/aind-data-schema/test.py", line 25, in <module>
    print(parent_set.model_dump())
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/dan/proj/aind/aind-data-schema/.venv/lib/python3.12/site-packages/pydantic/main.py", line 390, in model_dump
    return self.__pydantic_serializer__.to_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: unhashable type: 'dict'
```

bruno-f-cruz:
For reference, this is related to https://github.com/AllenNeuralDynamics/aind-data-schema/issues/960. I raised this exact issue here (https://github.com/AllenNeuralDynamics/aind-data-schema/issues/960#issuecomment-2155055024). While this PR solves the specific problem, please consider the larger issue: These enum-like classes can't be easily hashed, and this will always present problems when trying to use Set or Dict, which is a big limitation IMO.

dbirman:
> For reference, this is related to #960. I raised this exact issue here ([#960 (comment)](https://github.com/AllenNeuralDynamics/aind-data-schema/issues/960#issuecomment-2155055024)). While this PR solves the specific problem, please consider the larger issue: These enum-like classes can't be easily hashed, and this will always present problems when trying to use Set or Dict, which is a big limitation IMO.

Yes, same issue. But we're still trapped by the goal/requirement of keeping information about registries attached to their corresponding objects

bruno-f-cruz:
> > For reference, this is related to #960. I raised this exact issue here ([#960 (comment)](https://github.com/AllenNeuralDynamics/aind-data-schema/issues/960#issuecomment-2155055024)). While this PR solves the specific problem, please consider the larger issue: These enum-like classes can't be easily hashed, and this will always present problems when trying to use Set or Dict, which is a big limitation IMO.
> 
> Yes, same issue. But we're still trapped by the goal/requirement of keeping information about registries attached to their corresponding objects

A few possible solutions come to mind:

- Modify the serialization/schema building behavior to "mock" it as a simple enum (this API is a bit annoying, but we have done it in the past https://github.com/AllenNeuralDynamics/aind-behavior-curriculum/blob/e18289f084ba325df0bfe3c3ada79a5cbf18fcdc/src/aind_behavior_curriculum/curriculum.py#L53). It should be noted that this may not even fix the problem completely depending on what class you deserialize to.
- Attach the information as metadata via annotated (this is a bit clunky...):
```python
from aind_data_schema_models.modalities import Modality
import enum
from typing import Annotated, get_args

class StrModality(enum.Enum):
    BEHAVIOR = "behavior"
    VIDEO = "video"

StrModalityWithMetadata = Annotated[StrModality, Modality]

def materialize_metadata(value: StrModality):
    return get_args(StrModalityWithMetadata)[1].from_abbreviation(value.value)

print(StrModalityWithMetadata.BEHAVIOR)
print(materialize_metadata(StrModality.BEHAVIOR))

a = {StrModalityWithMetadata.BEHAVIOR: 1, StrModalityWithMetadata.VIDEO: 2}
print(a)
```
- If the information is not necessary during schema building, and only later during analysis, one simple way to go about this is to add a single method (e.g. materialize_enum_reference(modality: Enum.Modality) -> FullyReferencedModality). This method decouples the ontology from the code that generates the fully referenced structure. The fully referenced structure could be fetched from a separate database (or even from the library itself, worst-case scenario).
Anyway just something to get the discussion going!

PR #1395, opened 2025-04-05
breaking change: compilation of changes from remaining issues

dbirman:
This PR merges a large number of issues together

1) Channel refactor
The core of the reorganization is now the `ImagingConfig` class, which pulls together a list of `Channel` and either `FieldOfView` or `Image` objects. The coordinate_system is only required when Images are provided (which have transforms).

All of the various optional fields have been pulled out into their own configs (SlapConfig, SampleChamberConfig, etc).

2) Ephys refactor 

On the device side, I've separated EphysProbe and FiberProbe from the assembly versions (which include a manipulator). Assemblies can be used on rigs where the manipulators and probes are attached, while the separate probe classes can be used for chronic recordings.

To match this, there are now separate ManipulatorConfig and ProbeConfig classes, as well as corresponding EphysAssemblyConfig and FiberAssemblyConfig classes. The manipulator configuration stores information about the manipulator local axis positions, independent of the acquisition coordinate system. The probe config class stores information about the target coordinate in both the acquisition coordinate system and optionally in an atlas. ProbeConfig is now generic enough to be re-used by all kinds of devices that get stuck into brains.

Previously there were lasers attached to ephys probes, presumably for NeuropixelsOpto, users should now report those as a Connection

3) Microscope

Added Microscope device and SlapMicroscopeConfig

4) Fiber changes

Removed PatchCordConfig.output_power and .fiber_name fields. Replaced with a Connection.

**Minor changes**

- rename ConnectionData.channel to .port
- Added ProbeImplant.dye
- Cleaned up MRIScan to match the new coordinate system pattern and removed the unnecessary processing_steps fields


src/aind_data_schema/components/configs.py
```diff
@@ -87,72 +76,115 @@ class DetectorConfig(DeviceConfig):
     )
 
 
+class LaserConfig(DeviceConfig):
+    """Description of laser settings in an acquisition"""
+
+    wavelength: int = Field(..., title="Wavelength (nm)")
+    wavelength_unit: SizeUnit = Field(default=SizeUnit.NM, title="Wavelength unit")
+    excitation_power: Optional[Decimal] = Field(default=None, title="Excitation power (mW)")
```
dbirman, line 84: Minor thing, I think we probably copied excitation_ from somewhere, this could probably just be power/power_unit now?

src/aind_data_schema/components/configs.py
```diff
@@ -87,72 +76,115 @@ class DetectorConfig(DeviceConfig):
...
-class FieldOfView(DataModel):
-    """Description of an imaging field of view"""
+class Channel(DataModel):
+    """Description of a channel"""
+
+    channel_name: str = Field(..., title="Channel")
+    intended_measurement: Optional[str] = Field(
+        default=None, title="Intended measurement", description="What signal is this channel measuring"
+    )
+    detector_configuration: DetectorConfig = Field(..., title="Detector configuration")
```
dbirman, line 102: Another minor comment, sometimes we use _configuration and sometimes not, e.g. for excitation_filters. I think when you're already inside of a configuration class we probably don't need the extra suffix?

src/aind_data_schema/components/configs.py
```diff
@@ -87,72 +76,115 @@ class DetectorConfig(DeviceConfig):
...
+
+
+class SinglePlaneConfig(DataModel):
+    """Description of a single plane ophys config"""
+
+    imaging_depth: int = Field(..., title="Imaging depth (um)")
+    imaging_depth_unit: SizeUnit = Field(default=SizeUnit.UM, title="Imaging depth unit")
+
+
+class MultiPlaneConfig(DataModel):
```
dbirman, line 146: A multiplane config appears to be a single-plane config with additional fields (since imaging_depth/_unit are duplicated). Inheritance would simplify these a tiny bit.

src/aind_data_schema/components/configs.py
```diff
@@ -87,72 +76,115 @@ class DetectorConfig(DeviceConfig):
...
-    fov_coordinate_ml: float = Field(..., title="FOV coordinate ML")
-    fov_coordinate_ap: float = Field(..., title="FOV coordinate AP")
+
+
+class FieldOfView(DataModel):
+    """Description of an imaging field of view"""
+
+    channel_name: str = Field(..., title="Channel name")
+    targeted_structure: CCFStructure.ONE_OF = Field(..., title="Targeted structure")
+    fov_coordinate_ml: Decimal = Field(..., title="FOV coordinate ML")
```
dbirman, line 186: AP/ML coordinates should be in a `Coordinate` object (in this PR, they will get changed when I fix coordinate systems downstream of this)

src/aind_data_schema/components/configs.py
```diff
@@ -162,12 +194,29 @@ class Stack(DataModel):
     fov_width: int = Field(..., title="FOV width (pixels)")
     fov_height: int = Field(..., title="FOV height (pixels)")
     fov_size_unit: SizeUnit = Field(default=SizeUnit.PX, title="FOV size unit")
-    magnification: Optional[str] = Field(default=None, title="Magnification")
-    fov_scale_factor: float = Field(..., title="FOV scale factor (um/pixel)")
+    magnification: str = Field(..., title="Magnification")
```
dbirman, line 197: I'm just noticing this is a string, is this redundant with fov_scale_factor or because it's a string is it supposed to represent something else?

dbirman (commented):
All the general logic here makes sense to me, just minor comments in a few places

examples/exaspim_acquisition.py
```diff
@@ -29,45 +29,67 @@
...
+            light_sources=[
+                LaserConfig(
+                    device_name="539251",
+                    wavelength=561,
+                    wavelength_unit=SizeUnit.NM,
+                    excitation_power=200,
+                    excitation_power_unit=PowerUnit.MW,
+                ),
+            ],
+            emission_filters=[
```
dyf, line 77: it's weird that there are no configuration values here.

tests/test_composability_merge.py
```diff
@@ -255,7 +255,7 @@ def test_merge_acquisition(self):
                                             ),
                                         ],
                                     ),
-                                    channel=tile.Channel(
+                                    channel=Channel(
```
dyf, line 258: this needs to move up to `InVitroImagingConfig.channels`

src/aind_data_schema/components/configs.py
```diff
@@ -414,10 +452,38 @@ class Immersion(DataModel):
     refractive_index: Decimal = Field(..., title="Index of refraction")
 
 
+class Image(DataModel):
+    """Description of an image"""
+
+    channel_name: str = Field(..., title="Channel name")
```
dyf, line 458: I don't think we need this anymore. My assumption is that every image in an InVitroImagingConfig has all the channels.

src/aind_data_schema/components/configs.py
```diff
@@ -177,10 +222,12 @@ class SlapAcquisitionType(str, Enum):
...
 
 
-class SlapFieldOfView(FieldOfView):
+class SlapConfig(DataModel):
     """Description of a Slap2 scan"""
 
-    experiment_type: SlapAcquisitionType = Field(..., title="Acquisition type")
+    channels: List[SlapChannel] = Field(..., title="Channel")
+    field_of_view: FieldOfView = Field(..., title="Field of view")
+    slap_experiment_type: SlapAcquisitionType = Field(..., title="Slap experiment type")
```
dyf, line 230: `slap_acquisition_type`

src/aind_data_schema/components/configs.py
```diff
@@ -87,72 +76,115 @@ class DetectorConfig(DeviceConfig):
...
     number_of_plane_repeats_per_volume: int = Field(..., title="Number of repeats per volume")
     number_of_volume_repeats: int = Field(..., title="Number of volume repeats")
-    fov_coordinate_ml: float = Field(..., title="FOV coordinate ML")
-    fov_coordinate_ap: float = Field(..., title="FOV coordinate AP")
+
+
+class FieldOfView(DataModel):
+    """Description of an imaging field of view"""
+
+    channel_name: str = Field(..., title="Channel name")
```
dyf, line 184: a FOV can have multiple channels, yes? same as comment below re: InVitroImagingConfig - either we assume that every FOV has all channels in the config above and remove this field, or we make this a list.

src/aind_data_schema/components/configs.py
```diff
@@ -162,12 +194,25 @@ class Stack(DataModel):
...
-    magnification: Optional[str] = Field(default=None, title="Magnification")
-    fov_scale_factor: float = Field(..., title="FOV scale factor (um/pixel)")
+    magnification: str = Field(..., title="Magnification")
+    fov_scale_factor: Decimal = Field(..., title="FOV scale factor (um/pixel)")
     fov_scale_factor_unit: str = Field(default="um/pixel", title="FOV scale factor unit")
-    frame_rate: Decimal = Field(..., title="Frame rate (Hz)")
+    frame_rate: Decimal = Field(default=..., title="Frame rate (Hz)")
     frame_rate_unit: FrequencyUnit = Field(default=FrequencyUnit.HZ, title="Frame rate unit")
-    targeted_structure: Optional[CCFStructure.ONE_OF] = Field(default=None, title="Targeted structure")
+    planes: List[
```
dyf, line 202: can you really do multiple of these at a given location? should this just be `plane_configuration: ...`?

src/aind_data_schema/components/configs.py
```diff
@@ -162,12 +194,25 @@ class Stack(DataModel):
...
+        Annotated[
+            Union[SinglePlaneConfig, MultiPlaneConfig, StackConfig],
+            Field(discriminator="object_type"),
+        ]
+    ] = Field(..., title="Two photon imaging configurations")
+    notes: Optional[str] = Field(default=None, title="Notes")
+
+
+class TwoPhotonImagingConfig(DataModel):
```
dyf, line 211: Does one photon look different? Is there a more generic name for this?

src/aind_data_schema/components/configs.py
```diff
@@ -87,72 +76,115 @@ class DetectorConfig(DeviceConfig):
...
-    """Description of an imaging field of view"""
+class Channel(DataModel):
+    """Description of a channel"""
+
+    channel_name: str = Field(..., title="Channel")
+    intended_measurement: Optional[str] = Field(
+        default=None, title="Intended measurement", description="What signal is this channel measuring"
+    )
+    detector_configuration: DetectorConfig = Field(..., title="Detector configuration")
+    additional_device_names: Optional[List[DeviceConfig]] = Field(default=None, title="Additional device names")
```
dyf, line 103: what is this for?

src/aind_data_schema/components/configs.py
```diff
@@ -87,72 +76,115 @@ class DetectorConfig(DeviceConfig):
...
+    # excitation
+    light_source_configurations: List[
```
dyf, line 105: I wish this could be simple like emission. the nesting here is rough.

examples/exaspim_acquisition.py
dyf: Or maybe pydantic will just like...do the right thing for single-field objects like this, and my desired syntax just works?

src/aind_data_schema/components/configs.py
```diff
@@ -87,72 +76,115 @@ class DetectorConfig(DeviceConfig):
...
+    # emission
+    emission_filters: Optional[List[DeviceConfig]] = Field(default=None, title="Emission filter names")
+    emission_wavelength: Optional[int] = Field(default=None, title="Emission wavelength")
+    emission_wavelength_unit: Optional[SizeUnit] = Field(default=None, title="Emission wavelength unit")
+
+
+class SlapChannel(Channel):
+    """Description of a channel for Slap"""
+
+    dilation: Optional[int] = Field(default=None, title="Dilation (pixels)")
```
dyf, line 124: are we sure this isn't a deviceconfig, rather than a property of the channel?

examples/exaspim_acquisition.py
dbirman: The reason I requested this is that `DeviceConfig.device_name` is a special field, since it triggers a validator that goes and checks that you also listed this device in the `Acquisition.active_devices` list. If we just let people list strings then we don't get that

src/aind_data_schema/components/configs.py
saskiad: Are there multichannel images? I don't think that's possible.

saskiad: There can be images of the same location from two channels, but those are separate images.

src/aind_data_schema/components/configs.py
saskiad: we need to document each channel independently. We can image the same location on multiple channels, those end up as separate images, and we need to know which image is which. I feel like there is an important disconnect here

src/aind_data_schema/components/configs.py
saskiad: mirrors, dichroics

src/aind_data_schema/components/configs.py
saskiad: all light sources require wavelength and some require power.

src/aind_data_schema/components/configs.py
saskiad: no idea, we're not going to figure it out easily though.

src/aind_data_schema/components/configs.py
saskiad: It's functional information that is ultimately redundant with fov_scale_factor. Operationally people pay attention to what magnification setting they use (2X, 4X, etc), which is what this is, but it should wrap up into the scale factor.

src/aind_data_schema/components/configs.py
saskiad: I just don't think this is worth it - it's the kind of thing that makes users track from one class to the next to the next, and for just one field I don't think it helps much.

dbirman: This shouldn't factor in the decision here I think, since my plan is to fix this in the 2.0 documentation. The documentation tables for each class will pull all the fields down from classes that are inherited, so a MultiPlaneConfig in the documentation will show the fields from SinglePlaneConfig.

src/aind_data_schema/components/configs.py
dbirman: So magnification is like "2X" and fov_scale_factor ends up being "2"? I feel like this can be dropped, unless people use it in a more complicated way sometimes.

examples/exaspim_acquisition.py
dbirman: It is kind of ugly though... We could tag device name fields with field metadata to fix this instead of using the specific name of the field for the validator

src/aind_data_schema/components/configs.py
dbirman: Oh okay got it, this is the scope's magnification setting, but doesn't necessarily tell you about the scale factor. We really need the Microscope class so that we can pull this information out into MicroscopeConfig... coming later I guess

src/aind_data_schema/components/configs.py
dbirman: Refactored

src/aind_data_schema/components/configs.py
dbirman: Done

src/aind_data_schema/components/configs.py
dbirman: Hmm yeah is this supposed to be a list @saskiad? Given that this is one field of view I would have expected it to be mutually exclusive

src/aind_data_schema/components/configs.py
dyf: We can leave this as-is for now, but I bet someone is going to ask us to turn this into an array at some point. We are kind of artificially limiting ourselves by thinking of images as files that are segregated by channel. Once we can acquire to OME-Zarr directly, someone will ask this to be a list.

dbirman:
Flagging this issue https://github.com/AllenNeuralDynamics/aind-data-migration-scripts/issues/342 which influences our choice of where to laser power, because we need to track this on the per-tile(image) level

src/aind_data_schema/components/configs.py
dbirman: I'm pretty sure this isn't supposed to be a list, changing it but we can always change it back

dbirman: It's now much more obvious btw that Stack is just a List[SinglePlaneConfig]. The only difference is that Stack has two extra fields `number_of_plane_repeats_per_volume` and `number_of_volume_repeats`, can someone confirm what these mean?

src/aind_data_schema/components/configs.py
dbirman: I expect this is a property of the Slap Microscope device. I'm going to move it there in the microscope PR https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1273

dbirman:
> Flagging this issue [AllenNeuralDynamics/aind-data-migration-scripts#342](https://github.com/AllenNeuralDynamics/aind-data-migration-scripts/issues/342) which influences our choice of where to laser power, because we need to track this on the per-tile(image) level

Possible solution, some kind of SlapConfig class that captures the power per-side and tracks which images used which path?

dbirman:
> > Flagging this issue [AllenNeuralDynamics/aind-data-migration-scripts#342](https://github.com/AllenNeuralDynamics/aind-data-migration-scripts/issues/342) which influences our choice of where to laser power, because we need to track this on the per-tile(image) level
> 
> Possible solution, some kind of SlapConfig class that captures the power per-side and tracks which images used which path?

Actually this is fine -- a channel has to be defined for each power x wavelength combination and then different channels get used in each image. Gross, but works fine.

src/aind_data_schema/components/coordinates.py
```diff
@@ -190,6 +190,12 @@ class NonlinearTransform(DataModel):
     )
 
 
+TRANSFORM_TYPES = List[Annotated[Union[Translation, Rotation, Scale, Affine], Field(discriminator="object_type")]]
+TRANSFORM_TYPES_NONLIN = List[
```
saskiad, line 194: let's just spell out NONLINEAR.

dbirman: changed

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+
+class PowerFunction(str, Enum):
+    """Power functions"""
+
+    CONSTANT = "Constant"
+    LINEAR = "Linear"
+    EXPONENTIAL = "Exponential"
+
+
+class StimulusModality(str, Enum):
```
saskiad, line 43: Not Deal Breaker: should we move this to the models repo?

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+    CoordinateSystem,
+    Translation,
+    Scale,
+    TRANSFORM_TYPES,
+)
+from aind_data_schema.components.identifiers import Code
+from aind_data_schema.components.wrappers import AssetPath
+
+
+class PowerFunction(str, Enum):
```
saskiad, line 35: We should add an OTHER option here in case people do wonky stuff.

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+    FREE_MOVING = "Free moving"
+    OLFACTORY = "Olfactory"
+    OPTOGENETICS = "Optogenetics"
+    NONE = "None"
+    VIRTUAL_REALITY = "Virtual reality"
+    VISUAL = "Visual"
+    WHEEL_FRICTION = "Wheel friction"
+
+
+class Valence(str, Enum):
```
saskiad, line 56: StimulusValence? 

Additionally, is it of the stimulus? Technically a reward is a stimulus, so that's probably okay. I think people will be confused but I might be willing to live with it.

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+
+
+class LightEmittingDiodeConfig(DeviceConfig):
+    """Configuration of LED settings"""
+
+    power: Optional[Decimal] = Field(default=None, title="Excitation power")
+    power_unit: Optional[PowerUnit] = Field(default=None, title="Excitation power unit")
+
+
+LIGHT_TYPES = List[
```
saskiad, line 149: I don't love calling this LIGHT_TYPES just because I would assume that's for devices rather than configs. Not a deal breaker though, but I'd prefer LIGHT_CONFIGS. I know it should be obvious from context of when it's being used, but I'm not sure it will be for users.

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+        ],
+        Field(discriminator="object_type"),
+    ]
+]
+
+
+class MicroscopeConfig(DeviceConfig):
+    """Configuration of a generic microscope"""
+
+    magnification: Optional[str] = Field(default=None, title="Magnification", description="e.g. 10x")
```
saskiad, line 163: I don't think of magnification as a microscope config. This is which objective is being used. And objectives are not really part of the microscope. Even a scope off the shelf, often you buy the objectives separately, and when they come with it, they are usually spec'ed independently.

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+
+
+class SlapMicroscopeConfig(MicroscopeConfig):
```
saskiad, line 166: This isn't a microscope config. It really is an imaging strategy thing. I think it needs to attach to the Image somehow... let me look when I get to there.

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+        default=None, title="Intended measurement", description="What signal is this channel measuring"
+    )
+    detector: DetectorConfig = Field(..., title="Detector configuration")
+    additional_device_names: Optional[List[DeviceConfig]] = Field(
+        default=None, title="Additional device names", description="Mirrors, dichroics, etc"
+    )
+
+    # excitation
+    light_sources: LIGHT_TYPES = Field(default=[], title="Light source configurations")
+    variable_power: Optional[bool] = Field(
```
saskiad, line 191: The point of this is to indicate the power is specified in the config and not in the device config? 
We need to message it better in the description. I think this suggests that it is actively changing across time, whereas in multiplane imaging the power is constant for each plane, but each plane is different. So variable might not be the name we want here.

src/aind_data_schema/components/configs.py
dbirman: Sure

src/aind_data_schema/components/configs.py
dbirman: So you're saying this is redundant with the objective? Fine with me to remove it.

src/aind_data_schema/components/configs.py
dbirman: I don't know what the original requirements were for this. If this is a per-image thing then we should have `Plane` handle it with targeted_structure, or add a new kind of Plane for SLAP that supports this.

src/aind_data_schema/components/configs.py
dbirman: Moved, once merged I'll take it out here.

src/aind_data_schema/components/configs.py
dbirman: Added

src/aind_data_schema/components/configs.py
dbirman: Removed

src/aind_data_schema/components/configs.py
dbirman: Improved description

saskiad (commented):
I'm not done, but I'm a bit worried about my wifi and all my comments are marked as pending and I don't want to lose them.

src/aind_data_schema/core/procedures.py
```diff
@@ -149,8 +150,9 @@ class Section(DataModel):
     targeted_structure: Optional[CCFStructure.ONE_OF] = Field(default=None, title="Targeted structure")
 
     # Coordinates
-    start_coordinate: Coordinate = Field(..., title="Start coordinate")
-    end_coordinate: Optional[Coordinate] = Field(default=None, title="End coordinate")
+    system_name: str = Field(..., title="Coordinate system name")
```
saskiad, line 153: can this be coodinate_system_name? 
And remind me how this is different from coordinate_system we're using in configs?

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+
+    relative_position: List[AnatomicalRelative] = Field(..., title="Initial relative position")
+
+    # Transform
+    coordinate_system: Optional[CoordinateSystem] = Field(default=None, title="Device coordinate system")
+    transform: Optional[TRANSFORM_TYPES] = Field(
+        default=None,
+        title="Device to acquisition transform",
+        description="Entry coordinate, depth, and rotation in the Acquisition.coordinate_system",
+    )
```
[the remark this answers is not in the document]
dbirman: Volume makes sense. This makes me think that both volume here and duration should be lists, since in many experiments they're going to be varying across trials, thoughts?

src/aind_data_schema/core/procedures.py
saskiad: this is so they can say which coordinate system is used that might be defined in the Procedures.coordinate_system or the PlanarSectioning.coordinate_system? Am I remembering that right?

src/aind_data_schema/core/procedures.py
```diff
@@ -281,7 +283,8 @@ class Craniotomy(DataModel):
     protocol_id: Optional[str] = Field(default=None, title="Protocol ID", description="DOI for protocols.io")
     craniotomy_type: CraniotomyType = Field(..., title="Craniotomy type")
 
-    position: Optional[Union[Coordinate, List[AnatomicalRelative]]] = Field(default=None, title="Craniotomy position")
+    system_name: Optional[str] = Field(default=None, title="Coordinate system name")
```
saskiad, line 286: ditto: can we make these all coordinate_system_name.

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+
+
+class MISModuleConfig(DeviceConfig):
+    """Modular insertion system module configuration"""
+
+    arc_angle: Decimal = Field(..., title="Arc Angle (deg)")
+    module_angle: Decimal = Field(..., title="Module Angle (deg)")
+    rotation_angle: Optional[Decimal] = Field(default=None, title="Rotation Angle (deg)")
+    angle_unit: AngleUnit = Field(default=AngleUnit.DEG, title="Angle unit")
+    calibration_date: Optional[datetime] = Field(
```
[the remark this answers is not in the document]
dbirman: done

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+    dye: Optional[str] = Field(default=None, title="Dye")
+    notes: Optional[str] = Field(default=None, title="Notes")
+
+
+class EphysAssemblyConfig(DeviceConfig):
+    """Group of configurations for an ephys assembly"""
+
+    manipulator: ManipulatorConfig = Field(..., title="Manipulator configuration")
+    probes: List[ProbeConfig] = Field(..., title="Probe configurations")
+
```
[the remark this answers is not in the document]
dbirman: If we want them here then I would go with `modules: Optional[MISModuleConfig]` and if we ever add another kind of module we can add that here as a discriminated list?

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+    """Configuration of auditory speaker configuration"""
+
+    volume: Optional[Decimal] = Field(default=None, title="Volume (dB)")
+    volume_unit: Optional[SoundIntensityUnit] = Field(default=None, title="Volume unit")
+
+
+# EPHYS CONFIGS
+
+
+class MISModuleConfig(DeviceConfig):
```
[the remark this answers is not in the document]
dbirman: I like the idea of attaching to the assembly config, which would let us remove the inheritance from DeviceConfig.

src/aind_data_schema/components/coordinates.py
```diff
@@ -311,41 +285,63 @@ class CoordinateSystemLibrary:
...
             Axis(name=AxisName.X, direction=Direction.LR),
             Axis(name=AxisName.Y, direction=Direction.FB),
-            Axis(name=AxisName.Z, direction=Direction.BT),
+            Axis(name=AxisName.Z, direction=Direction.DU),
         ],
     )
 
-    DEVICE_SIPE = CoordinateSystem(
-        name="CAMERA_SIPE",
+    SIPE_CAMERA_RBF = CoordinateSystem(
```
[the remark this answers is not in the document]
dbirman: Yes Alex and I discussed and this is the only way to directly use the values coming from SIPE without users having to pass them through some kind of transformation. I requested to be looped in if there's ever a discussion about changing this on the SIPE side to make a pitch for using a consistent space.

src/aind_data_schema/core/acquisition.py
```diff
@@ -14,27 +14,29 @@
...
     SpeakerConfig,
-    Stack,
     StimulusModality,
+    SampleChamberConfig,
+    ImagingConfig,
+    SlapMicroscopeConfig,
+    MicroscopeConfig,
+    ProbeConfig,
+    EphysAssemblyConfig,
+    DomeModule,
```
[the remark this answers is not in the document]
dbirman: typo, removed

src/aind_data_schema/components/devices.py
```diff
@@ -58,6 +58,49 @@ class Device(DataModel):
...
+            notes = values.notes
+
+            if manufacturer == Organization.OTHER and not notes:
+                raise ValueError("Device.notes cannot be empty if manufacturer is 'other'")
+
+        return values
+
+
+class PositionedDevice(DataModel):
+    """Device with a position"""
```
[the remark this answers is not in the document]
dbirman: Changed it to `DevicePosition`

src/aind_data_schema/core/acquisition.py
```diff
@@ -14,27 +14,29 @@
     DiscriminatedList,
     GenericModel,
 )
-from aind_data_schema.components.acquisition_configs import (
+from aind_data_schema.components.configs import (
```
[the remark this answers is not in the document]
dbirman: Yes I've been waiting to do an `isort` call until the last minute because it will cause all the imports to jump around and show up as changed

src/aind_data_schema/core/instrument.py
```diff
@@ -72,13 +73,14 @@ class ConnectionDirection(str, Enum):
 
     SEND = "send"
     RECEIVE = "receive"
+    BOTH = "send and receive"
```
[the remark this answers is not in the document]
dbirman: Changed, also fixed capitalization in the enum strings

src/aind_data_schema/core/procedures.py
dbirman: The rules for CoordinateSystem/transforms are as follows:

- Each major file Instrument/Acquisition/Procedures has a top-level CoordinateSystem, these can all be identical or different
- Anytime the code finds a Translation/Rotation/Scale object it checks that there is a system_name field at the same level and that the name matches the coordinate system that was defined. This is really just a reminder for users that the numbers they are giving us are in the right system, since we can't enforce that.
- If it runs into a field called `coordinate_system` it overrides the current CoordinateSystem. This is how devices can have local coordinate systems, for example for use with manipulators

For most users they are always in a bregma-centric coordinate system with the same set of axes, usually ARI, so they don't have to think about this much.

We can rename the field but seeing "coordinate_system_name" all over the place is just a lot of text so I was trying to simplify.

src/aind_data_schema/core/procedures.py
```diff
@@ -548,11 +557,13 @@ class Surgery(DataModel):
...
-        description="Only use this field when different from the Procedures.coordinate_system",
-    )
+        description=(
+            "Only required when the Surgery.coordinate_system " "is different from the Procedures.coordinate_system"
+        ),
+    )  # note: exact field name is used by a validator
 
     # Measured coordinates
-    measured_coordinates: Optional[Dict[Origin, Coordinate]] = Field(
+    measured_coordinates: Optional[Dict[Origin, Translation]] = Field(
```
saskiad, line 566: This syntax confuses me. This takes a dictionary where origin and translation are the keys? I don't think that's right, but I'm just unsure how this works. And we need to know what the measurement is too.

dbirman: This is defining a dictionary where the key: value pairs are {Origin: Translation}

src/aind_data_schema/core/procedures.py
```diff
@@ -591,14 +602,19 @@ class Procedures(DataCoreModel):
     specimen_procedures: List[SpecimenProcedure] = Field(default=[], title="Specimen Procedures")
 
     # Implanted devices
-    implanted_devices: List[Union[FiberProbe, MyomatrixArray]] = Field(default=[], title="Implanted devices")
+    implanted_devices: DiscriminatedList[EphysProbe | FiberProbe | MyomatrixArray] = Field(
+        default=[], title="Implanted devices"
+    )
+    configurations: DiscriminatedList[ProbeConfig | DeviceConfig] = Field(
```
saskiad, line 608: I'm back to my myomatrix confusion. I think we want a MyomatrixConfig here. I think the MyomatrixThread and MyomatrixContact should be worked into the a Config, rather be part of the device. But that won't be part of this PR.

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+    # emission
+    emission_filters: Optional[List[DeviceConfig]] = Field(default=None, title="Emission filters")
+    emission_wavelength: Optional[int] = Field(default=None, title="Emission wavelength")
+    emission_wavelength_unit: Optional[SizeUnit] = Field(default=None, title="Emission wavelength unit")
+
+
+class SlapChannel(Channel):
+    """Configuration of a channel for Slap"""
+
+    dilation: Optional[int] = Field(default=None, title="Dilation")
```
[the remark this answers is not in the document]
dbirman: Changed

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,552 @@
...
+        """Check that only one plane is defined"""
+
+        if any(not isinstance(plane, CoupledPlane) for plane in self.planes) and len(self.planes) > 1:
+            raise ValueError("For single-plane optical physiology only a single Plane should be in PlanarImage.planes")
+
+        return self
+
+
+class PlanarImageStack(PlanarImage):
+    """Description of a stack of N-D images acquired in a specific imaging plane"""
```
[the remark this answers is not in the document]
dbirman: Not clear why it has to be 2D? The configuration doesn't enforce that. I removed the N-D reference though.

src/aind_data_schema/core/procedures.py
saskiad: it might be easy:

`MyomatrixConfig(DeviceConfig):
    threads: List[MyomatrixThread] = Field(..., title="Array threads")`
    
with MyomatrixThread and MyomatrixContact in Procedures rather than device.

src/aind_data_schema/core/acquisition.py
```diff
@@ -14,27 +14,29 @@
...
     AirPuffConfig,
     DetectorConfig,
-    DomeModule,
+    MISModuleConfig,
```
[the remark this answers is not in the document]
dbirman: Changed

src/aind_data_schema/core/procedures.py
saskiad: (that didn't format right, but if it's not clear lmk)

dbirman: This is a mistake caused by a merge, removed.

src/aind_data_schema/core/instrument.py
```diff
@@ -49,6 +48,8 @@
     Tube,
     Wheel,
     Computer,
+    Microscope,
+    MyomatrixArray,
```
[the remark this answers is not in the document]
dbirman: This was a typo, removed

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
dbirman: Done

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
dbirman: Done

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
saskiad: I'm not aware of the reward duration/volume being variable, but we should ask some scientists using them.

src/aind_data_schema/core/procedures.py
dbirman: Sorry -- the other issue with myomatrix showing up in configs was a bug caused by a merge. This pattern here is correct. I'm not sure how the threads/contacts change per acquisition but yes we can create a config for them here or somewhere else.

src/aind_data_schema/core/procedures.py
saskiad: got it. 
But we still want to know what is being measured right? So Bregma-Lambda distance would be 
{"Bregma": (4,0,0)} but how do we know that's Bregma->Lambda?

dbirman: It's just another key. This is copied from the `test_procedures.py` file:

```
                    coordinate_system=CoordinateSystemLibrary.BREGMA_ARID,
                    measured_coordinates={
                        Origin.BREGMA: Translation(
                            translation=[0, 0, 0],
                        ),
                        Origin.LAMBDA: Translation(
                            translation=[-4.1, 0, 0],
                        ),
                    },
```

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
dbirman: I'm not saying anybody is doing it now, rather getting ahead of it. Most classical conditioning tasks would have different reward amounts for different cues running through the same lick spout

src/aind_data_schema/core/procedures.py
saskiad: I think system_name is too vague. I know it's verbose but I think it should be coordinate_system_name

src/aind_data_schema/core/procedures.py
saskiad: oh.... I see. It's not the distance, it's the measured points. yeah, I see that now... got it

src/aind_data_schema/core/procedures.py
saskiad: they don't change per acquisition, the config should be set in the Procedures (similar to ProbeConfig)

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
saskiad: A list is fine. It should be a short list though - e.g. if it's varying across all the trials, it's data and not metadata. But that's not a schema thing, it's a compliance thing.

src/aind_data_schema/components/configs.py
saskiad: exactly. 
`SlapConfig(SinglePlaneConfig):
    dmd_dilation_x: int = Field(..., title="DMD Dilation X (pixels)")
    dmd_dilation_y: int = Field(..., title="DMD Dilation Y (pixels)")
    dilation_unit: SizeUnit = Field(default=SizeUnit.PX, title="Dilation unit")
    target_neuron: Optional[str] = Field(default=None, title="Target neuron")
    target_branch: Optional[str] = Field(default=None, title="Target branch")
    path_to_array_of_frame_rates: str = Field(..., title="Array of frame rates")`

src/aind_data_schema/core/procedures.py
dbirman: So then it's just a device (the current setup), I'll resolve this conversation

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
dbirman: We'll just leave it for now

src/aind_data_schema/components/configs.py
dbirman: Refactored to `SlapPlane` and added in the appropriate spots

examples/exaspim_acquisition.py
```diff
@@ -23,58 +32,87 @@
...
-        filter_names=["Em_600"],
-        detector_name="PMT_1",
-        filter_wheel_index=0,
-    ),
-    notes="these are my notes",
+image1 = ImageSPIM(
+    channel_name="561",
+    file_name=AssetPath("tile_X_0000_Y_0000_Z_0000_CH_561.ims"),
+    dimensions=Scale(scale=[512, 512, 256]),
+    image_to_acquisition_transform=[tile_scale],
```
saskiad, line 46: ditto here. I know the original example was boring because we only put 2 tiles in, but people should see what infor we're expecting them to provide.

examples/multiplane_ophys_acquisition.py
```diff
@@ -29,207 +45,129 @@
...
+                    images=[
+                        PlanarImage(
+                            channel_name="Green channel",
+                            image_to_acquisition_transform=[
+                                Translation(
+                                    translation=[1500, 1500],
+                                ),
+                            ],
+                            dimensions=Scale(
+                                scale=[800, 800],
```
saskiad, line 85: let's keep this consistent with what it was before. 512, 512

examples/multiplane_ophys_acquisition.py
```diff
@@ -29,207 +45,129 @@
...
+                            ),
+                            planes=[
+                                CoupledPlane(
```
saskiad, line 88: each coupled plane needs an index so we know which is coupled to which. There's nothing to maintain the order these show up in.

examples/multiplane_ophys_instrument.py
```diff
@@ -0,0 +1,423 @@
...
+                        0.02298,
+                    ],
+                ),
+            ],
+        ),
+        CameraAssembly(
+            name="Behavior Camera",
+            target=CameraTarget.BODY,
+            relative_position=[AnatomicalRelative.SUPERIOR],
+            camera=Camera(
```
saskiad, line 112: we had wanted to make detectors more reasonable... oh well.

examples/multiplane_ophys_instrument.py
```diff
@@ -0,0 +1,423 @@
...
+                driver="Vimba",
+                driver_version="Vimba GigE Transport Layer 1.6.0",
+            ),
+            lens=Lens(
+                name="Behavior Camera Lens",
+                serial_number=None,
+                manufacturer=Organization.THORLABS,
+                model="MVL6WA",
+                notes=None,
+                focal_length=6.0,
```
saskiad, line 152: didn't we have an issue to remove some of these things? focal length, lens size, etc.

saskiad: #1330

examples/multiplane_ophys_instrument.py
```diff
@@ -0,0 +1,423 @@
...
+            manufacturer=Organization.HAMAMATSU,
+            data_interface=DataInterface.PCIE,
+        ),
+        Objective(
+            name="Mesoscope JenOptik Objective",
+            numerical_aperture=0.8,
+            magnification=3.6,
+            manufacturer=Organization.THORLABS,
+            immersion="water",
+            notes="Part from JenOptik: 14163000",
```
saskiad, line 322: manufacturer isn't supplier. If it's a jenoptik objective, the manufacturer should be JenOptilk. This isn't a schema thing, but this is a bad example.

examples/multiplane_ophys_rig.json
```diff
@@ -0,0 +1,824 @@
+{
```
saskiad, line 1: why is this json file here when none of the others are?

examples/multiplane_ophys_rig.py
```diff
@@ -0,0 +1,624 @@
...
+                name="Behavior Camera Filter",
+                serial_number=None,
+                manufacturer=Organization.SEMROCK,
+                model="FF01-747/33-25",
+                path_to_cad=None,
+                port_index=None,
+                additional_settings=GenericModel(),
+                notes=None,
+                filter_type="Band pass",
+                diameter=None,
```
saskiad, line 193: I thought we removed filter sizes. Did a previous merge get lost?

examples/multiplane_ophys_rig.py
```diff
@@ -0,0 +1,624 @@
...
+                wavelength_unit=SizeUnit.NM,
+                max_aperture="f/1.4",
+            ),
+            filter=Filter(
+                device_type="Filter",
+                name="Behavior Camera Filter",
+                serial_number=None,
+                manufacturer=Organization.SEMROCK,
+                model="FF01-747/33-25",
+                path_to_cad=None,
```
saskiad, line 188: Also, didn't we want to remove path_to_cad and port_index? If people need them, they can use the additional _settings. 
But this was in that same issue on cleaning up devices. I'm confused.

src/aind_data_schema/components/configs.py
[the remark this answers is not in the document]
saskiad: But this still has to be a config for a specific device, right? What device does it get linked to? Or should this be a DataModel?

src/aind_data_schema/components/configs.py
```diff
@@ -0,0 +1,538 @@
...
+        title="Coordinate system",
+        description=(
+            "Required for ImageSPIM objects and when the imaging coordinate system differs from the "
+            "Acquisition.coordinate_system"
+        ),
+    )  # note: exact field name is used by a validator
+    images: List[Annotated[Union[PlanarImage, PlanarImageStack, ImageSPIM], Field(discriminator="object_type")]] = (
+        Field(..., title="Images")
+    )
+    sampling_strategy: Optional[SamplingStrategy] = Field(
```
saskiad, line 321: do we have a validator to make sure this is provided for in vivo imaging?

src/aind_data_schema/core/procedures.py
saskiad: I'm lost - why aren't we adding MyomatrixConfig here? The myomatrix device is the array (manufacturer, array type), then myomatrixConfig is about how it's being used, a list of threads which then uses the thread and contact classes to describe who it was implanted.

saskiad (commented):
Still some confusion - and I think something wonky with the ophys examples.

examples/exaspim_acquisition.py
```diff
@@ -23,58 +32,87 @@
...
-        filter_names=["Em_600"],
-        detector_name="PMT_1",
-        filter_wheel_index=0,
-    ),
-    notes="these are my notes",
+image0 = ImageSPIM(
+    channel_name="488",
+    file_name=AssetPath("tile_X_0000_Y_0000_Z_0000_CH_488.ims"),
+    dimensions=Scale(scale=[512, 512, 256]),
+    image_to_acquisition_transform=[tile_scale],
```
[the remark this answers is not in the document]
dbirman: It was empty in the [original file](https://github.com/AllenNeuralDynamics/aind-data-schema/blob/dev/examples/exaspim_acquisition.py), so I took it out. I see now that your intention was to indicate that a translation is a part of the transform even if the values were zeroes, I'll put it back but with actual values.

examples/multiplane_ophys_acquisition.py
dbirman: They have the indexes? `coupled_plane_index=1,`

examples/multiplane_ophys_instrument.py
dbirman: The issue just says "Do we need any of these fields?", can you be specific about which ones should stay? I don't know what the requirements for Lens were originally.

examples/multiplane_ophys_rig.json
dbirman: Git doing a bad job merging

dbirman: Fixed

examples/multiplane_ophys_rig.py
dbirman: This file shouldn't be there, I removed it.

examples/multiplane_ophys_rig.py
dbirman: This file shouldn't be there, I removed it.

saskiad (approved):

PR #1398, opened 2025-04-05
refactor: split PerformanceMetrics and RewardMetrics

dbirman:
This PR solves the issue of having a lot of empty fields in `PerformanceMetrics` if users didn't provide a reward to the subject or vice versa if they didn't have trials during an acquisition.

### DIFF
--- a/examples/ophys_acquisition.py
+++ b/examples/ophys_acquisition.py
@@ -68,7 +68,7 @@ a = Acquisition(
                         excitation_wavelength=410,
                         excitation_power=10,
                         emission_wavelength=600,
-                    )
+                    ),
                 ),
                 PatchCordConfig(
                     device_name="Patch Cord B",
@@ -84,7 +84,7 @@ a = Acquisition(
                         excitation_wavelength=560,
                         excitation_power=7,
                         emission_wavelength=700,
-                    )
+                    ),
                 ),
             ],
             notes="Internal trigger.",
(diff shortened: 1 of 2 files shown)


saskiad:
I don't understand the reason to separate the reward volume from the trials information?

saskiad:
I also don't see what problem this solves - I think it makes it more confusing because I think these are performance metrics. Separating performance metrics and reward metrics with a weird arbitrary line seems awkward to me.

dbirman:
> I also don't see what problem this solves - I think it makes it more confusing because I think these are performance metrics. Separating performance metrics and reward metrics with a weird arbitrary line seems awkward to me.

I see this as another example of the "Optional things become List[Unions]" pattern that we've established. You can have trials without reward (for example in passive stim) and you can have reward without trials (for example in habituation), so why not separate them so that there aren't a bunch of empty fields showing up with users needing to pay attention which optional fields apply to their situation.

saskiad:
The ask for the trial summary was specifically in the context of behavior. For a passive stimulus this is encoded in the stimulus parameters in the stimulus epoch. It's only in a behavioral context where this number is not fixed and people want to know what it is. That's why it has the other fields for trials rewarded, etc.

dbirman:
> The ask for the trial summary was specifically in the context of behavior. For a passive stimulus this is encoded in the stimulus parameters in the stimulus epoch. It's only in a behavioral context where this number is not fixed and people want to know what it is. That's why it has the other fields for trials rewarded, etc.

Well sure, but the whole point of 2.0 is to abstract the specific requests users made into simplified/standardized principles. Anyway I need the rc out at this point so I'm going to give up on this one and just change it to be backward-compatible proof by using the list[union] pattern.

dbirman:
This isn't possible, Union doesn't allow single types. Closing for now, backward incompatible change probably for 3.0

PR #1399, opened 2025-04-07
refactor: change sectioning to use coordinate systems

dbirman:
This PR refactors the `Sectioning` class making a few changes:

- Sectioning now uses the `CoordinateSystem` concept, either inheriting from procedures or defining its own system
- Sections are now defined by a series of cuts and an orientation. The number of output IDs must be one less than the number of cuts.
- Partial slices are defined using a List[AnatomicalRelative]. For example, a coronal slice that only includes the left hemisphere would be marked `partial_slice=[AnatomicalRelative.LEFT]`.

### DIFF
--- a/src/aind_data_schema/core/procedures.py
+++ b/src/aind_data_schema/core/procedures.py
@@ -230,6 +230,7 @@ class Section(DataModel):
     """Description of a slice of brain tissue"""
 
     output_specimen_id: str = Field(..., title="Specimen ID")
+    targeted_structure: Optional[CCFStructure.ONE_OF] = Field(default=None, title="Targeted structure")
 
     # Coordinates
     start_coordinate: Coordinate = Field(..., title="Start coordinate")
@@ -264,7 +265,6 @@ class PlanarSectioning(DataModel):
         title="Sectioning coordinate system",
         description="Only required if different from the Procedures.coordinate_system",
     )
-    targeted_structure: CCFStructure.ONE_OF = Field(..., title="Targeted structure")
 
     sections: List[Section] = Field(..., title="Sections")
     section_orientation: SectionOrientation = Field(..., title="Sectioning orientation")
--- a/src/aind_data_schema/core/procedures.py
+++ b/src/aind_data_schema/core/procedures.py
@@ -22,7 +22,6 @@ from aind_data_schema_models.units import (
     VolumeUnit,
 )
 from pydantic import Field, SkipValidation, field_validator, model_validator
-from pydantic_core.core_schema import ValidationInfo
 from typing_extensions import Annotated
 
 from aind_data_schema.base import AwareDatetimeWithDefault, DataCoreModel, DataModel
@@ -32,6 +31,7 @@ from aind_data_schema.components.identifiers import Person
 from aind_data_schema.components.reagent import Reagent
 from aind_data_schema.utils.merge import merge_notes
 from aind_data_schema.utils.validators import subject_specimen_id_compatibility
+from aind_data_schema.utils.exceptions import FieldLengthMismatch
 
 
 class ImmunolabelClass(str, Enum):
@@ -72,13 +72,6 @@ class SectionOrientation(str, Enum):
     CORONAL = "Coronal"
     SAGITTAL = "Sagittal"
     TRANSVERSE = "Transverse"
-
-
-class SectionStrategy(str, Enum):
-    """Section strategy"""
-
-    WHOLE = "Whole Brain"
-    HEMI = "Hemi Brain"
 
 
 class ProtectiveMaterial(str, Enum):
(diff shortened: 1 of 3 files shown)


src/aind_data_schema/core/procedures.py
```diff
@@ -233,32 +226,35 @@ class Antibody(Reagent):
...
-    section_thickness: Decimal = Field(..., title="Section thickness")
-    section_thickness_unit: SizeUnit = Field(default=SizeUnit.MM, title="Section thickness unit")
-    section_distance_from_reference: Decimal = Field(..., title="Section distance from reference")
-    section_distance_unit: SizeUnit = Field(default=SizeUnit.MM, title="Distance unit")
-    section_strategy: SectionStrategy = Field(..., title="Slice strategy")
 
-    reference: Origin = Field(..., title="Reference origin")
-
-    targeted_structure: CCFStructure.ONE_OF = Field(..., title="Targeted structure")
+    section_cuts: List[Coordinate] = Field(..., title="Section start coordinates", min_length=1)
```
saskiad, line 240: this gives us where is cut, but not how thick, correct? We need the thickness

dbirman: You have to specify both cuts, that's why the section_cuts list has to be 1 longer than the number of specimens produced

src/aind_data_schema/core/procedures.py
```diff
@@ -233,32 +226,35 @@ class Antibody(Reagent):
...
+    section_orientation: SectionOrientation = Field(..., title="Sectioning orientation")
+    partial_slice: Optional[List[AnatomicalRelative]] = Field(
```
saskiad, line 242: so for a slice of only one hemisphere, the idea would be to say "RIGHT" if we're sectioning the right hemisphere? This feels incomplete. i.e. is RIGHT always the full right hemisphere or could it be just a part of the right hemisphere? 
But without modeling the full coordinates of the chunk that's being sectioned I'm not sure what the alternative is

src/aind_data_schema/core/procedures.py
saskiad: i.e. the distance between sections might not be the same as the thickness of the section.

saskiad (commented):
see comments

src/aind_data_schema/core/procedures.py
saskiad: I think it needs to be an explicit field. Asking users to figure that out is a bit much.

dbirman: Hmm, I don't follow... what situation are you thinking of where the thickness is not the same as the distance between cuts? We can use mappers so that users can continue to provide us with the data they provided for, but I'm pretty sure the information here is identical to the previous definition (which was origin + offset to the first cut, then orientation + thickness).

src/aind_data_schema/core/procedures.py
dbirman: Yeah this is already more information than we had before which was strictly "whole brain" or "one hemisphere"

Sectioning never had directions at all before, so we didn't even know left/right! Pretty sure we're doing better here and you're right that we can do even better later once we have a chance to talk to people who are doing other kinds of sectioning and give them non-planar sectioning classes.

src/aind_data_schema/core/procedures.py
dbirman: Oh wait I just got it, your point is if people do like 10um sections but space them every mm. I'll change it so you explicitly list the start + end coordinate of each slice.

dbirman: Done

saskiad: I really think it will be clearer to users to have an explicit thickness field rather than two coordinates. Especially with multiple slices, I don't think the list of list of coordinates will be clear.

src/aind_data_schema/core/procedures.py
```diff
@@ -233,32 +226,48 @@ class Antibody(Reagent):
...
 
-        if output_id_len != expected_len:
-            raise AssertionError("List of output specimen ids does not match the number of slices.")
-        return v
+    coordinate_system: Optional[CoordinateSystem] = Field(
+        default=None,
+        title="Sectioning coordinate system",
+        description="Only required if different from the Procedures.coordinate_system",
+    )
+    targeted_structure: CCFStructure.ONE_OF = Field(..., title="Targeted structure")
```
saskiad, line 267: this maybe can't be required. Or maybe it doesn't belong in the planar sectioning section? There are lots of structures in a single plane, which is the one that matters. but not planar sectioning can be for a specific structure...

saskiad: wait - but put it in the Section class, and make it optional. because there can be sections for different planes that would be targeting different structures. I'm thinking of those barseq slides where they have 4 slices from different atlas planes. So

saskiad (approved):
Just move the targeted structure into the section class and make it optional.

src/aind_data_schema/core/procedures.py
dbirman: done

PR #1407, opened 2025-04-15
refactor: replace DataProcess.pipeline_steps with Processing.pipelines and DataProcess.pipeline_name

dbirman:
This PR removes the field `DataProcess.pipeline_steps` in favor of having each `DataProcess` point to its pipeline using a name. Pipelines are now listed in `Processing.pipelines`.

### DIFF
--- a/src/aind_data_schema/core/processing.py
+++ b/src/aind_data_schema/core/processing.py
@@ -12,7 +12,7 @@ from pydantic import Field, SkipValidation, ValidationInfo, field_validator, mod
 from aind_data_schema.base import AwareDatetimeWithDefault, DataCoreModel, DataModel, GenericModel, GenericModelType
 from aind_data_schema.components.identifiers import Code, Person
 from aind_data_schema.components.wrappers import AssetPath
-from aind_data_schema.utils.merge import merge_notes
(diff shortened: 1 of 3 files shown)


src/aind_data_schema/core/processing.py
```diff
@@ -106,14 +101,22 @@ class Processing(DataCoreModel):
...
+        default=None,
+        title="Pipelines",
+        description=(
+            "For processing done with pipelines, list the repositories here. Pipelines must use the name field "
+            ",and be referenced in the pipeline_name field of a DataProcess."
+        ),
+    )
     notes: Optional[str] = Field(default=None, title="Notes")
 
     dependency_graph: Dict[str, List[str]] = Field(
```
saskiad, line 114: should this be required? Not everything has a pipeline so it won't all need a dependency graph like this?

dbirman: Tom and I debated on this but settled that we really want this data and if we put a default on it we won't get it. We want users to provide an empty dictionary in situations where there are no dependencies. But I think the vast majority of processing happens in more than one capsule even if it isn't explicitly in a pipeline. I'll have a section in the documentation about this

Issue #1403, opened 2025-04-14
Need a validation step to ensure that fiber connections from session (acquisition) to procedures


dougollerenshaw:
As I'm looking closer at fiber data, I'm seeing that many fiber sessions have more fiber connections than there are implanted fibers. For example, for `{'_id': "14579108-63f0-4c22-851a-32e4ca86edf4"}`, here are the fiber connections in the data streams, indicating four fiber connections:

<img width="342" alt="Image" src="https://github.com/user-attachments/assets/64fd75f5-038d-479a-9cf9-c356169a91dd" />

But in the procedures, there are only two implanted fibers:

<img width="393" alt="Image" src="https://github.com/user-attachments/assets/5c6abe6c-f3f7-46f4-b87e-4b75c4a6247c" />

<img width="431" alt="Image" src="https://github.com/user-attachments/assets/2b9aa7f3-77e6-4a2e-a2f9-b0c40f529702" />

Can there be some validation step that ensures that this isn't the case? I think it points to incorrect session metadata being generated. @saskiad pointed out that this is difficult because this will only be apparent after the session (acquisition in 2.0) object is joined with procedures.

dbirman:
@dougollerenshaw I think we're good on this with the PR I sent you for the 2.0 release, the `Connection` objects in `DataStream.connections` will get validated against devices in either instrument or procedures. Let me know if there's an additional validator I need though.

dougollerenshaw:
@dbirman , given that #1408 is merged, should we close this issue?

PR #1408, opened 2025-04-15
refactor: Add DataStream.connections and Procedures.implanted_devices

dbirman:
This PR makes it possible to create new `Connection` objects in an acquisition. This is useful for fiber, where day-to-day the fibers might be connected in different patterns or not connected at all. Since this PR impacts implant devices I'm also making that change here, so `Procedures.implanted_devices` is now the location where you put an ephys or fiber probe, and you refer to that object by name in the corresponding implant procedure.

Other changes:
- Noticed that LickSpoutAssembly was missing it's `name` field, causing problems for one of the new device name validators
- Moved a bunch of Reagant subclasses into the reagent.py file, should have done that a while ago.
- Moved some validators up to Metadata where we now have to ensure that devices listed in `DataStream.active_devices` are *either* in instrument or in procedures
- `ProbeImplant` replaces `OphysProbe` and `FiberImplant` and the missing chronic Neuropixels implant procedure. All three now require you to place your device in the `implanted_devices` list and then reference by name.

### DIFF
--- a/examples/ophys_procedures.py
+++ b/examples/ophys_procedures.py
@@ -10,11 +10,9 @@ from aind_data_schema.components.identifiers import Person
 from aind_data_schema.core.procedures import (
     Anaesthetic,
     Antibody,
-    FiberImplant,
-    FiberProbe,
     Headframe,
     BrainInjection,
-    OphysProbe,
+    ProbeImplant,
     Perfusion,
     Procedures,
     SpecimenProcedure,
@@ -24,6 +22,7 @@ from aind_data_schema.core.procedures import (
     InjectionDynamics,
     InjectionProfile,
 )
+from aind_data_schema.components.devices import FiberProbe
 from aind_data_schema_models.units import VolumeUnit
 from aind_data_schema_models.brain_atlas import CCFStructure
 from aind_data_schema.components.coordinates import CoordinateSystemLibrary, Coordinate
@@ -31,8 +30,19 @@ from aind_data_schema.components.coordinates import CoordinateSystemLibrary, Coo
 t = datetime.datetime(2022, 7, 12, 7, 00, 00)
 t2 = datetime.datetime(2022, 9, 23, 10, 22, 00)
 
+implanted_devices = [
+    FiberProbe(
+        name="Probe A",
+        core_diameter=200,
+        numerical_aperture=0.37,
+        ferrule_material="Ceramic",
+        total_length=0.5,
+    )
+]
+
 p = Procedures(
     subject_id="625100",
+    implanted_devices=implanted_devices,
     subject_procedures=[
         Surgery(
             start_date=t.date(),
@@ -80,24 +90,14 @@ p = Procedures(
                     ],
                     targeted_structure=CCFStructure.VTA,
                 ),
-                FiberImplant(
+                ProbeImplant(
                     protocol_id="TO ENTER",
-                    probes=[
-                        OphysProbe(
-                            ophys_probe=FiberProbe(
-                                name="Probe A",
-                                core_diameter=200,
-                                numerical_aperture=0.37,
-                                ferrule_material="Ceramic",
-                                total_length=0.5,
-                            ),
-                            targeted_structure=CCFStructure.VTA,
-                            coordinate=Coordinate(
-                                system_name="BREGMA_ARID",
-                                position=[-600, -3050, 0, 4200],
-                            ),
-                        )
-                    ],
+                    implanted_device_names=["Probe A"],
+                    targeted_structure=CCFStructure.VTA,
+                    coordinate=Coordinate(
+                        system_name="BREGMA_ARID",
+                        position=[-600, -3050, 0, 4200],
+                    ),
                 ),
             ],
         ),
--- a/src/aind_data_schema/components/devices.py
+++ b/src/aind_data_schema/components/devices.py
@@ -38,6 +38,7 @@ from aind_data_schema_models.units import (
     UnitlessUnit,
     VoltageUnit,
 )
+from aind_data_schema_models.mouse_anatomy import MouseAnatomyModel
 from pydantic import Field, ValidationInfo, field_validator, model_validator
 from typing_extensions import Annotated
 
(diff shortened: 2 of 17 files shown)


dbirman:
Need to decide how to handle validation for implanted_devices -- recursive validator looking for field names? `implanted_device` or `implanted_devices`?

dougollerenshaw:
Note: This closes https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1403#issuecomment-2807454117

Also related to:
https://github.com/AllenNeuralDynamics/aind-metadata-mapper/issues/277
https://github.com/AllenNeuralDynamics/aind-fiber-photometry-pipeline/issues/5

examples/ophys_procedures.py
```diff
@@ -24,15 +22,27 @@
...
 )
+from aind_data_schema.components.devices import FiberProbe
 from aind_data_schema_models.units import VolumeUnit
 from aind_data_schema_models.brain_atlas import CCFStructure
 from aind_data_schema.components.coordinates import CoordinateSystemLibrary, Coordinate
 
 t = datetime.datetime(2022, 7, 12, 7, 00, 00)
 t2 = datetime.datetime(2022, 9, 23, 10, 22, 00)
 
+implanted_devices = [
```
dougollerenshaw, line 33: I know this is just an example, but it might be worth emphasizing in a comment that these fields should be modified by the scientist to match their experiment. And maybe for the sake of consistency with actual fiber data, the example name should be "Fiber 0"?

dougollerenshaw:
Would it be possible to create an example that shows how the fiber connections will be defined on a day to day basis? From what I understand, different mice have different numbers of fiber connections (it seems experiment/project specific), so this is presumably something that scientists will need to be updating for each experiment/acquisition. I think it'd be helpful to demonstrate what that will look like.

saskiad:
> Would it be possible to create an example that shows how the fiber connections will be defined on a day to day basis? From what I understand, different mice have different numbers of fiber connections (it seems experiment/project specific), so this is presumably something that scientists will need to be updating for each experiment/acquisition. I think it'd be helpful to demonstrate what that should look like.

does the ophys_acquisition example not do this?

src/aind_data_schema/components/devices.py
```diff
@@ -635,7 +637,28 @@ class Scanner(Device):
...
+        ..., title="Location of ground electrode", description="Use GroundWireLocations"
+    )
+    contacts: List[MyomatrixContact] = Field(..., title="Contacts")
+
+
 class MyomatrixArray(Device):
     """Description of a Myomatrix array"""
 
     array_type: MyomatrixArrayType = Field(..., title="Array type")
+    threads: List[MyomatrixThread] = Field(..., title="Array threads")
```
saskiad, line 664: have you confirmed with Lili that threads can be embedded into array like this? I have a vague recollection that there was a reason we didn't do it this way originally, but I can't recall what that reason was. 
I think the fact that that the Array is the Device, but the procedural information about how it's implanted lives in the `contacts` that lives `threads` made it hard to nest the threads in the array like this.

src/aind_data_schema/components/devices.py
```diff
@@ -635,7 +637,28 @@ class Scanner(Device):
...
+        ..., title="Relative position", description="Position relative to procedures coordinate system"
+    )
+    muscle: MouseAnatomyModel = Field(..., title="Muscle of contact insertion", description="Use MouseEmgMuscles")
+    in_muscle: bool = Field(..., title="In muscle")
+
+
+class MyomatrixThread(DataModel):
+    """Description of a thread of a myomatrix array"""
+
+    ground_electrode_location: MouseAnatomyModel = Field(
```
saskiad, line 654: is ground_electrode per thread? Isn't it the same ground for the whole thing?

src/aind_data_schema/core/procedures.py
```diff
@@ -586,33 +499,15 @@ class WaterRestriction(DataModel):
...
-
 class MyomatrixInsertion(DataModel):
     """Description of a Myomatrix array insertion for EMG"""
 
-    ground_electrode: GroundWireImplant = Field(..., title="Ground electrode")
     protocol_id: Optional[str] = Field(default=None, title="Protocol ID", description="DOI for protocols.io")
-    myomatrix_array: MyomatrixArray = Field(..., title="Myomatrix array")
-    threads: List[MyomatrixThread] = Field(..., title="Array threads")
+
+    ground_electrode: GroundWireImplant = Field(..., title="Ground electrode")
```
saskiad, line 507: did we add this from the original myomatrix thing?? We have two separate ground_electrode fields for this - I'm confused.

saskiad (changes requested):
I'm a little nervous about the rejiggering of myomatrix - can we make sure this makes sense?

saskiad:
> > > Would it be possible to create an example that shows how the fiber connections will be defined on a day to day basis? From what I understand, different mice have different numbers of fiber connections (it seems experiment/project specific), so this is presumably something that scientists will need to be updating for each experiment/acquisition. I think it'd be helpful to demonstrate what that should look like.
> > 
> > 
> > does the ophys_acquisition example not do this?
> 
> That would be this file, right? https://github.com/AllenNeuralDynamics/aind-data-schema/blob/release-v2.0.0/examples/ophys_acquisition.py
> 
> I don't see any reference to connections there.

It uses the `PatchCordConfig`. Connections is primarily connections to daq devices that don't convey additional parameters.

src/aind_data_schema/components/devices.py
dbirman: I didn't change this I just moved it into the devices, this is what we had before https://github.com/AllenNeuralDynamics/aind-data-schema/blob/31793c313472c5bad33d0627e75c80bc0386495f/src/aind_data_schema/core/procedures.py#L627

src/aind_data_schema/core/procedures.py
dbirman: This didn't change? I just moved it below the protocol_id. In general I put the protocol_id at the top of classes where it shows up.

src/aind_data_schema/core/acquisition.py
```diff
@@ -130,6 +131,12 @@ class DataStream(DataModel):
         ]
     ] = Field(..., title="Device configurations")
 
+    connections: List[Connection] = Field(
+        default=[],
+        title="Connections",
+        description="Connections that are specific to this acquisition, and are not present in the Instrument",
```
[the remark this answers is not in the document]
dbirman: Do you mean that you want an example here?

src/aind_data_schema/components/devices.py
dbirman: Lili doesn't know why these were separated. I will leave them combined and we can pull them apart at the release candidate if we need to

dbirman:
I'm force merging this to allow me to move forward with development. We can revisit these decisions at the release candidate stage if we need to.

PR #1412, opened 2025-04-16
refactor: rename ConnectionData.channel to .port

dbirman:


dbirman:
This is merged into the channel PR

PR #1415, opened 2025-04-22
docs: adds Code of Conduct

jtyoung84:
- Adds Code of Conduct as per https://github.com/AllenNeuralDynamics/aind-library-template/issues/83


dbirman (approved):
Not sure if Saskia wants to look through before merge, seems fine to me though

PR #1419, opened 2025-04-30
Simplify types by defining Annotated Field with discriminator in base

tmchartrand:
This is the simplification I mentioned to you - should be functionally equivalent I believe, and tests all run.
Defines `Discriminated = Annotated[T, Field(discriminator="object_type")]` and uses that to cut out a bunch of boilerplate


dbirman:
I had to pin the aind-data-schema-models dependency to get tests to pass

dbirman (approved):

Issue #1424, opened 2025-05-19
[2.0] Separating instrument_id components


dbirman:
@dougollerenshaw and I have been discussing how to improve the `Instrument.instrument_id` field to better support the needs of scientists. From these discussions we've decided that the three fields that currently get combined into an `instrument_id` need to be separated. This simplifies maintenance of instrument metadata while still maintaining our ability to easily query for a unique "rig" (the combination of data acquisition hardware, at some location, at some point in time)

- `Instrument.instrument_id: str` no additional validation on this field. Within AIND we will enforce requirements on this field through the metadata-mapper.
- `Instrument.room: Optional[str]` required in AIND, enforced by the metadata-mapper
- `Instrument.modification_date` remains the same

This fulfills the requirement that we be able to query the unique state of a rig, which is now done by using a query against all three fields. Yes this is more complicated than a single str, but given the garbage we get in the rig_id field right now this is a necessary simplification of the schema. Here are some examples of the variety of things we get...

```
"323_EPHYS1_2024-06-17"
"323_EPHYS1-OPTO_001"
"323_EPHYS1"
"323_EPHYS1_NM_717121_2024-06-16_11-39-34"
"323_EPHYS2-RF_2023-07-28_01"
"323_EPHYS2-RF_2023-08-03_01"  # this one changes every day... not clear if modifications actually happened
"323_Rig3_111023"
"342_BEHDEV.B-1_20240401"
"347T4_FIP2_1"
"422-mesoscope-20220218"
"426_4A_20241126"
"442 Bergamo 2p photostim"
"446-8-A_20240925"
```

You get the point. Only three of these are valid against the regex in v1.4.

Note this change is independent from discussion about what exactly should go into the `instrument_id`

axlec:
Hi Doug, in the above example, when you query the various compIDs for NP.0 what do you do with all of the compIDs that are returned? 

How/where do those get saved?

If we only care about systems that record data or interface with the mouse and each of those should have a compID. The question becomes, how do we exclude certain compID's that were not pertinent to the experiment?

[...]

PR #1425, opened 2025-05-19
refactor: separate location/instrument_id/modification_date

dbirman:
PR separates the location/instrument_id/modification_date fields. See linked issue

Note: this is unrelated to the decision of what the format of the `instrument_id` should be. This will be enforced by the aind-metadata-mapper.

Closes https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1424


dyf (changes requested):
let's call it `location`, to avoid semantics of what a room is.

PR #1428, opened 2025-05-29
1197 docs cleanup 2

dbirman:
Second PR for ongoing docs cleanup. Make sure you are looking at this documentation page https://aind-data-schema.readthedocs.io/en/1197-docs-cleanup-2/

As part of this PR I realized the connections should really be moved since they are re-used in both acquisition and instrument.


dbirman:
@dbirman note to self: add information to docs about how platform has been replaced (tags and acquisition_type)

dbirman:
Can you clarify where you think this should go: `Clarify model metadata schema vs models used within the data schemas`? The models subpage seems pretty clear?

leesuyee:
> Can you clarify where you think this should go: `Clarify model metadata schema vs models used within the data schemas`? The models subpage seems pretty clear?

I agree the models subpage has a good description. Maybe adding a subpage for "Models" that describes the logic for organizing the models into components vs registries vs core. (Perhaps that's an obvious thing and that's my own confusion).

saskiad:
Front page:
- [ ] Probably need some information to explain the Models section and how it relates to the Core section

Generating metadata 
- [ ] can we not use `OTHER` as the Institution. 
- [ ] Remove the "# all subjects are mice" It's already established that this example is a mouse, and I read this as saying that all subjects must be mice.

Coordinate Systems:
- [ ] "These can be identical, or different, depending on your needs." I might suggest "The coordinate system can be the same or different in each of these files, depending on your needs"
- [ ] The bit about coordinates in an Atlas doesn't make sense to me
- [ ] I think we should document the abbreviations we have for axis names and directions. eg. ML is medial lateral and LR is left right. SI is superior inferior which we prefer to DV. 
- [ ] Request additions to the CoordinateSystem library - maybe we should say if there is a coordinate system you use frequently that is not in the library, you can request an addition... I want to avoid people thinking they have to have their system in the library to use it
- [ ] I think this page needs more information about describing device positions in coordinate systems (e.g. need to define coordinate system of the device and the transform of it's origin position in the instrument's coordinate system...)

Procedures:
- [ ] make the header subject procedure vs specimen procedure. 
- [ ] specimen id = subject id_### First, it's underscore, second, there's no concept of a specimen number. This is the specimen number, The current phrasing suggests that there is a specimen number that's different from this.
- [ ] Remove the perfusion header - this isn't info about perfusions really, it's a way of unpacking subject vs specimen procedures.
- [ ] remove the ethics review id vs protocol id. The confusion existed because we use "iacuc protocol" and the use of the word protocol was what confused people. We do want people to know what the protocol id is and where they can find them.
- [ ] Add information on what TARS identifiers are

Instrument
- [ ] Maintenance isn't the only reason an instrument changes - especially for phys/behavior. 

... more to come

dbirman:
>  Add information on what TARS identifiers are

We can't put this here since it's AIND specific, we need to put this somewhere else.

saskiad:
> > Add information on what TARS identifiers are
> 
> We can't put this here since it's AIND specific, we need to put this somewhere else.

I was thinking something to the effect that this point to the AIBS TARS database and is not expected for external users at this time.

dbirman:
> 
> I was thinking something to the effect that this point to the AIBS TARS database and is not expected for external users at this time.

I don't think this should be in the documentation where it's already hard to get people to read stuff. Instead I would say we should combine the identifiers fields on ViralMaterial, see: https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1448, and then in the description of the TarsVirusIdentifier we should label it as an internal database.

dbirman:
@dbirman note to self: remove the ## Model definitions header from core files if there are no models (e.g. for subject and procedures)

dbirman:
@dbirman note to self: improve explanation of connections and configurations in acquisition. Go through metadata in general and look for opportunities to add better descriptions and better docstrings.

saskiad (approved):

PR #1444, opened 2025-06-05
fix: allow uncontrolled strings as coordinate system origins

dbirman:
This PR allows users to use uncontrolled strings to describe the origin of a coordinate system. We strongly prefer that they give us a defined Origin from the enum, but we need this for other kinds of situations like blood vessels, etc.


dbirman:
Debating separately whether this is worth pursuing: if vocabulary is small then it's unnecessary.

PR #1449, opened 2025-06-07
1448 20 combine viralmaterial identifiers field

dbirman:
PR combines the tars_identifier and addgene_id fields into one.

### DIFF
--- a/examples/ophys_procedures.py
+++ b/examples/ophys_procedures.py
@@ -80,11 +80,13 @@ p = Procedures(
                         ViralMaterial(
                             material_type="Virus",
                             name="AAV2/1-Syn-Flex-ChrimsonR-tdT",
-                            identifiers=[PIDName(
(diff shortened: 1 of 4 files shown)


saskiad:
I suppose a Union allows both to be provided.... I'm reluctant to remove the addgene_id field because the specificity is valuable. Give me a second to think

dbirman:
> I suppose a Union allows both to be provided.... I'm reluctant to remove the addgene_id field because the specificity is valuable. Give me a second to think

I made it a list so that you could include both if you wanted to. The root problem though is that the TARS field doesn't exist for people outside AIND, so it doesn't make sense to have a field that provides that specific functionality.

dbirman:
@saskiad this is still a problem, we can't have fields for things that don't exist outside AIND. But I'm closing because I have enough stuff to do to close out 2.0

PR #1465, opened 2025-06-13
refactor: flatten QC

dbirman:
This PR flattens the QC, entirely removing the `QCEvaluation` concept.

What used to be the QCEvaluation name is now expected to be put into the `QCMetric.tags` list. New features at the level of `QualityControl` include:

- `allow_tag_failures`: works the way that the `allow_failed_metrics` used to work inside of evaluations
- default_grouping: the set of tags that will be used to group QCMetrics in the QC portal by default. Users will be able to modify this as desired
- status: Instead of being the single overall status this now evaluates the per-tag, per-modality, and per-stage status of all metrics. This is closer to what shows up in the QC Portal status table anyways so this should save some complexity there.

If you have any documentation comments please still **approve this PR**, I will put the doc changes on the other PR https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1428 so that I can move forward with the upgrader.

### DIFF
--- a/src/aind_data_schema/core/quality_control.py
+++ b/src/aind_data_schema/core/quality_control.py
@@ -181,7 +181,7 @@ class QualityControl(DataCoreModel):
 
             # Compute modality statuses
             for modality in self.modalities:
-                computed_status[modality] = self.evaluate_status(modality=modality)
+                computed_status[modality.abbreviation] = self.evaluate_status(modality=modality)
 
             # Compute stage statuses
             for stage in self.stages:
--- a/tests/test_quality_control.py
+++ b/tests/test_quality_control.py
@@ -412,6 +412,21 @@ class QualityControlTests(unittest.TestCase):
         # Confirm that the status filters work
         q = QualityControl(metrics=test_metrics, default_grouping=["test_group", "test_group2", "tag1"])
 
+        # Check that the status field was built correctly
+        self.assertEqual(q.status, {
+            # Stages
+            "Processing": Status.PASS,
+            "Raw data": Status.FAIL,
+            # Modalities
+            "behavior": Status.FAIL,
+            "behavior-videos": Status.PENDING,
+            "ecephys": Status.PASS,
+            # Tags
+            "test_group": Status.PASS,
+            "test_group2": Status.FAIL,
+            "tag1": Status.PENDING,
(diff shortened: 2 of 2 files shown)


dyf (commented):
Should we support overriding `QualityControl.status` for a given tag? At the moment it appears to be a "pass if any matching metrics pass". Is it possible to say "`plane0` passed" even though one of its metrics might have failed?

examples/exaspim_quality_control.py
```diff
@@ -5,105 +5,116 @@
...
-            description="Unknown",
-            value="Thalamus",
-            status_history=[QCStatus(evaluator="Shirali Amin", status=Status.PASS, timestamp=datetime(2024, 7, 12))],
-        ),
-    ],
-    created=datetime(2024, 7, 12),
-)
+metrics = [
+    # N001 metrics
+    QCMetric(
```
dyf, line 12: "Horta space" -> "Sample space", also below

dyf (changes requested):
nitpick

examples/exaspim_quality_control.py
```diff
@@ -13,7 +13,7 @@
         name="Horta coordinates",
```
dyf, line 13: "Sample coordinates" sorry, also below

dyf (changes requested):
nitpick again

saskiad:
(thinking out loud) I still think we want a modality level status. I like the flexible groupings, especially for the visualizations, but at some point we have to say "the ophys passes, the videos don't". Are we just hoping that people assign their groupings appropriately?

dbirman:
> (thinking out loud) I still think we want a modality level status. I like the flexible groupings, especially for the visualizations, but at some point we have to say "the ophys passes, the videos don't". Are we just hoping that people assign their groupings appropriately?

This still exists. In the status dictionary you get each modality, take a look at the evaluate_status() function you'll see it calculating that.

examples/exaspim_quality_control.py
dbirman: fixed

examples/exaspim_quality_control.py
dbirman: fixed

PR #1472, opened 2025-06-18
feat: exploring how to add support for curriculums

dbirman:
PR adds curriculum support for training protocols:

- TrainingProtocol.curriculum_code: Code <- stores the actual code for the curriculum
- StimulusEpoch.training_protocol_name <- name pointing to the training protocol
- StimulusEpoch.curriculum_status <- uncontrolled string for users to put things like "Off curriculum", "Training Stage 1", etc

### DIFF
--- a/src/aind_data_schema/core/metadata.py
+++ b/src/aind_data_schema/core/metadata.py
@@ -262,7 +262,7 @@ class Metadata(DataCoreModel):
             for stimulus_epoch in self.acquisition.stimulus_epochs:
                 if stimulus_epoch.training_protocol_name:
                     if stimulus_epoch.training_protocol_name not in training_protocol_names:
-                        raise ValueError(
+                        warnings.warn(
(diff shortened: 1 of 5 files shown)


src/aind_data_schema/core/metadata.py
```diff
@@ -246,6 +247,28 @@ def validate_acquisition_connections(self):
 
         return self
 
+    @model_validator(mode="after")
+    def validate_training_protocol_references(self):
```
saskiad, line 251: This is going to error out until we have a system for tracking training protocols and ingesting this into the procedures. (I'm not arguing against this validator, I'm just acknowledging that our current procedures don't have this information because the ingest doesn't exist)

saskiad (approved):

src/aind_data_schema/core/metadata.py
dbirman: Hmm, okay well I'll have it throw a warning for now and we'll upgrade it to an error later.

PR #1474, opened 2025-06-24
feat: add library IMAGE_XYZ coordinate system

dbirman:
PR adds a new library coordinate system for images

### DIFF
--- a/src/aind_data_schema/components/coordinates.py
+++ b/src/aind_data_schema/components/coordinates.py
@@ -389,6 +389,17 @@ class CoordinateSystemLibrary:
         ],
     )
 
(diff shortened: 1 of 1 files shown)


dbirman:
> Looks good. Seems to consistent with the other definitions in the `CoordinateSystemLibrary` class. One larger suggestion would be to add inline comments or maybe a "detailed_description" parameter to each that better describes what each of these is. But since this is a small addition to the already-established class, that seems a bit out of scope for now.

I was hoping to be exhaustive enough to avoid a description but let's revisit that as more people start touching 2.0 based on whether people run into difficulty understanding these.

PR #1480, opened 2025-06-30
Feat 1445 stain2

### LINKED ISSUE #1445: Create a single fluorescent stain class
- [ ] allow union of targets: oligo probe, protein, small molecule
- [ ] use a fluorophore class that has type (e.g. AlexaFluor) and excitation wavelength separated
- [ ] optional sequence class for custom designed protein stains

related:

- [ ] oligo probe has a required sequence field (it is essentially only this) and in fact for HCR we don't have the sequence for our probes. We only have the gene sequence which comes from the accension id

saskiad:
closes #1445

### DIFF
--- a/src/aind_data_schema/components/reagent.py
+++ b/src/aind_data_schema/components/reagent.py
@@ -13,14 +13,6 @@ from pydantic import Field
 from aind_data_schema.base import DataModel, Discriminated
 
 
-class ImmunolabelClass(str, Enum):
-    """Type of antibodies"""
-
-    PRIMARY = "Primary"
-    SECONDARY = "Secondary"
-    CONJUGATE = "Conjugate"
-
-
 class StainType(str, Enum):
     """Stain types for probes describing what is being labeled"""
 
@@ -55,9 +47,6 @@ class Reagent(DataModel):
     rrid: Optional[PIDName] = Field(default=None, title="Research Resource ID")
     lot_number: str = Field(..., title="Lot number")
     expiration_date: Optional[date] = Field(default=None, title="Lot expiration date")
-
-
-# Gene-targeting oligonucleotide probes
 
 
 class OligoProbe(DataModel):
--- a/src/aind_data_schema/components/reagent.py
+++ b/src/aind_data_schema/components/reagent.py
@@ -1,34 +1,16 @@
 """Classes to define reagents"""
 
 from datetime import date
-from enum import Enum
 from typing import List, Optional
 
 from aind_data_schema_models.organizations import Organization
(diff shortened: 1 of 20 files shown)


dbirman:
I fixed one of the bugs on this PR, there's still some more issues though.

src/aind_data_schema/components/reagent.py
```diff
@@ -55,66 +49,48 @@ class Reagent(DataModel):
...
     """set of probes used in BarSEQ"""
 
-    gene_probes: List[GeneProbes] = Field(..., title="Gene probes")
-    # IDT scale stuff?
+    gene_probes: List[GeneProbe] = Field(..., title="Gene probes")
 
 
-class Readout(Reagent):
-    """Description of a readout"""
+class ProteinProbe(Reagent):
```
dbirman, line 72: The three Probe classes could inherit from a single `Probe` parent class and have a field `probe_target` to make them easier to understand/maintain.

src/aind_data_schema/components/reagent.py
```diff
@@ -29,20 +21,22 @@ class StainType(str, Enum):
...
 
 
-class Fluorophore(str, Enum):
-    """Fluorophores used in HCR and Immunolabeling"""
+class FluorophoreType(str, Enum):
+    """Fluorophores types"""
+
+    ALEXA = "Alexa Fluor"
+    ATTO = "Atto"
+    CF = "CF"
```
dbirman, line 29: Is CF short for something longer?

dbirman (approved):
Everything makes sense, two quick questions to check

src/aind_data_schema/components/reagent.py
saskiad: Tim said it was just CF. I'll poke around and see if anything comes up

saskiad: https://biotium.com/technology/cf-dyes/
Looks like it just CF

src/aind_data_schema/components/reagent.py
saskiad: the GeneProbe, ProteinProbe and SmallMoleculeProbe, but not OligoProbe ... would that be weird?

dbirman: Hmm. I think it's fine to change it and leave OligoProbe out. Most people are going to be interacting with these classes starting from the model list in the union, so I think they mostly won't notice that there's another one of these with the same name (that isn't in the union)

src/aind_data_schema/components/reagent.py
dbirman: Resolving

PR #1487, opened 2025-07-02
feat: ensuring all __add__ functions work

dbirman:
- Missed instrument entirely somehow? Not sure what happened but now it works
- Added support for Procedures.coordinate_system, which was missing

### DIFF
--- a/src/aind_data_schema/core/instrument.py
+++ b/src/aind_data_schema/core/instrument.py
@@ -52,6 +52,7 @@ from aind_data_schema.components.devices import (
     Wheel,
 )
 from aind_data_schema.components.measurements import CALIBRATIONS
+from aind_data_schema.utils.merge import merge_notes, merge_optional_list
 from aind_data_schema.utils.validators import recursive_get_all_names
 
 # Define the mapping of modalities to their required device types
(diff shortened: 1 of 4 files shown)


src/aind_data_schema/core/instrument.py
```diff
@@ -262,3 +263,60 @@ def validate_modality_device_dependencies(cls, value):
...
+                f"coordinate_system: {self.coordinate_system}/{other.coordinate_system}\n"
+                f"temperature_control: {self.temperature_control}/{other.temperature_control}"
+            )
+
+        # Combine modalities and sort
+        combined_modalities = list(set(self.modalities + other.modalities))
+        combined_modalities = sorted(combined_modalities, key=lambda x: x.abbreviation)
+
+        # Use the latest modification date
+        latest_modification_date = max(self.modification_date, other.modification_date)
```
saskiad, line 297: I wonder how we handle modification date. Because the use case I think of for merging intrument files is I have an ephys rig (mod date Feb 1) and a behavior set up (mod date June 1) and I want to use that behavior set up on my ephys rig. And when if I merge them today, the modification date should be today, not the most recent of the two components...
Is it possible to have modification date be something that can be passed into the add function, and defaults to today?

saskiad, line 297: I'd expect the modification date to be when the two instruments are combined. Isn't that a modification?

src/aind_data_schema/core/instrument.py
```diff
@@ -262,3 +263,60 @@ def validate_modality_device_dependencies(cls, value):
...
+
+        # Check for schema version incompatibility
+        if self.schema_version != other.schema_version:
+            raise ValueError(
+                "Cannot combine Instrument objects with different schema versions: "
+                f"{self.schema_version} and {other.schema_version}"
+            )
+
+        # Check for incompatible key fields
+        inst_id_check = self.instrument_id != other.instrument_id
```
saskiad, line 278: If instrument id and location have to match from the start, I think this limits how this can be used pretty narrowly.

dbirman: By design!

saskiad (approved):

PR #1491, opened 2025-07-03
feat: validate datetime objects in the schema (except QC and model)

dbirman:
This PR defines two kinds of time-based validation:

1. Name validation: this checks that the `Acquisition.acquisition_end_time` is matched with the `DataDescription.name` "creation_time" field by checking to make sure that the creation_time occurs on or after the same day the acquisition occurred.
2. Field validation: a recursive validator runs through all fields and if they are annotated with `TimeValidation` enums they get compared to the acquisition start and end times. There are three kinds of validation: before, after, and between. Before checks that the date is prior to the end time, after checks that the date is after the start time, and between does what it sounds like.

### DIFF
--- a/docs/source/aind_data_schema_models/brain_atlas.md
+++ b/docs/source/aind_data_schema_models/brain_atlas.md
@@ -4,7 +4,7 @@
 
 ### CCFv3
 
-CCFStructure
+CCFv3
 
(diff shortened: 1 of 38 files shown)


dbirman:
Test are not passing until the ophys_procedure example gets fixed (other PR)

PR #1499, opened 2025-07-08
1492 20 olfactometerconfig

dbirman:
PR moves OlfactometerConfig from stimulus.py to configs.py, leaving just the `stimulus_name` field in the stimulus classes.

### DIFF
--- a/docs/source/components/configs.md
+++ b/docs/source/components/configs.md
@@ -298,7 +298,7 @@ MRI scan sequence
 | `OTHER` | `Other` |
 
 
-### OlfactometerChannel
+### OlfactometerChannelInfo
 
 Configuration of a channel in an olfactometer
 
@@ -316,7 +316,7 @@ Configuration of olfactometer
 
 | Field | Type | Title (Description) |
 |-------|------|-------------|
-| `channel_configs` | List[[OlfactometerChannel](devices.md#olfactometerchannel)] | Channel configurations (List of channels with their odorant and concentration) |
+| `channel_configs` | List[[OlfactometerChannelInfo](#olfactometerchannelinfo)] | Channel configurations (List of channels with their odorant and concentration) |
 | `device_name` | `str` | Device name (Must match a device defined in the instrument.json) |
 
 


dbirman:
> Don't you need to add OlfactometerConfig to the Acquisition configurations list?

Yes good catch

dbirman:
- [ ] Add a test that checks that all subclasses of DeviceConfig are options in the configurations Union

src/aind_data_schema/components/configs.py
```diff
@@ -414,6 +415,25 @@ class SpeakerConfig(DeviceConfig):
...
+    channel_index: int = Field(..., title="Channel index")
+    odorant: str = Field(..., title="Odorant")
+    dilution: Decimal = Field(..., title="Odorant dilution")
+    dilution_unit: ConcentrationUnit = Field(default=ConcentrationUnit.VOLUME_PERCENT, title="Dilution unit")
+
+
+class OlfactometerConfig(DeviceConfig):
+    """Configuration of olfactometer"""
+
+    channel_configs: List[OlfactometerChannelConfig] = Field(
```
saskiad, line 430: in the stimulus class we had a stimulus_name field that I think can be beneficial so that the same config settings can be grouped together. i.e. there a configurations that will get used regularly, and being able to identify the key configurations easily will be beneficial. 
I can imagine that the config isn't the place for this information to live given the component-config organization, but might be worth thinking about how best to do that.

src/aind_data_schema/components/configs.py
```diff
@@ -414,6 +415,25 @@ class SpeakerConfig(DeviceConfig):
     volume_unit: Optional[SoundIntensityUnit] = Field(default=None, title="Volume unit")
 
 
+class OlfactometerChannelConfig(DataModel):
+    """Configuration of a channel in an olfactometer"""
+
+    channel_index: int = Field(..., title="Channel index")
+    odorant: str = Field(..., title="Odorant")
```
saskiad, line 422: might want to ask science team if it makes sense for odorant to be a Reagent or not.

saskiad (approved):
I think it looks good - two things that might be worth thinking about but aren't deal breakers.

saskiad:
@dbirman did this get merged? Do we want it merged?

src/aind_data_schema/components/configs.py
dbirman: There is still the `StimulusEpoch.stimulus_name` to capture this, which is better I think than putting it in the config or the stimulus parameters, which in 2.0 would have been hidden in `StimulusEpoch.code.parameters`

src/aind_data_schema/components/configs.py
dbirman: Tiffany says that although they are reagents there's no database storing this information, so I vote we punt this down to 3.0 and make an effort then to capture this in dataverse and then expose it in the metadata-service.

PR #1540, opened 2025-08-26
fix: deprecate the channel_index field

dbirman:
PR deprecates the channel_index field and throws a warning if a user uses the field. Also changes how the documentation handles the `deprecated` annotation to display ~~strikethrough~~ on the field name and a bold [DEPRECATED] in the description.

### DIFF
--- a/docs/source/components/devices.md
+++ b/docs/source/components/devices.md
@@ -179,7 +179,7 @@ Named input or output channel on a DAQ device
 | `channel_name` | `str` |  |
 | `channel_type` | [DaqChannelType](../aind_data_schema_models/devices.md#daqchanneltype) |  |
 | `port` | `Optional[int]` |  |
-| `channel_index` | `Optional[int]` |  |
+| <del>`channel_index`</del> | `Optional[int]` | **[DEPRECATED]** Use DAQChannel.port instead. |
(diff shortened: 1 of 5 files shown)


bruno-f-cruz:
You may want to take advatange of the `deprecated` decorator provided by PEP702. Pydantic seems to support it with the [following syntax](https://docs.pydantic.dev/latest/concepts/fields/#deprecated-as-a-string):

```python

from typing import Annotated

from pydantic import BaseModel, Field


class Model(BaseModel):
    deprecated_field: Annotated[int, Field(deprecated='This is deprecated')]


print(Model.model_json_schema()['properties']['deprecated_field'])
#> {'deprecated': True, 'title': 'Deprecated Field', 'type': 'integer'}
```

This has a few advantages in terms of maintainability (you dont need to raise a warning), but is also picked up by some language server tools. As a final plus, it also makes its way into the json-schema in case downstream tools (e.g. code gen) are relying on these fields for something.

dbirman:
Ah I knew this had to be somewhere, guess I searched the wrong keyword in the docs...

dbirman:
- [ ] Upgrade doc generator to grab deprecated field and annotate the markdown table in a nice way (strikethrough?)

dyf (approved):

PR #1553, opened 2025-09-04
refactor: replace use of DiscList with DiscTuple for TRANSFORM_TYPES

dbirman:
This PR creates a new type `DicriminatedTuple` and replaces the use of DiscriminatedList in TRANSFORM_TYPES. The purpose of this is to make it more clear in `BrainInjection` objects that there is a list of injections (usually multiple injections at different depths on the same trajectory), each of which is a set of Translations and/or Rotations. Previously you would get things like:

`coordinates = [[Inj0_Translation, Rotation], [Inj1_Translation, Rotation]]`

Now you get:

`coordinates = [(Inj0_Translation, Rotation), (Inj1_Translation, Rotation)]`

Making it a tiny bit more clear that this a set of transforms being applied together. This is a backwards compatible change since JSON arrays can deserialize to list or tuple.

### DIFF
--- a/docs/source/example_workflow/example_workflow.py
+++ b/docs/source/example_workflow/example_workflow.py
@@ -128,7 +128,7 @@ def generate_procedures(
             )
         ],
         targeted_structure=getattr(CCFv3, brain_area.upper()),
-        coordinates=[coord],  # Note: this is a list, because we could have multiple depths
+        coordinates=[(coord)],  # Note: this is a list, because we could have multiple depths
         dynamics=[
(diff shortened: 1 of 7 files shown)


dyf:
But lists of transforms do have an application order, especially since TRANSFORM_TYPES includes affine and TRANSFORM_TYPES_NONLINEAR also has a warp field.

dbirman:
Actually order matters here because the InjectionDynamics objects are in a list. So this plan doesn't work! Oh well

PR #1568, opened 2025-09-17
chore: bump adsm version

dbirman:


dougollerenshaw (approved):
Nice. I see 4.2.11 on pypi as of about 11 minutes ago! 
https://pypi.org/project/aind-data-schema-models/#history

Should there also be a corresponding tagged release here?
https://github.com/AllenNeuralDynamics/aind-data-schema-models
Looks like the most recent release is 2.0.0 in Feburary 2025

dbirman:
> Nice. I see 4.2.11 on pypi as of about 11 minutes ago! https://pypi.org/project/aind-data-schema-models/#history
> 
> Should there also be a corresponding tagged release here? https://github.com/AllenNeuralDynamics/aind-data-schema-models Looks like the most recent release is 2.0.0 in Feburary 2025

The tags are generated automatically, see https://github.com/AllenNeuralDynamics/aind-data-schema-models/tags. If you want to write notes for a release then you have to do it manually. We should probably make notes for every major/minor version.

PR #1588, opened 2025-10-13
feat: add from_derived and from_data_description helpers

dbirman:
PR adds `DataDescription.from_derived` and `from_data_description` functions.

- from_derived correctly builds a derived DataDescription from an already derived DataDescription. This includes a hotfix for a bug where the new derived DataDescription would include the entire name of its parent but should only include the raw and the new process + date/time.
- from_data_description wraps from_raw and from_derived so users don't have to think about which one to call.
- modifies get_or_default to raise errors if a user doesn't provide a value anywhere for a required field (i.e. don't use the default PydanticUndefined value)

### DIFF
--- a/tests/test_data_description.py
+++ b/tests/test_data_description.py
@@ -593,7 +593,7 @@ class DataDescriptionTest(unittest.TestCase):
 
         # Try to create derived data - should trigger the PydanticUndefined error path
         with self.assertRaises(ValueError) as context:
-            DataDescription.from_raw(base_data, "test_process", creation_time=dt)
+            DataDescription.from_data_description(base_data, "test_process", creation_time=dt)
 
         # Should raise error about the missing required field
         self.assertIn("Required field investigators must have a value", str(context.exception))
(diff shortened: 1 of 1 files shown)


src/aind_data_schema/core/data_description.py
```diff
@@ -250,3 +241,125 @@ def get_or_default(field_name: str) -> Any:
...
+        Returns
+        -------
+        DataDescription
+            New DERIVED DataDescription with name based on the original input, not the full derived name
+
+        """
+        if data_description.data_level != DataLevel.DERIVED:
+            raise ValueError(f"Input data_description must have data_level=DERIVED, got {data_description.data_level}")
+
+        def get_or_default(field_name: str) -> Any:
```
saskiad: I'm a little unsure of this behavior. If a field has a default value, but that field isn't provided for a record, do we really want to use the default? I'm having a hard time coming up with good examples for this, but I think that if we are taking a record and propagating it, if there isn't a value for a field, we shouldn't assume that it should be the default in the derived result. Because in that case, I would have assumed the default value would have been used in the raw version...

dbirman:
Changed so that required fields must be pulled from somewhere, they will not return defaults

PR #1604, opened 2025-10-22
Clarify timestamp format

### LINKED ISSUE #1603: More completely specify datetime format?
The documentation states that timestamps should be `datetime (timezone-aware)`[ here](https://aind-data-schema.readthedocs.io/en/latest/acquisition.html#id1). Eg:

Image

If I'm not mistake, UTC timestamps of format `2025-07-18T19:32:35.275046Z` technically meet this definition. However, I've been told by @saskiad that we want timestamps to contain timezone offsets to make it clear both which timezone the data was collected in and actual walltime in that timezone, eg `2025-07-18T12:32:35.275046-07:00`.

We should be more explicit about this in the documentation. We should add to the documentation to say something like `datetime: ISO 8601 with timezone offset (preserves local time)`

And maybe add a section to the documentation that is more verbose/unambiguous, such as
```
All timestamps should be in ISO 8601 format with timezone offset representing local time. This allows for queries that consider local time when experiments were performed. 

Format: YYYY-MM-DDTHH:MM:SS.ssssss±HH:MM (not Z)
Example: "2025-07-18T12:32:35.275046-07:00"
Invalid: "2025-07-18T19:32:35.275046Z" (UTC loses local time context)
```

dougollerenshaw:
Adds to documentation to clarify that timestamps should be formatted as ISO 8601 timezone aware and, more specifically, that they should be in local walltime with timezone offset.

This came to me verbally from @saskiad but as far as I can tell it was not explicitly documented.

Closes #1603

### DIFF
--- a/schemas/acquisition_schema.json
+++ b/schemas/acquisition_schema.json
@@ -515,6 +515,18 @@
                "title": "Description",
                "type": "string"
             },
+            "measured_at": {
+               "anyOf": [
+                  {
+                     "type": "string"
+                  },
+                  {
+                     "type": "null"
+                  }
+               ],
+               "default": null,
+               "title": "Measurement location"
+            },
             "input": {
                "description": "Calibration input",
                "items": {
@@ -582,6 +594,9 @@
                   },
                   {
                      "$ref": "#/$defs/PressureUnit"
+                  },
+                  {
+                     "$ref": "#/$defs/TorqueUnit"
                   }
                ],
                "title": "Input unit"
@@ -666,6 +681,9 @@
                   },
                   {
                      "$ref": "#/$defs/PressureUnit"
+                  },
+                  {
+                     "$ref": "#/$defs/TorqueUnit"
                   }
                ],
                "title": "Output unit"
@@ -740,6 +758,34 @@
             "fit_type"
          ],
          "title": "CalibrationFit",
+         "type": "object"
+      },
+      "CatheterConfig": {
+         "additionalProperties": false,
+         "description": "Configuration of a catheter",
+         "properties": {
+            "object_type": {
+               "const": "Catheter config",
+               "default": "Catheter config",
+               "title": "Object Type",
+               "type": "string"
+            },
+            "device_name": {
+               "description": "Must match a device defined in the instrument.json",
+               "title": "Device name",
+               "type": "string"
+            },
+            "targeted_structure": {
+               "$ref": "#/$defs/MouseAnatomyModel",
+               "description": "Use options from MouseBloodVessels",
+               "title": "Targeted blood vessel"
+            }
+         },
+         "required": [
+            "device_name",
+            "targeted_structure"
+         ],
+         "title": "CatheterConfig",
          "type": "object"
       },
       "Channel": {
@@ -1483,6 +1529,7 @@
                   "discriminator": {
                      "mapping": {
                         "Air puff config": "#/$defs/AirPuffConfig",
+                        "Catheter config": "#/$defs/CatheterConfig",
                         "Detector config": "#/$defs/DetectorConfig",
                         "Ephys assembly config": "#/$defs/EphysAssemblyConfig",
                         "Fiber assembly config": "#/$defs/FiberAssemblyConfig",
@@ -1541,6 +1588,9 @@
                      },
                      {
                         "$ref": "#/$defs/EphysAssemblyConfig"
+                     },
+                     {
+                        "$ref": "#/$defs/CatheterConfig"
                      }
                   ]
                },
@@ -2086,6 +2136,19 @@
                ],
                "default": null,
                "title": "Excitation power unit"
+            },
+            "power_measured_at": {
+               "anyOf": [
+                  {
+                     "type": "string"
+                  },
+                  {
+                     "type": "null"
+                  }
+               ],
+               "default": null,
+               "description": "For example: objective, patch cable, etc",
+               "title": "Power measurement location"
             }
          },
          "required": [
@@ -2244,6 +2307,19 @@
                ],
                "default": null,
                "title": "Excitation power unit"
+            },
+            "power_measured_at": {
+               "anyOf": [
+                  {
+                     "type": "string"
+                  },
+                  {
+                     "type": "null"
+                  }
+               ],
+               "default": null,
+               "description": "For example: objective, patch cable, etc",
+               "title": "Power measurement location"
             }
          },
          "required": [
@@ -2660,6 +2736,29 @@
          "title": "MemoryUnit",
          "type": "string"
       },
+      "MouseAnatomyModel": {
+         "description": "Base model for mouse anatomy",
+         "properties": {
+            "name": {
+               "title": "Name",
+               "type": "string"
+            },
+            "registry": {
+               "$ref": "#/$defs/Registry"
+            },
+            "registry_identifier": {
+               "title": "Registry Identifier",
+               "type": "string"
+            }
+         },
+         "required": [
+            "name",
+            "registry",
+            "registry_identifier"
+         ],
+         "title": "MouseAnatomyModel",
+         "type": "object"
+      },
       "MousePlatformConfig": {
          "additionalProperties": false,
          "description": "Configuration for mouse platforms",
@@ -2711,6 +2810,80 @@
          ],
          "title": "MriScanSequence",
          "type": "string"
+      },
+      "OlfactometerChannelInfo": {
+         "additionalProperties": false,
+         "description": "Configuration of a channel in an olfactometer",
+         "properties": {
+            "object_type": {
+               "const": "Olfactometer channel info",
+               "default": "Olfactometer channel info",
+               "title": "Object Type",
+               "type": "string"
+            },
+            "channel_index": {
+               "title": "Channel index",
+               "type": "integer"
+            },
+            "odorant": {
+               "title": "Odorant",
+               "type": "string"
+            },
+            "dilution": {
+               "anyOf": [
+                  {
+                     "type": "number"
+                  },
+                  {
+                     "type": "string"
+                  }
+               ],
+               "title": "Odorant dilution"
+            },
+            "dilution_unit": {
+               "$ref": "#/$defs/ConcentrationUnit",
+               "default": "% v/v",
+               "title": "Dilution unit"
+            }
+         },
+         "required": [
+            "channel_index",
+            "odorant",
+            "dilution"
+         ],
+         "title": "OlfactometerChannelInfo",
+         "type": "object"
+      },
+      "OlfactometerConfig": {
+         "additionalProperties": false,
+         "description": "Configuration of olfactometer",
+         "properties": {
+            "object_type": {
+               "const": "Olfactometer config",
+               "default": "Olfactometer config",
+               "title": "Object Type",
+               "type": "string"
+            },
+            "device_name": {
+               "description": "Must match a device defined in the instrument.json",
+               "title": "Device name",
+               "type": "string"
+            },
+            "channel_configs": {
+               "description": "List of channels with their odorant and concentration",
+               "items": {
+                  "$ref": "#/$defs/OlfactometerChannelInfo"
+               },
+               "title": "Channel configurations",
+               "type": "array"
+            }
+         },
+         "required": [
+            "device_name",
+            "channel_configs"
+         ],
+         "title": "OlfactometerConfig",
+         "type": "object"
       },
       "Origin": {
          "description": "Origin positions for coordinate systems",
@@ -3184,6 +3357,18 @@
                "default": "Power measured for various power or percentage input strengths",
                "title": "Description",
                "type": "string"
+            },
+            "measured_at": {
+               "anyOf": [
+                  {
+                     "type": "string"
+                  },
+                  {
+                     "type": "null"
+                  }
+               ],
+               "default": null,
+               "title": "Measurement location"
             },
             "input": {
                "description": "Power, voltage, or percentage input strength",
@@ -4529,6 +4714,7 @@
                         "Laser config": "#/$defs/LaserConfig",
                         "Light emitting diode config": "#/$defs/LightEmittingDiodeConfig",
                         "Mouse platform config": "#/$defs/MousePlatformConfig",
+                        "Olfactometer config": "#/$defs/OlfactometerConfig",
                         "Speaker config": "#/$defs/SpeakerConfig"
                      },
                      "propertyName": "object_type"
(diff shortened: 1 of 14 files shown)


dbirman:
> Do you mean that the timestamp in the name should be in local time? I just took a look at a sample asset on docdb with the name "behavior_789923_2025-09-11_16-09-19", which doesn't include the local timezone. Is that name correct as-is?

Right sorry, to be specific the acquisition_start_time ~~and acquisition_end_time~~ (actually just the start for v2.0 metadata) I think have to be local for this to work, because the name is constructed from those. The name itself doesn't show the timezone information.

All other times can be in any timezone, since we can always re-reference them to the timezone used in the acquisition fields.

dbirman:
Btw you have to try pretty hard to mess this up, because we have these wrapper classes `AwareDatetimeWithDefault` which automatically pull the local timezone from the hardware you're running on. So if you just pass `datetime.now()` everything works fine, you have to explicitly pass it a UTC timezone to force it to bypass that.

dougollerenshaw:
> Btw you have to try pretty hard to mess this up, because we have these wrapper classes `AwareDatetimeWithDefault` which automatically pull the local timezone from the hardware you're running on. So if you just pass `datetime.now()` everything works fine, you have to explicitly pass it a UTC timezone to force it to bypass that.

But without clear documentation, I don't think it's fair to say this was wrong. That's the problem I'm trying to fix with this PR.

docs/source/general.md
```diff
@@ -44,6 +44,18 @@ database stops working.
 
 Learn more about about [AIND data organization](https://docs.allenneuraldynamics.org/en/latest/metadata/philosophy.html) philosophy and the rules we use to standardize asset naming.
 
+## Datetime format requirements
+
+All timestamps should be in ISO 8601 format with timezone offset representing local time. This allows for queries that consider local time when experiments were performed.
```
dbirman: Given the discussion on the PR I think the right approach here is to say something like: 

Whenever possible, provide timestamps as Python `datetime` objects. `aind-data-schema` automatically adds your local timezone and stores this in the metadata files during serialization.

When this isn't possible, provide timestamps as ISO formatted strings ... [etc]

dbirman: Also I'm not sure you should hide this in the general page, it kind of deserves it's own page.

dougollerenshaw:
I just used the metadata mcp to do a little searching. Existing records seem to be all over the map in whether or not `acquisition.session_start_time` is in timezone aware and, if so, whether it's in local time with offset or UTC:

| Name | Project Name | Session Start Time |
|------|--------------|-------------------|
| SmartSPIM_750727_2024-10-22_15-29-56 | null | `2024-10-22T15:29:56-07:00` |
| SmartSPIM_762792_2024-11-15_12-26-06 | CTY Genetic Tools | `2024-11-15T12:26:06-08:00` |
| SmartSPIM_802549_2025-06-17_19-59-34 | BG AAV Toolbox | `2025-06-17T19:59:34-07:00` |
| SmartSPIM_683673_2023-11-02_22-56-18 | null | `2023-11-02T22:56:18` |
| SmartSPIM_732965_2024-09-11_21-13-31 | null | `2024-09-11T21:13:31` |
| SmartSPIM_607463_2023-08-16_15-50-42 | null | `2023-08-16T15:50:42` |
| behavior_754574_2024-11-22_09-39-39 | Cognitive flexibility in patch foraging | `2024-11-22T17:39:39.614464Z` |
| behavior_754571_2024-09-19_10-59-38 | Cognitive flexibility in patch foraging | `2024-09-19T17:59:38.319328Z` |
| behavior_745305_2024-12-06_11-25-45 | Cognitive flexibility in patch foraging | `2024-12-06T19:25:45.104210Z` |

dbirman (approved):
I forgot that for some mappers now there's a risk that if you pass a datetime then the Acquisition will be constructed on the data-transfer-service where there's no particular guarantee about the timezone. So I think pushing people to your "preferred" route is correct. 

We should also probably restrict the extractor models whenever possible to only accept datetimes with timezone information when this is a risk

bruno-f-cruz:
> All timestamps should be in ISO 8601 format with timezone offset representing local time. This allows for queries that consider local time when experiments were performed.

Most modern databases keep a UTC representation of all datetimes, when you query, you simply have to convert the query behind the scenes and return asset.

If you want to test this intuition:

```python
import pydantic
import datetime
from zoneinfo import ZoneInfo

class Foo(pydantic.BaseModel):
    gmt: pydantic.AwareDatetime
    pst: pydantic.AwareDatetime
    japan: pydantic.AwareDatetime
    
foo = Foo(
    gmt="2024-06-01T12:00:00+00:00",
    pst="2024-06-01T05:00:00-07:00",
    japan=datetime.datetime(2024, 6, 1, 21, 0, 0, tzinfo=ZoneInfo("Asia/Tokyo")),
)

print(foo)

print(foo.gmt == foo.pst == foo.japan)  # This will be True because both represent the same point in time
```

> But if a user provides timestamps in UTC, they'll get serialized as UTC, right? My understanding from what Saskia said to me was that we wanted to preserve actual wall time in the timezone the data was collected in since that's relevant biologically.

Inferring metadata (such as location, time of day, or lighting conditions) from timezone offsets is brittle and conceptually questionable. The offset tells you nothing about biological or environmental context — it only indicates the local legal time setting of the computer that recorded the data.

If knowing the sun angle is important, track GPS coordinates.
If knowing whether lights were on or off is important, track light state.
If knowing the timezone is important, store timezone name explicitly, e.g. using TimeZoneName
This approach aligns with the philosophy of aind-data-schema: make metadata explicit, not implicit.


>I just used the metadata mcp to do a little searching. Existing records seem to be all over the map in whether or not acquisition.session_start_time is in timezone aware and, if so, whether it's in local time with offset or UTC:

This is exactly why I have made the point in the past that the current format of the asset name should be refactored ASAP. We were close, unfurtunatly the change that would solve this issue did not go throught :P see https://github.com/AllenNeuralDynamics/aind-data-schema-models/issues/100 and https://github.com/AllenNeuralDynamics/aind-data-schema/discussions/1400

You will notice the vr-foraging examples you cite are consistent with UTC/offset since we ALWAYS use time-aware and UTC. The file name is actually generated automatically during the watchdog/data transfer, I have 0 control over it, but since Patrick just uses Iso8061 everything works out (long live standards!)

I know this sounds repetitive, but allowing non–timezone-aware datetimes leaves room for undefined and dangerous behavior. The world has agreed on how to represent time — we should follow those standards instead of inventing our own.

## Some notes:

- The point of datetime objects is to uniquely identify a point in time. In order for this to happen we must include a reference to UTC (either Z, or offset).
- Striping ANY datetime of offset information is a LOSSY transformation. That datetime must no longer be relied on for any queries! No only that, but it is also dangerous since you can actually go "back in time" during day light savings. This may appear unimportant, until you realize we run 24/7 experiments in the building (i.e. Pirouette)
- Adding timezone information using ZoneInfo is a bad idea if the goal is to keep track of locations. Pydantic serialization, as far as I know, for datetimes strips this information at the earliest chance. The example above serializes as:
`{"gmt":"2024-06-01T12:00:00Z","pst":"2024-06-01T05:00:00-07:00","japan":"2024-06-01T21:00:00+09:00"}`
- If you want to keep track of ZoneInfo, do it in a different field
- OS-level timezones are unreliable — users often travel without changing system settings. Relying on them produces the worst kind of metadata: wrong metadata.


## My take

Keep aind-data-schema exactly as it is, since it already follows ISO 8601 and supports UTC.
If anything, begin deprecating timezone-unaware datetimes by issuing warnings and encouraging use of timezone-aware fields.

Then, revisit asset naming conventions — Doug’s point about searchability is valid, but the fix should come from adopting a consistent standard, not changing the schema.

dougollerenshaw:
> Striping ANY datetime of offset information is a LOSSY transformation.

That's not being proposed here.

src/aind_data_schema/utils/docs/model_generator.py
```diff
@@ -14,7 +14,7 @@
 from aind_data_schema.utils.docs.utils import generate_enum_table
 
 special_cases = {
-    "pydantic.types.AwareDatetime": "datetime (timezone-aware)",
+    "pydantic.types.AwareDatetime": "datetime (timezone-aware, local time)",
```
dbirman: Small thing I just noticed, I think you have to run the doc generator commands (see contributing.md) to actually have this change the docs. It has to propagate this change into the source files.

dougollerenshaw: Should we wait until we're ready to merge the PR before doing that? I think we need someone to make a call on my options 1 or 2 above before we're ready to merge.

bruno-f-cruz:
@dougollerenshaw 
> 3. Using TZ-aware timestamps with offset for the local timezone avoids the need to store the local timezone in an additional metadata field, which would necessary to support local-time-based search queries (e.g. splitting sessions into morning/afternoon) if we were to use UTC datetimes.

Can you provide an example (I guess without using the name of the assets since it is out of the scope of this PR) of a query that doesn't work when we use UTC? I am sorry but I still dont understand the issue.

jtyoung84:
> Some clarifying points about this PR:
> 
>     1. This is separate from filename conventions. It is only meant to clarify how timestamps in the acquisition should be encoded.
> 
>     2. [This existing doc](https://github.com/AllenNeuralDynamics/aind-file-standards/blob/main/docs/core/core-standards.md#datetime) called for timezone aware timestamps in local time, which this PR accomplishes.
> 
>     3. Using TZ-aware timestamps with offset for the local timezone avoids the need to store the local timezone in an additional metadata field, which would necessary to support local-time-based search queries (e.g. splitting sessions into morning/afternoon) if we were to use UTC datetimes.
> 
> 
> However, the documentation this PR adds contradicts the datetime standards described [here](https://github.com/AllenNeuralDynamics/aind-file-standards/blob/main/docs/core/core-standards.md#datetime), which explicitly allow UTC timestamps.
> 
> So I think there are two possible paths forward (at least that I see):
> 
>     1. Merge this PR and also put in a PR to remove the UTC examples from [here](https://github.com/AllenNeuralDynamics/aind-file-standards/blob/main/docs/core/core-standards.md#datetime).
> 
>     2. Modify this PR to allow UTC timestamps in the acquisition.json, but clarify that users must then also provide the local timezone as a separate field, which would be necessary to allow local-time-based search queries. Presumably this would also require a PR to the schema to define a new optional `local_timezone` field.

My two cents:

A) For filenames that include timestamps, I'd strongly encourage people to adopt an ISO 8601 UTC representation like "2023-10-27T15:30:00Z". The benefits: they're sortable, no confusion during daylight savings time, and it's what most other people do. If someone really needs the timezone or location, then that's information they should get from the metadata files.

B) For the folders in S3, we're moving towards adopting GUIDs, so it's mostly how the `data_description.name` field should be generated. I can update aind-data-transfer-service to begin to require AwareDateTime or Timestamp objects. In the past, I've relied on this method:

https://github.com/AllenNeuralDynamics/aind-data-schema-models/blob/731a2cd141471c8945c66912923896e3aa52a4af/src/aind_data_schema_models/data_name_patterns.py#L99

which only required a naive datetime object.

C) For "how should AwareDateTime objects be JSON encoded," I don't think it should be that important? They get de-serialized to the same Timestamp regardless of the offset. If it really is important, then we need to add an additional field in the metadata and add a custom JSON serializer in aind-data-schema that will render the JSON representation appropriately.

jtyoung84:
> 1. Encode timestamps in local time w/ timezone offset (all necessary information in one place)

This is **extremely** risky as this information can get lost in serialization/de-serialization cycles and if the schema changes. If the timezone is important, it needs to be an explicit field in the metadata.

dougollerenshaw:
> > 1. Encode timestamps in local time w/ timezone offset (all necessary information in one place)
> 
> This is **extremely** risky as this information can get lost in serialization/de-serialization cycles and if the schema changes. If the timezone is important, it needs to be an explicit field in the metadata.

So a vote for option 2, then! @dbirman thoughts on the necessary schema changes to allow this? Would making it optional allow it to be backward compatible? Should there be a validator to ensure that it's not left empty if the acquisition times are in UTC?

jtyoung84:
> > Should we add a field for the location where the experiment was performed? That seems like a useful thing to track, and it would be simple to find the local timezone given a location and construct the local time with the UTC timestamp plus location.
> 
> This sounds like another good potential compromise!

It isn't as simple as it sounds. Time zones have changed over time. We would probably need to rely on a third-party library to convert locations to time zones.

dougollerenshaw:
> > > Should we add a field for the location where the experiment was performed? That seems like a useful thing to track, and it would be simple to find the local timezone given a location and construct the local time with the UTC timestamp plus location.
> > 
> > 
> > This sounds like another good potential compromise!
> 
> Do you mean something other than the Institution field?

I suppose that's already enough, but it seems indirect. Now to find all sessions run before noon local time, I'd need to first check the institution, then determine its timezone, then convert to local time for my query. Possible, but even more convoluted.

I'd still argue that option 1 is simplest.

dougollerenshaw:
@bruno-f-cruz I think we're talking about different things. I don't think it really matters where the experiment happened for the example I gave. What matters is what time it was _locally_ when it happened. I'm concerned about recovering local time _from_ UTC when timezone wasn't stored, which is currently happening. 

I think we just need a decision from @dyf or @saskiad about whether to purse options 1 or 2 [above](https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1604#issuecomment-3439072962) so we can move on.

jtyoung84:
Here is the Google AI answer to "monogdb query against timezone aware datetime"
```
MongoDB internally stores all Date objects as Coordinated Universal Time (UTC). When querying against timezone-aware datetimes, it is generally recommended to convert the query dates to UTC before sending them to MongoDB. This ensures that the comparison is performed against the stored UTC values, regardless of the client's local timezone.   
  
           
Here's how to approach querying with timezone-aware datetimes in MongoDB using PyMongo as an example:   
  
Ensure UTC Storage:
When inserting or updating documents, ensure that any datetime objects are converted to UTC before being stored in MongoDB. PyMongo automatically converts timezone-aware datetime objects to UTC when inserting them. If you're creating datetime objects manually, use datetime.now(tz=datetime.timezone.utc) or convert a local datetime to UTC using astimezone(pytz.utc).

Convert Query Dates to UTC:
When constructing a query, convert the datetime objects you are using for filtering to UTC.
```
we should be serializing datetime objects to the UTC representation and users can form reliable queries by converting their query dates to UTC. If timezone information is important, then that should be an extra field.

dbirman:
@dougollerenshaw I think we should close this PR and re-open as two issues to:

1. Convert all timezone-aware objects to UTC during serialization
2. *except* the acquisition_start/end_time fields, where we store the timezone separately

PR #1608, opened 2025-10-24
refactor: CurationMetric

dbirman:
This PR refactors `CurationMetric.value` to allow for the value field to change during repeated curations in a way that will not explode in size.

To do this, I run custom serialization on the field that compresses that value list[dict] to only maintain changes. During de-serialization the full list[dict] is recovered for easy access by users.

Unfortunately, this makes the curation value difficult to parse from the serialized JSON file.

### DIFF
--- a/tests/test_serialization.py
+++ b/tests/test_serialization.py
@@ -137,12 +137,12 @@ class SerializationTests(unittest.TestCase):
         self.assertEqual(result, data)
 
     def test_expand_list_of_dicts_delta_false_marker(self):
-        """Test expansion with _dc=False returns original"""
+        """Test expansion with _dc=False returns the list (unwrapped)"""
         data = {"_dc": False, "_v": [{"a": 1}, {"b": 2}]}
 
         result = expand_list_of_dicts_delta(data)
 
-        self.assertEqual(result, data)
+        self.assertEqual(result, [{"a": 1}, {"b": 2}])
 
     def test_expand_list_of_dicts_delta_carry_forward(self):
         """Test that values are carried forward correctly"""


src/aind_data_schema/core/quality_control.py
```diff
@@ -89,20 +89,28 @@ class CurationHistory(DataModel):
...
 
     value: List[Any] = Field(..., title="Curation value")
     type: str = Field(..., title="Curation type")
     curation_history: List[CurationHistory] = Field(default=[], title="Curation history")
 
 
+class ElementCurationMetric(QCMetric):
+    """Curation metric for individual elements within a data asset"""
+
+    value: dict[str, List[Any]] = Field(..., title="Curation value per element")
```
saskiad: Mostly thinking out loud .... what's the reason we can't have this structure be the only one? dict[List] When someone's doing ephys unit curation the list would include all the units, and there'd only be one key. When some is doing element-wise curation, then the keys are all the different cells...  I feel like I'm forgetting something about this.

dbirman: What you're forgetting is that we track the history. So the outer dictionary maps {"neuron1": list[curations for neuron1]}. And then in curation_history you have to track the history of each of those changes. It's a lot of meta-metadata but for neuron-level metadata I think this is unavoidable

dbirman:
New plan for this:
    Change the type of value in CurationMetric to match what we did with ElementcurationMetric
    Add a before validator that converts the old style to the new one 
    Communicate this out before we merge the PR

dbirman:
This will go in v2.4, too many changes to QC to put it in v2.3

docs/source/quality_control.md
```diff
@@ -20,7 +20,12 @@ Each [QCMetric](#qcmetric) is annotated with three pieces of additional metadata
...
+
+The difference between the two options is that:
+
+- [CurationMetric](#curationmetric) expects a single dictionary to describe the curation and any update to the curation requires writing a new copy of the entire dictionary. This makes `CurationMetric` best suited to curations that act on multiple elements of the data asset at once, for example: splitting and combining clusters after spike sorting.
+- [ElementCurationMetric](#elementcurationmetric) should be used when each element of a data asset has its own curation and these can change over time. Instead of accepting just one dictionary, the `ElementCurationMetric.value` wraps each element's curation data in an outer dictionary that is split by elements. This makes updating a single element's curation more efficient.
```
saskiad: would be good to add an example here paralleling the examples above. e.g. annotation of individual neuron reconstruction...

docs/base/core/quality_control.md
```diff
@@ -20,7 +20,12 @@ Each [QCMetric](#qcmetric) is annotated with three pieces of additional metadata
 
 ### Curations
 
-If you find yourself computing a value for something smaller than an entire modality of data in an asset you are performing *curation*, i.e. you are determining the status of a subset of a modality in the data asset. We provide the [CurationMetric](#curationmetric) for this purpose. You should put a dictionary in the `CurationMetric.value` field that contains a mapping between the subsets (usually neurons, ROIs, channels, etc) and their values.
+If you find yourself computing a value for something smaller than an entire modality of data in an asset you are performing *curation*, i.e. you are determining the status of a subset of a modality in the data asset. We provide the [CurationMetric](#curationmetric) and [ElementCurationMetric](#elementcurationmetric) models for this purpose. You should put a dictionary in the `CurationMetric.value` field that contains a mapping between the subsets (usually neurons, ROIs, channels, etc) and their values.
```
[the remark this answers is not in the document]
dbirman: Right, I guess a better way to say this is that QC is relative to the data you acquired -- if you acquired images then you QC the images if you acquired ephys data you QC the ephys data, same for processing if you run spikesorting you QC the entire spikesorting. But when you start asking questions about individual ROIs inside of the image you're now doing curation.

I think I'll also call out that this is a bit of a grey area and curations are partly to keep the # of QC metrics manageable + human readable.

PR #1671, opened 2026-01-05
fix: reduce requirement for SLAP2 to just ImagingConfig

dbirman:
PR fixes a bug where including Modality.SLAP2 incorrectly required users to provide a standalone SlapPlane outside of an ImagingConfig.


saskiad:
wait, hold on. we want this to use slapplane. That's the whole reason it's there.

PR #1672, opened 2026-01-05
add GenericModelType to fix serialization

tmchartrand:
adds GenericModelType to fix serialization of GenericModel fields
also sets all GenericModelType fields to Optional with default=None


tmchartrand:
the existing test here actually covers it, would be failing with the new test (removed model_dump call) and the old GenericModel: https://github.com/AllenNeuralDynamics/aind-data-schema/blob/a3818073a7f44288ec8e540d0421ca0b7a7d51b0/tests/test_aind_generic.py#L53-L65

PR #1705, opened 2026-01-30
refactor: allow instrument_id to merge in alphabetical order

dbirman:


dougollerenshaw:
I'm late to the game here, but I'm afraid that `merge_str_alphabetical` being limited to two strings will cause problems down the road. There is an identified use case of combining behavior, fip and ephys, so presumably we'll be combining three instrument files in that case.

dbirman:
> I'm late to the game here, but I'm afraid that `merge_str_alphabetical` being limited to two strings will cause problems down the road. There is an identified use case of combining behavior, fip and ephys, so presumably we'll be combining three instrument files in that case.

I lean towards having users merge their 3+ instruments ahead of time and asking them to ensure they do it in alphabetical order.

The alternative is to split the name on the _ character and re-order the list of instrument_id values during the merge. This means we would also need to validate that the instrument_id doesn't contain a _ ahead of time -- which would probably mean doing custom serialization/de-serialization from a list[str1, str2, str3] -> str1_str2_str3 etc. Basically kind of a headache...

dougollerenshaw:
> > I'm late to the game here, but I'm afraid that `merge_str_alphabetical` being limited to two strings will cause problems down the road. There is an identified use case of combining behavior, fip and ephys, so presumably we'll be combining three instrument files in that case.
> 
> I lean towards having users merge their 3+ instruments ahead of time and asking them to ensure they do it in alphabetical order.
> 

I really don't like that. I think we should have consistent rules. If users can merge instruments, they should be able to merge 2 or 3.

> The alternative is to split the name on the _ character and re-order the list of instrument_id values during the merge. This means we would also need to validate that the instrument_id doesn't contain a _ ahead of time -- which would probably mean doing custom serialization/de-serialization from a list[str1, str2, str3] -> str1_str2_str3 etc. Basically kind of a headache...

I don't quite follow this. The instrument names are separate to start with. Why not just cast to a list, alphabetize, then join? I'd have voted for joining on a `"+"` character if I'd reviewed this PR, but merging on `"_"` seems fine, even if some names already have `"_"` in the name. We don't need to split later, do we?

dbirman:
Merging is implemented by overloading the __add__ operator in Python. You can't add three things at once, so by definition add will happen two times with three instruments, hence the ordering issue. I don't see an easy way to get around this other than storing the individual names split in the Instrument objects. From a maintainability standpoint it's not great, and from a user standpoint it will be confusing to see two different things if they look at the object or if they look at the JSON. Not a disaster, but not great.

PR #1709, opened 2026-01-31
added crop to detector config

### LINKED ISSUE #1704: Add crop settings to camera/detector config
Crop settings (x,y,width,height) are optional fields in the camera/detector device. For HSFP and FIB, these can be changed more frequently. We should add these to the config class, still as optional

saskiad:
closes #1704


dbirman (commented):
Before approving I want to check if we need to remove this from the Instrument? If we don't remove it from the device properties then we run into the risk that someone puts values in both the device and in the configuration. Presumably the "rule" is that the configuration overrides the device properties but we don't have much guarantee of that. I lean toward moving things into the acquisition if they are variable like this. What do you think?

saskiad:
I think we discussed this the other week, but want to make sure. We agreed that acquisition overrides instrument when both are provided. I don't think we decided to remove it from the instrument though, did we? I'd like to finish this up if we can.

dbirman:
> I think we discussed this the other week, but want to make sure. We agreed that acquisition overrides instrument when both are provided. I don't think we decided to remove it from the instrument though, did we? I'd like to finish this up if we can.

Yes we discussed and decided to put it in the acquisition and it's just something you have to know for now, to avoid a breaking change.

For 3.0 we will go through all devices and remove anything that is "configuration" and move it to the acquisition.

dbirman (approved):
Bunch of linter changes snuck in but the actual content here looks correct to me!

Issue #1527, opened 2025-07-28
Rounding error


saskiad:
Originally noted by Mathew for U19 Smartspim data:

In this case, observed for specimen ID# 692909, injection_coordinate_ap should be 1.6 but is showing as 1.59999999

I'm not sure if this is an issue with Pydantic or something with service, but let's redirect this as needed?

dbirman:
Assuming this is v1.4 and a BrainInjection object the pydantic definition was using Decimal `injection_coordinate_ap: Decimal = Field(..., title="Injection coordinate AP (mm)")`

So that suggests the error came upstream, since the Decimal class shouldn't introduce floating point / rounding errors like this. We should look at the metadata-service and make sure it didn't come from there and check how it's stored in the actual database.

dbirman:
I don't know how this could happen since in v1 that field has always been a Decimal. I think this will remain a mystery...

bruno-f-cruz:
Another possibility is that people tried to use the type hint provided:

```python
from pydantic import BaseModel
from decimal import Decimal
class MyModel(BaseModel):
    value: Decimal
    value_float: float
    

print(MyModel(value="0.42", value_float=0.42).model_dump_json())
print(MyModel(value=0.42, value_float=0.42).model_dump_json())
print(MyModel(value=Decimal("0.42"), value_float=0.42).model_dump_json()) 
print(MyModel(value=Decimal(0.42), value_float=0.42).model_dump_json())
```

prints:

```cmd
{"value":"0.42","value_float":0.42}
{"value":"0.42","value_float":0.42}
{"value":"0.42","value_float":0.42}
{"value":"0.419999999999999984456877655247808434069156646728515625","value_float":0.42}
```

The pydantic default behavior is to coerce. In this case, it will try to:  `Decimal(str(value))`, as a result if people pass the Decimal(float) directly, you will get what you expect (the representation of the float with the imprecision).

dbirman:
In V2 we already removed Decimal in a bunch of places where it wasn't critical anymore to avoid this issue. I don't think there's anything actionable here other than ensuring users understand that you need to pass strings to the `Decimal` constructor.

PR #1710, opened 2026-02-01
Slap2 Metadata Updates

### LINKED ISSUE #1422: SLAP2 metadata requirements
SLAP2 microscope modes

ROI types AcquisitionContainer.ROIs.rois.imagingMode <- pull up to a top-level list
.isSimpleRaster <- sortable field

Drop ParsePlan (keep in data)

Image coordinate system origin needs to support references relative to another contextual image. Both for images of an FOV (image registration, preferred) or stage coordinates (potentially flawed).

Procedures need to capture viral injection
Imaging target: controlled vocabulary for cortical cell types? Maybe not possible right now. Set of options: "broad tufted", "layer 2", "broad tufted + layer 3" to allow people to mix possible targets.

Metadata mapper needs to deal with pulling in the asset paths for the images as well as copy the field of view image from somewhere into the data asset.

dbirman:
This PR modifies the slap2 metadata requirements following discussions with the scientific team. 

- Adds 3 new example files from @m-xie -- these are probably the best place to start reviewing from
- Support for analog detectors (SiPM)
- New Slap2Plane uses paths to metadata files that include information about framerates, dilation, etc
- DMD simplified

A few linter/docs changes snuck in that didn't get built properly in other PRs


src/aind_data_schema/components/configs.py
```diff
@@ -222,19 +208,29 @@ class CoupledPlane(Plane):
...
+
+    mask_image_path: AssetPath = Field(
+        ..., title="Mask image path",
+        description="Relative path from metadata json to imaging ROI masks (numbered by superpixel)"
+    )
 
-    dmd_dilation_x: int = Field(..., title="DMD Dilation X (pixels)")
-    dmd_dilation_y: int = Field(..., title="DMD Dilation Y (pixels)")
+    unique_y_dilations: List[int] = Field(..., title="Unique Y dilations")
+    y_dilation_image_path: AssetPath = Field(
```
saskiad: I'm not really a fan of this. 
We don't list the condition of each stimulus trial in our metadata. That's data. I kind of think of this in the same way. I the dilations are needed to analyze the scan path, that's data in the asset, but I don't think it needs to be pointed to with a specific field here.

src/aind_data_schema/components/configs.py
```diff
@@ -222,19 +208,29 @@ class CoupledPlane(Plane):
...
+    target_name: Optional[str] = Field(default=None, title="Name of imaged target")
+
+    mask_image_path: AssetPath = Field(
+        ..., title="Mask image path",
+        description="Relative path from metadata json to imaging ROI masks (numbered by superpixel)"
+    )
 
-    dmd_dilation_x: int = Field(..., title="DMD Dilation X (pixels)")
-    dmd_dilation_y: int = Field(..., title="DMD Dilation Y (pixels)")
+    unique_y_dilations: List[int] = Field(..., title="Unique Y dilations")
```
saskiad: is this important metadata? I'm fine if it is, but just want to make sure we're not including it because we had added the dmd_dilations the first time around and feel like we need to keep the field. Is this used to identify data or understand data?

src/aind_data_schema/components/configs.py
```diff
@@ -222,19 +208,29 @@ class CoupledPlane(Plane):
...
+    )
+    x_dilation: int = Field(..., title="X dilation")
     dilation_unit: SizeUnit = Field(default=SizeUnit.PX, title="Dilation unit")
 
-    slap_acquisition_type: SlapAcquisitionType = Field(..., title="Slap experiment type")
-    target_neuron: Optional[str] = Field(default=None, title="Target neuron")
-    target_branch: Optional[str] = Field(default=None, title="Target branch")
-    path_to_array_of_frame_rates: AssetPath = Field(
-        ..., title="Array of frame rates", description="Relative path from metadata json to file"
+    unique_frame_rates: List[float] = Field(..., title="Unique frame rates")
```
saskiad: same comments for this as above... is this a useful list and do we need the path to the data?

src/aind_data_schema/components/devices.py
```diff
@@ -499,19 +499,15 @@ class DigitalMicromirrorDevice(Device):
...
     padding_unit: SizeUnit = Field(default=SizeUnit.PX, title="Padding unit")
-    pixel_size: Decimal = Field(..., title="DMD Pixel size")
-    pixel_size_unit: SizeUnit = Field(default=SizeUnit.UM, title="Pixel size unit")
-    start_phase: Decimal = Field(..., title="DMD Start phase (fraction of cycle)")
-    dmd_flip: bool = Field(..., title="DMD Flip")
-    dmd_curtain: List[Decimal] = Field(..., title="DMD Curtain")
-    dmd_curtain_unit: SizeUnit = Field(default=SizeUnit.PX, title="dmd_curtain_unit")
-    line_shear: List[int] = Field(..., title="Line shear (pixels)")
-    line_shear_unit: SizeUnit = Field(default=SizeUnit.PX, title="Line shear unit")
+    pixel_size: float = Field(..., title="DMD Pixel size (fraction of line scan period)")
```
saskiad, line 506: there was a reason we made everything Decimal instead of float - are you sure you want float here? It was a pain switching everything the first time.

src/aind_data_schema/components/configs.py
```diff
@@ -222,19 +208,29 @@ class CoupledPlane(Plane):
     power_ratio: float = Field(..., title="Power ratio")
 
 
-class SlapPlane(Plane):
-    """Configuration of an imagine plane on a Slap microscope"""
+class Slap2Plane(Plane):
+    """Configuration of a SLAP2 imaging plane (all imaging ROIs of a specific acquisition type at a particular depth)"""
+
+    slap2_acquisition_type: Slap2AcquisitionType = Field(..., title="SLAP2 ROI acquisition type")
+    target_name: Optional[str] = Field(default=None, title="Name of imaged target")
```
saskiad: what does this mean? is this targeted structure? which would then be using the atlas. Or is this an identifier for the neuron? Or something else?

src/aind_data_schema/components/configs.py
```diff
@@ -222,19 +208,29 @@ class CoupledPlane(Plane):
...
+
+    mask_image_path: AssetPath = Field(
```
saskiad: same as other comments.

examples/slap2_acquisition.py
```diff
@@ -0,0 +1,425 @@
...
+)
+from aind_data_schema_models.brain_atlas import CCFv3
+from aind_data_schema_models.stimulus_modality import StimulusModality
+
+origin_ari = CoordinateSystemLibrary.BREGMA_ARI
+origin_ari.name = "Arbitrary Origin ARI"
+origin_ari.axis_unit = SizeUnit.UM
+origin_ari.origin = Origin.ORIGIN
+
+filter_red = Filter(
```
saskiad: why are there filters in the acquisition file?

examples/slap2_acquisition.py
```diff
@@ -0,0 +1,425 @@
...
+            code=Code(
+                url=(
+                    "https://github.com/AllenNeuralDynamics/ophys-passive-visual-stim/"
+                    "blob/main/src/RandomDriftingGratings_ContinuousTrials.bonsai"
+                ),
+                version="168e4ef1923c535f6c4d914a126526cc11168ac7",  # get git commit id
+                name="RandomDriftingGratings_ContinuousTrials.bonsai",
+                language="Bonsai",
+                language_version="2.9.0",  # get from repo or code directly
+                parameters={
```
saskiad, line 376: I would expect information on the sf, tf, contrast, and direction of the gratings here.
I do not expect portname, screen bluecolor (etc) and definitely not subject. 
There shouldn't be a reference to a parameters file, the parameters shoudl be here.

saskiad: I'd also expect trial duration and intertrial duration. 
I don't know what initial spont int is and I don't know what any of these units are

src/aind_data_schema/components/configs.py
saskiad: I also don't think I'd call it "unique_xxx" - To me the list implies this is unique values, and that's how we use lists elsewhere.

src/aind_data_schema/components/devices.py
dbirman: We need to avoid Decimal unless people actually understand how it works (you have to pass a string to create a Decimal object, not a float). If someone doesn't need the level of precision afforded by Decimal then float is better, see rounding error created by passing floats to Decimal: https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1527

examples/slap2_acquisition.py
m-xie: In my mind, since these are part of the Code object, the object structure implies (to me as a new aind-data-schema user) that what goes here should be the exact parameters that are passed to the Bonsai code that runs the stimulus. Our code is parameterized in this way, which is why these parameters are listed here. Since our code/parameters csv are in a public repo, with the code, commit hash, and parameters, anyone has all the information needed to exactly define the stimulus present.

We also want our structure here to be relatively generalizable to different kinds of visual stimuli that we are running now (movies, random vs non random gratings, RF mapping, etc.), so it seemed like this way of defining the stimulus worked well for that. I am happy to discuss more though and clarify if I'm misunderstanding anything.

src/aind_data_schema/components/configs.py
m-xie: `target_name` is an identifier for the specific neuron/FOV being imaged. It's one level below the `targeted_structure` which is the brain region (i.e. `target_name` provides a nickname for what within the brain region is being imaged and allows for efficiently linking of different sessions where we intend to image the same neurons/dendrites/axons/other structures). Happy to workshop or modify the naming of the field though

examples/slap2_acquisition.py
dbirman: They are copied from the slap2_instrument and used to construct the ImagingConfig...

m-xie: For these acquisition json examples, I can remove the filters and just hardcode the typical channel wavelength values, if that is better for the purpose of examples?

When we do our extraction and mapping of SLAP2 data into the aind-data-schema format, I read in the instrument.json when creating the acquisition.json to make sure that the channel wavelengths are consistent with the filters on the instrument, but in the example we don't need to do this.

src/aind_data_schema/components/configs.py
m-xie: I'll respond to a few of the related comments here, but let me know if I've missed anything!

Basically, there are three Slap2Plane fields that reference a Path to an image:

1. mask_image: With the SLAP2, not every pixel within the 800x1280 pixel field of view of a plane. This information represents the pixels that are actually imaged in a particular plane (and each plane may have a different set of pixels imaged).
2. frame_rate: Due to the scanning mechanism of the SLAP2 each imaged pixel is not necessarily imaged at the same rate (and importantly this can vary within a plane). I included this because I think it is interesting for users to easily access both the set of frame rates that are used for each plane, and which pixels are imaged at each frame rate.
3. y_dilation: The y dilation is a unique concept to the SLAP2 scanning mechanism that essentially affects the laser power of the scanned focus and the shape of the scanned focus (i.e. amplitude and spread of the point spread function used to imaged a specific pixel). Again, this could vary within a plane. Similar to frame_rate, I thought users would be interested in both the set of y_dilations used and where each y_dilation is used to have a sense of the laser power delivered and the imaging resolution of the parts of the image.

All this data is available in the metadata that comes off the microscope and is saved in the data asset too, and I'm not sure of the best data vs. metadata conventions, so I will defer to your judgement on which of these should be fields in the aind metadata. If anything is unclear or if any additional description would be helpful, I can provide those as well!

src/aind_data_schema/components/configs.py
m-xie: I agree that we could get rid of the "unique_xxx" prefix for y_dilations and frame_rate if we decide to keep those in. Looking back they are a bit redundant!

src/aind_data_schema/components/configs.py
m-xie: addressed above: https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1710#discussion_r2855960298

src/aind_data_schema/components/configs.py
m-xie: addressed above: https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1710#discussion_r2855960298

examples/slap2_acquisition.py
m-xie: Made changes to include the grating parameters here: https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1751

src/aind_data_schema/components/configs.py
saskiad: I do think we want to handle this differently
We're currently integrating AIBS patchseq modality, where the recorded neurons are identified using a specimen_id, and I think we want the same approach here so that we aren't doing the same concept in different ways.
Would different planes target different neurons? I thought they'd target the same neuron?
And if part of the intention is to describe axon vs soma vs dendrite, then I think we need an additional field for that. If all planes are the same neuron, I'd suggest the specimen id lives up a level and the structure stay in the plane (with a controlled vocabulary)

m-xie: Currently, there are two different types of experiments that we run that might need this:

1. The experimenter is aiming to image as many dendrites as possible on an individual neuron and they want to revisit that neuron across multiple days. Currently, if we revisit that neuron, we are aiming to revisit the same set of dendrites. (In the future, there likely will be cases where we want to revisit the same neuron but look at a different subset of dendrites on that neuron). The intended `target_name` in this case would be something like "Neuron1". If we revisit different dendrite subsets on that neuron, the `target_name` could be something like "Neuron1-FOV1"
2. The experimenter is aiming to image a bunch of dendrites in a field of view (but they don't know what neurons they belong to) and they want to revisit those exact set of dendrites across multiple days. This is more similar to what the BCI team is doing except they are looking at cell bodies instead of dendrites. The intended `target_name` in this case would be something like "FOV1"

Also, we don't have many experiments that do this now, but it is possible that in a single experiment you measure two different neurons on different planes (maybe a L2/3 neuron and a L5 neuron). I'm open to discussing what is the best way to handle this information. What is important for us is having a field that allows us to easily match data assets that image the same thing, similar to how the BCI team uses the FOV1,2,etc. labels.

src/aind_data_schema/components/devices.py
saskiad: yep, this was me forgetting things. Sorry about that. Fully on board.

src/aind_data_schema/components/configs.py
dbirman: After discussion we're going to have:

- specimen_id: Optional[str], description: "Unique index identifying the cell being imaged <subject_id>_###"
- fov_index: Optional[int], description: "Index for matching similar fields of view across sessions"
- structure_type: Enum (dendrite, soma, axon, other, more stuff?)

fov_index will be required when you provide a specimen_id, to avoid a situation where a specimen_id accidentally gets re-used across FOVs. 

Does that cover all the situations we're thinking about?

m-xie: Yeah I think that overall format should work! Specifically though, can structure_type be a list? We often image both some dendrites and the soma in a plane, so would be great to be able to have `structure_type=["dendrite","soma"]`

m-xie: I also think for now, dendrite, soma, axon will cover most cases for in vivo imaging. I'll update if I think of any others down the line.

dbirman: Thanks I'll make this change and then lets see what the examples look like with it in place.

src/aind_data_schema/components/configs.py
dbirman: After discussion we're going to reduce to `frame_rates`, `y_dilations` and remove paths (which will go in file standards, when those can be made)

src/aind_data_schema/components/configs.py
dbirman: Changed

src/aind_data_schema/components/configs.py
dbirman: Changed

m-xie (approved):

src/aind_data_schema/components/configs.py
```diff
@@ -75,6 +69,7 @@ class TriggerType(str, Enum):
 
     INTERNAL = "Internal"
     EXTERNAL = "External"
+    ANALOG = "Analog"
```
saskiad, line 72: Does this assume that internal/external triggers are digital? To me analog/digital is separate from internal/external...

examples/slap2_acquisition.py
```diff
@@ -0,0 +1,423 @@
...
+            y_dilations_unit=SizeUnit.PX,
+        ),
+        Slap2Plane(  # each fastZ plane will get its own Slap2Plane
+            depth=plane_depths["Path 1"][1],
+            depth_unit=SizeUnit.UM,
+            power=hwp_laser_power,
+            power_unit=PowerUnit.PERCENT,
+            targeted_structure=CCFv3.VISPL2_3,
+            specimen_id=f"{mouse_id}_{imaging_target_name}",
+            fov_index=1,
```
saskiad: are these two planes the same fov_index? Shouldn't they be separate fov?

examples/slap2_acquisition.py
```diff
@@ -0,0 +1,423 @@
...
+project_name = "SLAP2"
+acquisition_type = "multi-roi raster drifting gratings"
+
+mouse_id = "12345"
+
+num_paths = 2
+active_channels = ["Red", "Green"]
+
+plane_depths = {"Path 1": [184, 160], "Path 2": [100, 105]}
+fov_indices = {"Path 1": [1, 1], "Path 2": [1, 1]}
```
saskiad: see question below - I'm confused why these all have the same fov_index. Should they all be unique?

examples/slap2_acquisition.py
```diff
@@ -0,0 +1,423 @@
...
+                        "Num Trials": 40,
+                        "PortName": "COM7",
+                        "Screen_BlueColor (0-1)": 1,
+                        "Screen_GreenColor (0-1)": 0,
+                        "Screen_RedColor (0-1)": 0,
+                        "Subject": "000000",
+                        "GratingTrialParametersFile": ("ParameterFiles/8_direction_drifting_grating_params.csv"),
+                    },
+                    "GratingTrialParameters": {
+                        "Contrast": [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 0.0],
```
saskiad, line 388: these lists of parameters should only be the unique values of the parameters. If the contrast is always 1, then just list 1.

examples/slap2_acquisition.py
```diff
@@ -0,0 +1,423 @@
...
+                        "DiameterUnit": "degree",
+                        "Duration": [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0],
+                        "DurationUnit": "second",
+                        "Orientation": [0.0, 45.0, 90.0, 135.0, 180.0, 225.0, 270.0, 315.0, 359.0],
+                        "OrientationUnit": "degree",
+                        "SpatialFrequency": [0.02, 0.02, 0.02, 0.02, 0.02, 0.02, 0.02, 0.02, 0.02],
+                        "SpatialFrequencyUnit": "cycle/degree",
+                        "TemporalFrequency": [2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0, 2.0],
+                        "TemporalFrequencyUnit": "Hz",
+                        "X": [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
```
saskiad, line 401: is this x position?

examples/slap2_acquisition.py
```diff
@@ -0,0 +1,423 @@
...
+                        "Screen_GreenColor (0-1)": 0,
+                        "Screen_RedColor (0-1)": 0,
+                        "Subject": "000000",
+                        "GratingTrialParametersFile": ("ParameterFiles/8_direction_drifting_grating_params.csv"),
+                    },
+                    "GratingTrialParameters": {
+                        "Contrast": [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 0.0],
+                        "Delay": [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0],
+                        "DelayUnit": "second",
+                        "Diameter": [360.0, 360.0, 360.0, 360.0, 360.0, 360.0, 360.0, 360.0, 360.0],
```
saskiad, line 391: 360 degree diameter? That can't possibly be true.

saskiad (commented):
you can simplify the stim parameters. (the 360° is physically impossible, so maybe fix that)
my only real question is about the fov index. I thought this would be unique for each plane. is that not the idea here?
otherwise this looks really nice!

src/aind_data_schema/components/configs.py
dbirman: A triggered detector in my mind has to be digital, a PMT has no concept of "trigger" the signal is digitized downstream by some kind of converter? Unless there's some lens in front of it?

examples/slap2_acquisition.py
dbirman: I suspect these are full-field and that's what the X/Y (0,0) and 360 deg diameter mean. @m-xie can you confirm?

src/aind_data_schema/components/configs.py
saskiad: ah, I see, it's basically say N/A? 
I think we need to remove the trigger field because basically nobody knows what it means or how their device works. I think it was something someone asked for early on that wasn't fully vetted. But that's a 3.0 issue.

dbirman: Yeah it's a way of saying N/A in a way that felt more clear to me, but maybe not more clear to other people. I agree we can probably remove this field later, I opened a ticket.

m-xie: For now, if you think Analog would cause confusion, we could use External instead? The device itself is not triggered but the downstream digitization is based on the scanning clock that is "External" to the detector. But I agree with everything in the above discussion, so good with whatever you think is best

examples/slap2_acquisition.py
m-xie: There are 9 sets of grating parameters for the 9 stimuli that get randomized in each trial. One of the stimuli is a grey screen, which is shown here with the 0.0 contrast. I thought including the full set of parameters for each of the 9 possible stimuli was the clearest way to not create confusion of which unique parameter values go together. I'm essentially, recreating the CSV of stimulus parameters here: [8_direction_drifting_grating_params.csv](https://github.com/user-attachments/files/25754441/8_direction_drifting_grating_params.csv)

examples/slap2_acquisition.py
m-xie: Yeah these are the parameter names of the [BonVision GratingTrial](https://github.com/bonvision/BonVision/blob/main/BonVision/Collections/CreateGratingTrial.cs) and the values we pass in. These parameters are for full-field drifting gratings.

examples/slap2_acquisition.py
m-xie: Yeah, this is the ["position of the gratings along the x-axis, in degrees."](https://github.com/bonvision/BonVision/blob/785fa61d355c2a2fadc441d96012767273260db9/BonVision/Collections/CreateGratingTrial.cs#L59)

examples/slap2_acquisition.py
dbirman: Resolved

examples/slap2_acquisition.py
dbirman: Resolved in teams discussion

saskiad (approved):

PR #1718, opened 2026-02-03
feat: add MISCameraConfig

dbirman:
PR adds a new assembly configuration combining a DetectorConfig and a MISModuleConfig, for mounted cameras on the New Scale MIS system.

### DIFF
--- a/docs/source/acquisition.md
+++ b/docs/source/acquisition.md
@@ -79,7 +79,8 @@ while the StimulusEpoch represents all stimuli being presented.
 |-------|------|-------------|
 | `subject_id` | `str` | Subject ID (Unique identifier for the subject) |
 | `specimen_id` | `Optional[str]` | Specimen ID (Specimen ID is required for in vitro imaging modalities) |
-| `acquisition_start_time` | `datetime (timezone-aware)` | Acquisition start time  |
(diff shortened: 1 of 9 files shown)


src/aind_data_schema/components/configs.py
```diff
@@ -505,6 +505,13 @@ class EphysAssemblyConfig(DeviceConfig):
     )
 
 
+class MISCameraConfig(DeviceConfig):
+    """Configuration for a camera used in a New Scale modular insertion system"""
+
+    detector_config: DetectorConfig = Field(..., title="Detector configuration")
+    module: MISModuleConfig = Field(..., title="Module")
```
saskiad: doesn't this need the `ManipulatorConfig`?

saskiad (changes requested):
see comment

src/aind_data_schema/components/configs.py
jsiegle: No, there's no manipulator on these modules. Just a camera+lens on an arc with A/P and M/L angles.

dbirman: I added the DevicePosition mix-in @saskiad

PR #1720, opened 2026-02-03
fix: forgot to implement Acquisition.instrument_id merge

dbirman:
This PR adds the same instrument_id merge logic that we added to Instrument to Acquisition where we forgot to implement it. See https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1705


saskiad:
I'm a bit confused why we need this. For instrument I get it - we're combining instruments, they have independent instrument metadata. We merge them to make the new metadata. 
I don't understand the need to merge instrument ids in the acquisition - that should have already been done.

PR #1738, opened 2026-02-18
nhp subject

### LINKED ISSUE #1701: Add non-human-primate subject model
fields include:

- species
- sex
- year-of-birth (unless people think we might get date of birth)
- source
- mating_status (or a better name for this?), optional

anything else @gouwens

saskiad:
closes #1701


saskiad:
Two suggestions:
- move restrictions to the subject class (instead of in the mouse subject) (or remove from subject altogether and only use the one in data description)
- add mating status to mouse subject as optional field

src/aind_data_schema/components/subjects.py
```diff
@@ -148,8 +156,24 @@ def validate_species_strain(self):
...
+
+
+class NonHumanPrimateSubject(DataModel):
+    """Description of a non-human primate subject"""
+
+    species: Species.ONE_OF = Field(..., title="species")
     sex: Sex = Field(..., title="Sex")
+    date_of_birth: Optional[Annotated[date_type, TimeValidation.BEFORE]] = Field(default=None, title="Date of birth")
     year_of_birth: int = Field(..., title="Year of birth")
+    mating_status: MatingStatus = Field(..., title="Mating status")
```
gouwens, line 183: I'm not sure, but maybe mating status should also be optional for NHP subject?

gouwens (commented):
Overall looks good - just a question about mating status being required or optional.

src/aind_data_schema/components/subjects.py
saskiad: we do have an "unknown" option, which I think is a bit better than having it optional because then people are explicit about not knowing vs having to guess did they just not bother to provide it or is it unknown? 
But we can make it optional too...

gouwens: Got it - having "unknown" works for me, then

src/aind_data_schema/components/subjects.py
```diff
@@ -148,8 +156,24 @@ def validate_species_strain(self):
 class HumanSubject(DataModel):
     """Description of a human subject"""
 
+    species: Species.ONE_OF = Field(..., title="Species")
```
dbirman: I feel like there's a sensible default we could put here. More seriously this is a breaking change, we should add a field_validator in mode="before" that pre-fills this if it's not present to avoid crashing old assets (if any have HumanSubject, I'm not sure they do though)

dbirman (changes requested):
Need to add a validator for HumanSubject.species to avoid crashing old code (unless we're sure none of these assets exist)

src/aind_data_schema/components/subjects.py
saskiad: It gave me all sorts of errors when I tried to make it a default and I didn't have the patience to figure out why. But yes, I agree. 
I think there is one very old exaspim asset with human data, that may or may not have any of the rest of its metadata...I'll look into it. But yes, let's not break everything.

dbirman (approved):
Looks good

PR #1739, opened 2026-02-19
added IcEphysAssembly

### LINKED ISSUE #1725: icephys - make an assembly in devices for a manipulator + headstage
the key functional device is a headstage that is controlled by a manipulator. Each are just device fields.

the name here is important - we need to prevent users from getting icephys and ecephys confused, ideally without making everything icephys_probe and ecephys_probe...

saskiad:
closes #1725


saskiad:
I feel like there's an opportunity to consolidate these assembly concepts between EphysAssembly and IcEphysAssembly and FiberAssembly in v3.0

saskiad:
also, weight in on the name "IcEphysAssembly" - I don't love it but don't have better ideas

docs/source/components/devices.md
```diff
@@ -376,6 +376,17 @@ Module for inserted fiber photometry recording
 | `fibers` | List[[FiberProbe](#fiberprobe)] | Probes that are held by this module  |
 
 
+### IcEphysAssembly
+
+Assembly combining a manipulator and headstage used for Intracellular ephys
+
+| Field | Type | Title (Description) |
+|-------|------|-------------|
+| `name` | `str` | Fiber assembly name  |
```
gouwens: Is "fiber" supposed to be something different here?

src/aind_data_schema/components/devices.py
```diff
@@ -487,6 +487,14 @@ class FiberAssembly(DataModel):
     fibers: List[FiberProbe] = Field(..., title="Probes that are held by this module")
 
 
+class IcEphysAssembly(DataModel):
```
gouwens: Is `PatchClampEphysAssembly` too specific? I don't think we're probably going to be handling a lot of sharp electrode recording data, but maybe it still should be named in a more general way

saskiad: yeah, I like that better

docs/source/components/devices.md
saskiad: nope, that's copy and paste error on my part. Thanks!

dbirman (changes requested):
PatchClampEphysAssembly needs to also be added to the DiscriminatedList for instrument.components

dbirman (approved):
Looks good

gouwens (approved):

PR #1741, opened 2026-02-23
Remove ref to metadata app in instrument.md

dougollerenshaw:
Removes the reference to the metadata app in instrument.md. It sounds like no one is using it and we'd rather push people to create instrument JSONs from reusable python scripts. Also, the app seems to be AIND-specific (is it even available off-prem?) and documentation should be more general.


dougollerenshaw:
Closed. I should have branched from dev. Just making a new PR

Issue #1727, opened 2026-02-06
Detector config exposure time


saskiad:
Michael Xie was asking about the exposure time required for the detector config and pointing out how that information is not available/doesn't make sense for PMTs. He looked through other records which don't have it or have put in a default value that isn't logical. He discussed with Jerome who agreed that it doesn't make sense for PMTs.

Let's make this optional (non-breaking change)
We should figure out if cameras need it, in which case we keep it as optional (and maybe add a validator for camera detector types), or remove it in v3.0

dbirman:
We implemented this through the changes in the slap2 metadata PR https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1744

PR #1744, opened 2026-02-24
updated SLAP2 metadata components

m-xie:
https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1422

Updated aind-data-schema for SLAP2 acquisitions and instruments

- added examples for generating SLAP2+VCO instrument.json, SLAP2 acquisition.json, and SLAP2 structural z-stack acquisition.json
- updated Slap2Plane to include most current and useful fields
- updated DigitalMicromirrorDevice to remove legacy fields and long arrays
- made exposure_time optional in DetectorConfig to be compatible with SLAP2 SiPM detectors (https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1727)


examples/slap2_acquisition.py
```diff
@@ -0,0 +1,364 @@
...
+origin_ari.name = "Arbitrary Origin ARI"
+origin_ari.axis_unit = SizeUnit.UM
+origin_ari.origin = Origin.ORIGIN
+
+instrument_json_path = "instrument.json"
+instrument_json = Path(instrument_json_path).read_text()
+instrument = Instrument.model_validate_json(instrument_json)
+instrument_components = instrument.get_component_names()
+
+# temporary fix to remove manufacturer and harp type names from instrument components (recent updates in main branch fix this)
```
dbirman: I think this shouldn't be needed anymore if you are up-to-date with dev. I bumped 1422 to be up to date, please do so with your branch as well.

src/aind_data_schema/components/configs.py
```diff
@@ -205,24 +205,27 @@ class CoupledPlane(Plane):
...
+    target_name: Optional[str] = Field(default=None, title="Name of imaged target")
 
     mask_image_path: AssetPath = Field(
-        ..., title="Mask image path", description="Relative path from metadata json to file"
+        ..., title="Mask image path", description="Relative path from metadata json to imaging ROI masks (numbered by superpixel)"
     )
 
-    dilation_image_path: AssetPath = Field(
-        ..., title="Dilation image path", description="Relative path from metadata json to file"
+    unique_y_dilations: List[int] = Field(..., title="Unique Y dilations")
```
dbirman, line 221: What's the argument for splitting up x/y dilations like this and having unique values for one axis but not the other. We should future-proof ourselves here -- is this setting us up for problems if there's a change in slap2 in the future?

src/aind_data_schema/core/acquisition.py
```diff
@@ -28,6 +28,7 @@
     PatchCordConfig,
     ProbeConfig,
     SampleChamberConfig,
+    Slap2Plane,
```
dbirman: Are you adding this here because there are situations where you don't want to wrap a plane in an `ImagingConfig` object? Or this is maybe just a typo?

examples/slap2_structure_acquisition.py
```diff
@@ -0,0 +1,210 @@
...
+    wavelength=1030,
+    wavelength_unit=SizeUnit.NM,
+    power=150,
+    power_unit=PowerUnit.MW,
+)
+
+sipm_config = {
+    "Red": DetectorConfig(
+        device_name="Red SiPM",
+        trigger_type=TriggerType.INTERNAL,
```
dbirman: These are the PMTs right? I think we can just skip these configs potentially, or we should add a `TriggerType.ANALOG`

examples/slap2_instrument.py
```diff
@@ -0,0 +1,458 @@
...
+photodiode=Device(
+    name="Photodiode",
+    manufacturer=Organization.AIND,
+)
+
+slap2_1_computer=Computer(name=computer_names['SLAP2'])
+
+sipm_red=Detector(
+    name="SiPM Red",
+    detector_type=DetectorType.OTHER,
```
dbirman: There's a DetectorType.PMT, or are these something different? We should add the type in any case.

dbirman (changes requested):
This looks great, a few questions that we should iterate on in the various comments and then we should loop Saskia in to take a look as well.

examples/slap2_instrument.py
m-xie: These are silicon photomultipliers (SiPMs), an alternative detector to PMTs for 2p imaging (https://opg.optica.org/oe/fulltext.cfm?uri=oe-27-24-35830). If a SiPM DetectorType can be added, can definitely incorporate that here.

src/aind_data_schema/components/configs.py
m-xie: I set it up this way because X and Y dilations are split up in this way in the SLAP2 native metadata, where there is always only one X dilation per light path but there are potentially multiple Y dilations. Y dilation can vary across the plane, but X dilation is always constant in a plane.

X and Y dilations are conceptually different (i.e. they serve distinct purposes), so I don't forsee a situation in the future where we'd want to vary X dilation across a plane like we do with Y dilation (maybe @KasparP can confirm to make sure I'm not missing any potential future use cases). Also, since they are distinct, it might make it confusing to handle the metadata in a parallel way.

It is relatively straightforward to handle X dilation in the same way that Y dilation is handled now in the data schema, so if it is still easier to make that change, I am happy to! Just presenting my original reasoning, but I don't have any strong preferences either way.

examples/slap2_structure_acquisition.py
KasparP: These are similar to PMTs. The SiPMs have a model number that should probably be registered somewhere. In practice we use the same model in all cases.

examples/slap2_acquisition.py
m-xie: Will do!

src/aind_data_schema/components/configs.py
KasparP: Confirming that Michael is not missing any potential future use cases. It was a mistake to use the same word for 'X dilation' and 'Y dilation'. Although at base the mathematical operation is similar the purpose for imaging is very different and it leads to confusion like above.

examples/slap2_structure_acquisition.py
m-xie: The reason I made the `DetectorConfigs` for the SiPMs was that I thought they were required for the `detector` field in the `Channel` class. If we can put something else there, or if it can be skipped, I can remove these `DetectorConfigs`

src/aind_data_schema/core/acquisition.py
m-xie: I think this was a typo or a remnant from an old test, will remove!

examples/slap2_structure_acquisition.py
dbirman: I think we do want to know which detector matches each channel, so lets just add a new trigger type to flag for ourselves that these are not being triggered (internal or external) to avoid confusion later.

examples/slap2_instrument.py
m-xie: ~Added a new `DetectorType.SIPM`!~ Nevermind, realized that this needs to be added in `aind-data-schema-models` and I just changed my local copy

examples/slap2_structure_acquisition.py
m-xie: If I add a new `TriggerType` to [`configs.py`](https://github.com/m-xie/aind-data-schema/blob/f2d8349b8b0b4ae20cde3ec5d495ed3cf138367d/src/aind_data_schema/components/configs.py#L67
), is this what you were thinking?

```
class TriggerType(str, Enum):
    """Types of detector triggers"""

    INTERNAL = "Internal"
    EXTERNAL = "External"
    ANALOG = "Analog output, no trigger"
```

dbirman: Yes, enum strings though should just be a single word ideally, so just "Analog"

examples/slap2_instrument.py
dbirman: I'm going to add it now so that we can point to the new aind-data-schema-models version from this PR.

examples/slap2_structure_acquisition.py
m-xie: done!

examples/slap2_instrument.py
dbirman: Bump to aind-data-schema-models>=5.4.0,<6 and you should have access to it

m-xie: done!

dbirman (approved):

PR #1748, opened 2026-02-25
docs: improve docstring

### LINKED ISSUE #1743: Improve specimen_id description
"identifier for tissue removed from an animal" also specify that a standard approach is "<subject_id>" and if needed "_###"

dbirman:


docs/source/components/surgery_procedures.md
```diff
@@ -147,7 +147,7 @@ Description of a perfusion procedure that creates a specimen
 | Field | Type | Title (Description) |
 |-------|------|-------------|
 | `protocol_id` | `Optional[str]` | Protocol ID (DOI for protocols.io) |
-| `output_specimen_ids` | `List[str]` | Specimen ID (IDs of specimens resulting from this procedure.) |
+| `output_specimen_ids` | `List[str]` | Specimen ID (IDs of specimens resulting from this procedure. Single IDs should use the {subject_id}, multiple IDs should use a suffix format such as {subject_id}_###) |
```
saskiad: I'd rephrase that full brain should use the {subject_id}, but sectioned tissue should use {subject_id}_###

src/aind_data_schema/components/surgery_procedures.py
```diff
@@ -221,7 +221,10 @@ class Perfusion(DataModel):
     output_specimen_ids: List[str] = Field(
         ...,
         title="Specimen ID",
-        description="IDs of specimens resulting from this procedure.",
+        description=(
+            "IDs of specimens resulting from this procedure."
+            " Single IDs should use the {subject_id}, multiple IDs should use a suffix format such as {subject_id}_###"
```
saskiad: same comment as above

saskiad (changes requested):

docs/source/components/surgery_procedures.md
dbirman: What if they section out the spinal cord and nothing else? Is that {subject_id} or {subject_id}_###? I think we need to be more clear here if we want there to be implicit rules

saskiad: I don't have a strong sense - my inclination is yes, that should have a ## suffix. Do we do a lot of spinal cord sectioning? I was under the impression the spinal cord imaging we're doing is with it attached to the brain, but I'm not an expert on this

dbirman: I just want the rule to be clear to users, is it:

- If you section the entire brain, including any attached parts, use <subject_id>

Or:

- If you generate a single section, use <subject_id>

Either way the second rule is:

- If you generate subsections of an existing section, use <section_id>_###

So some examples:

- Section an entire brain: <subject_id>
- Section a brain + spinal cord: in both rules <subject_id>
- Section just a spinal cord: in option 1 unclear what to do, in option 2 <subject_id>
- Section just a single brain region: in option 1 unclear what to do, in option 2 <subject_id>
- Subsection a sectioned brain: both options <subject_id>_###

Does that make the problem more clear?

saskiad: First, I think that the rule that a whole brain (with or without the spinal cord) is subject id, everything else is subject_### is clear for all of those examples.

I think many people will read "generate a single section" and not understand what to do for the intact brain. There's no sectioning step involved, only perfusion. 
If I make a bunch of chunks, but I'm only looking at one of them and the rest are in a vial, is that one section or many? 

Ideally, I'd say that anything removed from an animal has to have ### attached, but that ship has sailed and I want to keep it contained to full brain.

dbirman: Fixed

saskiad (approved):

PR #1765, opened 2026-03-05
fix: remove behavior modality requirement for a LickSpoutAssembly

### LINKED ISSUE #1746: Behavior modality vaidation
Currently requires a lickspoutassembly in the DEVICES_REQUIRED check. These aren't being used in the chronic behavior/ephys experiments, because water is ad lib. So we need to remove that.

dbirman:
Removes the requirement that behavior assets include a LickSpoutAssembly


saskiad (approved):
That was so easy, it's like a bot could do it...

PR #1766, opened 2026-03-05
docs: clarify how specimen_id should be set

### LINKED ISSUE #1735: Docs: clarify that a whole-brain specimen ID is the subject_id
https://aind-data-schema.readthedocs.io/en/latest/procedures.html

dbirman:


dougollerenshaw:
Looks good. Nice simple clarification.

Naive question since I don't fully understand how doc building works: why is the same change in both `docs/base/core/procedures.md` and `docs/source/procedures.md`? Is one the result of the docs being auto-built?

dougollerenshaw (approved):

dbirman:
> Looks good. Nice simple clarification.
> 
> Naive question since I don't fully understand how doc building works: why is the same change in both `docs/base/core/procedures.md` and `docs/source/procedures.md`? Is one the result of the docs being auto-built?

The best solution I could come up with to construct a single file that includes both header information (manually written) and the object tables (auto-generated) was to write the header sections in the base/ folder, then run the auto-generator code which appends the object tables and moves the file to source/, Then finally sphinx runs off of the source/ files to create the actual html files.

PR #1774, opened 2026-03-10
fix: validate modalities

### LINKED ISSUE #1688: Validate modality between acquisition data streams and data_description.modalities

dbirman:
This PR validates modalities between:

- data_description and acquisition.data_streams/stimulus_epochs. I.e. all data_description modalities must appear in the acquisition streams or epochs and vice versa
- data_description and instrument, i.e. instrument must be a superset of data_description
- data_description and quality_control, i.e. QC must be a subset of data_description

Errors are raised as warnings. In V3 we will uplevel to validation errors.

While writing tests I also fixed a bug where the "BEFORE" and "AFTER" time validators weren't allowing identical times. This causes problems when people put in their instrument a calibration that occurred at the same time as the acquisition_start_time. I think it's fine for them to be equal since people are just using the start time as a short-hand for "happened as part of this acquisition".

### DIFF
--- a/src/aind_data_schema/core/metadata.py
+++ b/src/aind_data_schema/core/metadata.py
@@ -373,7 +373,9 @@ class Metadata(DataCoreModel):
                 parts.append(f"in acquisition but not data_description: {acq_modalities - dd_modalities}")
             if dd_modalities - acq_modalities:
                 parts.append(f"in data_description but not acquisition: {dd_modalities - acq_modalities}")
-            warnings.warn(f"Modality mismatch between acquisition.data_streams and data_description. {'; '.join(parts)}")
+            warnings.warn(
+                f"Modality mismatch between acquisition.data_streams and data_description. {'; '.join(parts)}"
+            )
 
     def _check_instrument_modalities(self, dd_modalities):
         """Warn if instrument modalities are not a superset of data_description modalities."""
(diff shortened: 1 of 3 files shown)


src/aind_data_schema/core/metadata.py
```diff
@@ -360,6 +360,56 @@ def validate_time_constraints(self):
...
+        for data_stream in self.acquisition.data_streams:
+            acq_modalities.update(data_stream.modalities)
+        if acq_modalities != dd_modalities:
+            parts = []
+            if acq_modalities - dd_modalities:
+                parts.append(f"in acquisition but not data_description: {acq_modalities - dd_modalities}")
+            if dd_modalities - acq_modalities:
+                parts.append(f"in data_description but not acquisition: {dd_modalities - acq_modalities}")
+            warnings.warn(
+                f"Modality mismatch between acquisition.data_streams and data_description. {'; '.join(parts)}"
```
saskiad, line 377: but something like "behavior" won't necessarily show up in the data_stream as a modality. How do we handle that?

saskiad (commented):
need to handle the behavior exception

dbirman:
We've decided to hold on this PR until v3 upgrades. At that time we'll handle via the upgrader repairing all the assets that have behavior in data_description but no corresponding DataStream object.

PR #1690, opened 2026-01-15
Add barseq acquisition

dougollerenshaw:
Work in progress. Lots of ancillary files that we won't actually want in the repo. Remove before merging. 

Closes: https://github.com/AllenNeuralDynamics/aind-scientific-computing/issues/569

PR #1781, opened 2026-03-20
Barseq blackbox acquisition

### LINKED ISSUE #1761: "Black box" acquisition
Need to accomodate use case where sample is mailed to someone else who collects data and sends data back.

- [ ] acquisition probably needs some modifications - e.g. not require data stream, or modify data stream?
- [ ] instrument not required
- [ ] validators hinge on modalities to determine which fields are required for valid metadata.

dougollerenshaw:
Creates simple black-box style acquisition.json files for barseq acquisitions for subjects 780345 and 780346. 

Note that I had a much more involved/complicated PR that attempted to capture the actual acquisition details here: https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1690

I abandoned that effort after Polina shared the raw data paths and @saskiad and I spent some time trying to understand the structure of the actual raw acquisition files. We realized there was simply too much we didn't understand about how the image files were acquired and subsequently processed for us to be able to accurately capture the process for now. 

This PR closes https://github.com/AllenNeuralDynamics/aind-scientific-computing/issues/569 and closes https://github.com/AllenNeuralDynamics/aind-data-schema/issues/1761


examples/barseq_780345_acquisition.json
```diff
@@ -0,0 +1,91 @@
...
+   "experimenters": [
+      "Mara Rue",
+      "Shannon Khem",
+      "Tracy"
+   ],
+   "protocol_id": [
+      "https://www.protocols.io/view/barseq-2-5-kqdg3ke9qv25/v1"
+   ],
+   "ethics_review_id": null,
+   "instrument_id": "Dogwood",
```
saskiad: there shouldn't be an instrument for black box

examples/barseq_780345_acquisition.json
```diff
@@ -0,0 +1,91 @@
...
+      "780345_bar038",
+      "780345_bar039",
+      "780345_bar040",
+      "780345_bar041",
+      "780345_bar042",
+      "780345_bar043",
+      "780345_bar044"
+   ],
+   "acquisition_start_time": "2025-02-24T00:00:00Z",
+   "acquisition_start_tz": 0,
```
saskiad: is that the right way to do timezone?

examples/barseq_acquisition.py
```diff
@@ -0,0 +1,137 @@
...
+    "Final processed output is a cell x gene x barcode table registered to Allen CCFv3."
+)
+
+SUBJECTS = {
+    "780345": {
+        # Experimenters from experiment_detail.txt files in raw data slide folders.
+        # Source: /allen/aind/stage/barseq/20250303_780345_slide7_maxprojection/experiment_detail.txt
+        #         /allen/aind/stage/barseq/20250305_780345_slide8_maxprojection/experiment_detail.txt
+        #         /allen/aind/stage/barseq/20250317_780345_slide9_maxprojection/experiment_detail.txt
+        "experimenters": ["Mara Rue", "Shannon Khem", "Tracy"],
```
saskiad: if we're doing black box, I'd have this be Barseq team and not name people.

examples/barseq_acquisition.py
```diff
@@ -0,0 +1,137 @@
...
+# Raw data location on allen network share.
+# Source: Polina Kosillo, 2026-02-16 Teams chat (MapSeq/BARseq metadata channel)
+BARSEQ_RAW_DATA_PATH = "/allen/aind/stage/barseq/"
+
+# Source: MAPseq-BARseq methods_forSciComp.pdf, header section
+BARSEQ_PROTOCOL_ID = ["https://www.protocols.io/view/barseq-2-5-kqdg3ke9qv25/v1"]
+
+# Instrument ID for the Nikon Ti2-E "Dogwood" spinning disk confocal.
+# Source: aind-data-schema barseq instrument PR #1685
+INSTRUMENT_ID = "Dogwood"
```
saskiad: no instrument id for black box.

saskiad (dismissed):

examples/barseq_780345_acquisition.json
dougollerenshaw: It's schema compliant this way, but we can pass local time with offset also. See https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/commits/62a9dceb5ea90c3bbb695256c0492df9f85dc2ce

examples/barseq_acquisition.py
dougollerenshaw: Done. See https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/commits/02724590d7f69c36da42a5b232128878c5b19780

examples/barseq_780345_acquisition.json
dougollerenshaw: Are you sure here? I can understand instrument not being required in a black-box acquisition. But in this case we have the instrument name and the corresponding instrument.json. I don't see a downside to including it. I can remove if you'd still prefer, but I just wanted to double check first.

examples/barseq_acquisition.py
dougollerenshaw: See my comment above: https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781#discussion_r2990469261

examples/barseq_780346_acquisition.json
```diff
@@ -0,0 +1,97 @@
...
+      "Mara Rue",
+      "Shannon Khem"
+   ],
+   "protocol_id": [
+      "https://www.protocols.io/view/barseq-2-5-kqdg3ke9qv25/v1"
+   ],
+   "ethics_review_id": null,
+   "instrument_id": "Dogwood",
+   "acquisition_type": "BarcodeSequencing",
+   "notes": "BARseq acquisition performed across multiple slides imaged over multiple days on the Dogwood spinning disk confocal. Raw data stored at smb://allen/aind/stage/barseq/. Full acquisition includes gene sequencing (7 cycles), barcode sequencing (15 cycles), and one hybridization cycle. Each slide folder contains all three acquisition types. Final processed output is a cell x gene x barcode table registered to Allen CCFv3. Processed output: s3://aind-private-data-prod-o5171v/780346_2025-06-11_00-00-00/BARseq/combined_neurons_clust_CCFv2.mat.",
```
dbirman: I wouldn't store the VAST or s3 paths here. If they are critical and persistent they can go in the `Metadata.other_identifiers`, but in this case the location should already match the s3://path you are listing. We also don't normally store asset information like the location of the .mat file.

dbirman (approved):
Seems fine from my perspective. I wouldn't include paths to S3, relative, or VAST locations in general

examples/barseq_780346_acquisition.json
dougollerenshaw: OK, fixed. See https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/commits/289eec52c6296432e2938bc2772e7b886e843a35

examples/barseq_780345_acquisition.json
saskiad: yeah, I'm sure. We are treating this as we send it to a team and they send us something back. What they do is a black box.

dougollerenshaw:
This PR now includes a new "ExternalDataStream" class that makes `instrument_id` optional. 

Note that the two json files in the `examples/2026.01.16_barseq_acquisition` will be removed before merging. These are only here to verify that the actual acquisition files are reasonable.

src/aind_data_schema/core/acquisition.py
```diff
@@ -406,12 +429,13 @@ def coerce_fixed_offset_tz_string(cls, v):
     )
 
     # Acquisition data
-    data_streams: List[DataStream] = Field(
+    data_streams: List[Union[DataStream, ExternalDataStream]] = Field(
```
dbirman: We have a `DiscriminatedList` wrapper for doing List[Union], it's important to use it since it specifies that the object_type field should be used as the discriminator.

src/aind_data_schema/core/acquisition.py
```diff
@@ -478,8 +512,12 @@ def specimen_required(self):
         return self
 
     @classmethod
-    def _merge_data_streams(cls, streams: List[DataStream], overlap_s: int = 120) -> List[DataStream]:
-        """Merge two lists of data streams"""
+    def _merge_data_streams(
```
dbirman: I would prefer that you filter out the external data streams outside of this function and pass it just the data streams to merge. This is almost acting like a side effect which doesn't seem ideal.

tests/test_acquisition.py
```diff
@@ -53,6 +59,46 @@ def test_constructors(self):
             scan1_dict["notes"] = ""
             MRIScan.model_validate(scan1_dict)
 
+    def test_external_data_stream(self):
```
dbirman, line 62: I find that it's kind of an anti-pattern in the aind-data-schema tests to construct examples inside of the tests. What happens is that when we make changes later we then have to come back and update all the tests which can get really tedious. My recommendation and the pattern I try to use is to import the objects directly from the examples/ folder for testing. Can you change the code here to do that using the barseq examples?

examples/2026.01.16_barseq_acquisition/barseq_780345_acquisition.json
```diff
@@ -0,0 +1,85 @@
+{
```
dbirman: Now that we're preparing to merge we shouldn't be generating these .json files.

examples/barseq_acquisition.py
```diff
@@ -0,0 +1,111 @@
...
+                stream_start_time=params["acquisition_start"],
+                stream_end_time=params["acquisition_end"],
+                modalities=[Modality.BARSEQ],
+                notes="Acquired externally by BARseq imaging team.",
+            )
+        ],
+    )
+
+
+if __name__ == "__main__":
```
dbirman, line 33: Can you remove the arguments from this? The other examples don't have this extra functionality so we shouldn't add it here. Also I don't think the way this generates two files is going to work with the `test_examples.py` code, so we should fix that. Each .py file needs to generate a single matching .json file for the tests to succeed (the fact that they are apparently passing worries me a little bit, I think it should have failed here)

dbirman (changes requested):
Requesting some changes, see comments.

Also, I think you still need to add a test in `test_metadata.py` that creates a Metadata object with an acquisition that has only an external data stream and no instrument. This should crash the Metadata-level validator that checks the core file requirements and then we can modify how the requirements work to allow this.

src/aind_data_schema/core/acquisition.py
dougollerenshaw: Like this? 
https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/commits/b303d4beac773b769aef01ba85fc051999eb8bbb

src/aind_data_schema/core/acquisition.py
dougollerenshaw: How's this? https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/changes/89250f233794a9a2025e9d26fbeae604a5485105

examples/barseq_acquisition.py
dougollerenshaw: I dramatically simplified the example to make it look like other example files. I moved the specific logic for these subjects to a separate repo so I can create the JSONs we need for Polina there.

dougollerenshaw: see https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/changes/ca8c6832a0515ec52c2db584f3b516f55367c717

examples/2026.01.16_barseq_acquisition/barseq_780345_acquisition.json
dougollerenshaw: See https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/commits/2ab00e1e9bf3a8acd75d05e9511a7ef26bb7d558

tests/test_acquisition.py
dougollerenshaw: see https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/changes/6fc71ae97a90b64129dca8f8e19243ffac524451

dougollerenshaw:
> Requesting some changes, see comments.
> 
> Also, I think you still need to add a test in `test_metadata.py` that creates a Metadata object with an acquisition that has only an external data stream and no instrument. This should crash the Metadata-level validator that checks the core file requirements and then we can modify how the requirements work to allow this.

See this commit: https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/changes/aa5f5205fba42913b9c48b69ad6594608fdf6787

dougollerenshaw:
OK, I think everything is addressed. Mind another look?

docs/source/acquisition.md
```diff
@@ -85,13 +85,13 @@ while the StimulusEpoch represents all stimuli being presented.
 | `experimenters` | `List[str]` | experimenter(s)  |
 | `protocol_id` | `Optional[List[str]]` | Protocol ID (DOI for protocols.io) |
 | `ethics_review_id` | `Optional[List[str]]` | Ethics review ID  |
-| `instrument_id` | `str` | Instrument ID (Should match the Instrument.instrument_id) |
+| `instrument_id` | `Optional[str]` | Instrument ID (Should match the Instrument.instrument_id. Required when data_streams contains a DataStream.) |
```
dbirman: Not blocking but I think my preferred language is "Required when instrument metadata is available."

docs/source/acquisition.md
```diff
@@ -85,13 +85,13 @@ while the StimulusEpoch represents all stimuli being presented.
...
 | `calibrations` | List[[Calibration](components/measurements.md#calibration) or [VolumeCalibration](components/measurements.md#volumecalibration) or [PowerCalibration](components/measurements.md#powercalibration)] | Calibrations (List of calibration measurements taken prior to acquisition.) |
 | `maintenance` | List[[Maintenance](components/measurements.md#maintenance)] | Maintenance (List of maintenance on instrument prior to acquisition.) |
-| `data_streams` | List[[DataStream](acquisition.md#datastream)] | Data streams (A data stream is a collection of devices that are acquiring data simultaneously. Each acquisition can include multiple streams. Streams should be split when configurations are changed.) |
+| `data_streams` | List[[DataStream](acquisition.md#datastream) or [ExternalDataStream](acquisition.md#externaldatastream)] | Data streams (A data stream is a collection of devices that are acquiring data simultaneously. Each acquisition can include multiple streams. Streams should be split when configurations are changed. Use ExternalDataStream for acquisitions performed externally where full instrument metadata is unavailable.) |
```
dbirman: Same comment as above here "for acquisitions where instrument metadata is unavailable."

docs/source/acquisition.md
```diff
@@ -131,6 +131,18 @@ same time.
 | `connections` | List[[Connection](components/connections.md#connection)] | Connections (Connections are links between devices that are specific to this acquisition (i.e. not already defined in the Instrument)) |
 
 
+### ExternalDataStream
+
+A simplified data stream for acquisitions performed externally, where full instrument metadata is unavailable.
```
dbirman: You can drop the "full". If there's partial instrument metadata we would still want it.

dbirman (approved):
This looks good to me, hard to know whether every edge case is handled for data_stream having two kinds of objects now but we'll find out pretty quickly if there are issues!

A few minor comments on wording, not critical to change them but I think it will be a tiny bit easier for users with the changes.

docs/source/acquisition.md
dougollerenshaw: See https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/commits/d29c08d128bd5cb3e50e0fd3b0ce63cc54cfe945

docs/source/acquisition.md
dougollerenshaw: see https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/commits/d29c08d128bd5cb3e50e0fd3b0ce63cc54cfe945

docs/source/acquisition.md
dougollerenshaw: See https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1781/commits/d29c08d128bd5cb3e50e0fd3b0ce63cc54cfe945

Issue #1782, opened 2026-03-24
specimen_id validator for procedures not handling lists of IDs


gouwens:
Giving a specimen procedure a list of specimen IDs produces a validation error even when each of the specimen IDs conforms to the proper format. This is because the validator is not handling the possibility that the `specimen_id` field could be either a `str` or a list of `str`s. 

Here is a test that is currently failing:
```python
def test_validate_subject_specimen_id_list_valid(self):
    """Test that specimen_id accepts a list of strings when all contain subject_id"""

    valid_procedure = Procedures(
            subject_id="12345",
            specimen_procedures=[
                SpecimenProcedure(
                    specimen_id=["12345_001", "12345_002"],
                    procedure_type="Other",
                    start_date=date.fromisoformat("2020-10-10"),
                    end_date=date.fromisoformat("2020-10-11"),
                    experimenters=["Mam Moth"],
                    protocol_id=["10"],
                    notes="some notes",
                )
            ],
        )
    self.assertIsNotNone(valid_procedure)
```

This currently results in the validation error:
```
Value error, specimen_id must be an extension of the subject_id. [type=value_error, input_value={'subject_id': '12345', '...], notes='some notes')]}, input_type=dict]
```

dbirman:
Closed by https://github.com/AllenNeuralDynamics/aind-data-schema/pull/1783

PR #1783, opened 2026-03-24
Handle lists of specimen IDs in procedures validator

gouwens:
Addressing #1782


gouwens:
Hmm, thought it would handle cases of all single strings, but it's not - working on an update.

dbirman (approved):

PR #1797, opened 2026-04-15
fix: add Reagent as an option for Acquisition.manipulations

### LINKED ISSUE #1796: Add to Acquisition.manipulations DiscriminatedList SpecimenProcedure (for HCR)

dbirman:
This PR adds Reagent as an option for Acquisition.manipulations and an optional protocol_id list field to the Manipulation class. It also refactors `protocol_id` as a Mixin class that we include in various places. I was getting a little worried about duplicating this field all over the place, this ensures that the field always looks exactly the same with the same description.


dbirman:
Add optional protocol_id to Manipulation, only Reagent as option

src/aind_data_schema/components/identifiers.py
```diff
@@ -92,3 +92,15 @@ class Code(DataModel):
         title="Core dependency",
         description="For code with a core software package dependency, e.g. Bonsai",
     )
+
+
+class ProtocolMixin(BaseModel):
```
saskiad: not sure I get the mixin thing, but I think i do...
BUT, I do think all protocol_id fields should be lists, so if we're creating ways to add this to things, let's only use the list one below, and then don't need this one.

saskiad: on second thought ... there are situations where it should be a single item. So keep it.

src/aind_data_schema/components/identifiers.py
```diff
@@ -92,3 +92,15 @@ class Code(DataModel):
...
+    """Mixin that adds a protocol_id field (single DOI string)"""
```
saskiad: can you explain the advantage of using the Mixin to me? I think I get it - it just puts this into whatever class has it - but ... why is that better than having the line of code? I'm mostly thinking about other folks understanding the code. It feels like it adds some amount of opacity to the code.

saskiad (approved):

PR #1802, opened 2026-04-16
refactor: only allow wrapped NonSurgeryInjections in Procedures.subject_procedures

### LINKED ISSUE #1801: Injections have no date, we need a wrapper class in Procedures
Currently Injection objects are allowed without a Surgery, but they then have no date. We need a wrapper similar to how we did `Manipulation` in acquisition, it can just be called `Procedure` or `SubjectProcedure`.

dbirman:
This PR fixes an oversight we had in the `Procedures.subject_procedures` where we were allowing raw `Injection` objects. These don't have a start_date field since typically they show up inside a Surgery object. This PR creates a new wrapper `NonSurgeryInjection` that adds a required start_date, ethics_review_id, and notes.

**Alternative**
We could also remove Injection from the top-level and only allow Surgery objects, that would avoid adding a new wrapper class to the schema.


src/aind_data_schema/components/subject_procedures.py
```diff
@@ -39,6 +39,17 @@ class GenericSubjectProcedure(DataModel):
     notes: Optional[str] = Field(default=None, title="Notes")
 
 
+class InjectionProcedure(DataModel):
+    """Injection procedure, which may include one or more injections at different locations/depths"""
+
+    start_date: date = Field(..., title="Start date")
+    ethics_review_id: str = Field(..., title="Ethics review ID")
+    protocol_id: Optional[str] = Field(default=None, title="Protocol ID", description="DOI for protocols.io")
```
saskiad, line 49: Injection class already has protocol_id. I'm okay moving it here, but we should remove it from the injection class

src/aind_data_schema/components/subject_procedures.py
```diff
@@ -39,6 +39,17 @@ class GenericSubjectProcedure(DataModel):
...
+    injections: List[Injection] = Field(..., title="Injections", min_length=1)
+    anaesthesia: Optional[Anaesthetic] = Field(default=None, title="Anaesthesia")
```
saskiad: no anaesthesia. If there's anaesthesia involved it becomes a surgery.

src/aind_data_schema/components/subject_procedures.py
```diff
@@ -39,6 +39,17 @@ class GenericSubjectProcedure(DataModel):
     notes: Optional[str] = Field(default=None, title="Notes")
 
 
+class InjectionProcedure(DataModel):
+    """Injection procedure, which may include one or more injections at different locations/depths"""
```
saskiad: I'm trying to think if there's a different name that will help with disambiguating when this is used rather than BrainInjection or Injection.

saskiad (changes requested):
let's discuss this one

src/aind_data_schema/components/subject_procedures.py
dbirman: Deprecated

src/aind_data_schema/components/subject_procedures.py
dbirman: Removed

src/aind_data_schema/components/subject_procedures.py
dbirman: Renamed to NonSurgicalInjection

saskiad (approved):

PR #1807, opened 2026-04-27
fix: change DataProcess.output_path to list

tmchartrand:
Fixes AllenNeuralDynamics/biodata-schema#21
may break existing code for reading the output_path field, but writing/creation and queries are backwards compatible


tmchartrand:
postponing for 3.0

PR #1813, opened 2026-05-01
Add git hash to Code model

### LINKED ISSUE #1789: Add commit_hash Optional[str] field to Code
The Code object doesn't make it easy to track both the semver and commit hash, which makes it difficult to track dev branches when they are being used (they may share a semver with a release version even though they are on different commits).

Users get around this by putting commit hashes into the version field, but it would be better to track these separately.

https://aind-data-schema.readthedocs.io/en/latest/components/identifiers.html#code

bruno-f-cruz:
Closes #1789

### DIFF
--- a/src/aind_data_schema/components/identifiers.py
+++ b/src/aind_data_schema/components/identifiers.py
@@ -5,8 +5,8 @@ from pathlib import Path
 from typing import Dict, List, Optional
 
 from aind_data_schema_models.registries import Registry
-from pydantic import Field
-
+from pydantic import Field, model_validator
+import warnings
 
 from typing import Annotated
 from pydantic import StringConstraints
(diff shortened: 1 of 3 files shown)


src/aind_data_schema/components/identifiers.py
```diff
@@ -67,12 +70,24 @@ class Container(DataModel):
     uri: str = Field(..., title="URI", description="URI of the container, e.g. Docker Hub URL")
 
 
+GitHash = Annotated[
+    str,
+    StringConstraints(
+        pattern=r"^[0-9a-fA-F]{7,40}$",
```
dbirman: I would probably allow up to 64 since at some point git will switch to sha-256

src/aind_data_schema/components/identifiers.py
```diff
@@ -67,12 +70,24 @@ class Container(DataModel):
...
+]
+
+
 class Code(DataModel):
     """Code or script identifier"""
 
     url: str = Field(..., title="Code URL", description="URL to code repository")
     name: Optional[str] = Field(default=None, title="Name")
     version: Optional[str] = Field(default=None, title="Code version")
+    git_hash: Optional[GitHash] = Field(
```
dbirman: Can we do "commit_hash" / CommitHash to avoid putting the name of a tool in the schema?

src/aind_data_schema/components/identifiers.py
```diff
@@ -67,12 +70,24 @@ class Container(DataModel):
...
+        default=None, title="Git hash", description="Git hash of the code repository state"
+    )
 
     container: Optional[Container] = Field(default=None, title="Container")
     run_script: Optional[Path] = Field(default=None, title="Run script", description="Path to run script")
```
dbirman, line 91: Can you please add a model_validator that throws a warnings.warn message if neither version or git_hash are present -- it can just say something like "One of Code.version or Code.git_hash will be expected to be set in future versions of the schema". I think we can't enforce this now since it might break old code.

dbirman (approved):
Looks good, I requested some small changes to future proof this. I went ahead and approved so you can merge whenever those are changed.

PR #1832, opened 2026-05-13
refactor: rework coordinates

### LINKED ISSUE #1828: Circular arena
NBA Issue

the Arena device class has a size field that uses Scale for a 3D size. How do we use this for a circular arena?

### LINKED ISSUE #1831: Make coordinate systems / transforms completely flexible
Add:

- intrinsic/extrinsic flag for Translation
- intrinsic/extrinsic flag for Rotation
- rotation direction setting for Rotation, default right-hand rule
- rotation axis order setting for Rotation
- rotation pivot (global/local origin)
- specify handedness in coordinate systems because LLMs struggle to figure this out otherwise
- stop using Scale when we mean RectangleGeometry (see also #1828)

dbirman:
This PR adds the full suite of coordinate system definitions:

- Local/Global reference frames for translations/rotations
- Pivots for scale/rotation
- Order for rotation
- All mouse anatomical structures available as Origins

It deprecates the to_matrix functions which are no longer applicable if you use local frames (we were assuming global before).
And finally it deprecates the coordinate system library classes with Depth axes since those are now represented by local transforms.

A new coordinate_system_builder in the docs helps users construct their coordinate systems properly (more work to do there, but better to get it out and then make changes)

**Important**: we decided that `coordinate_system` as a name was confusing. The names are now explicitly `global_coordinate_system` or `local_coordinate_system` to make it more clear.


dbirman:
- [x] Make the monitor a little more realistic
- [x] Make it more obvious which axes are which (thicker line for global and make them longer?)

dyf (changes requested):
Rename `reference_frame` attributes to `reference_coordinate_system`.

Confirm that Rotations need both a `pivot` and a `reference_coordinate_system`.

dyf (approved):

PR #1833, opened 2026-05-15
chore: build ISI instrument example

### LINKED ISSUE #1794: ISI instrument.json
See excel file from Shiella.

- [ ] Build instrument.json as an example
- [ ] Send to metadata-service
- [ ] Request SIPE to add instrument metadata request to their upload job settings (assuming V2) or include the instrument.json (for V1)

dbirman:
Adds an ISI instrument example, this will be used to backfill all existing ISI records. There are three ISI rigs, "ISIV.1" "ISIV.2" and "ISIV.3" that show up in existing data assets, unfortunately variations between them aren't captured here but I'm told they are basically identical.

Small note: the existing session metadata has a device listed "ISI LED" but since the actual rig has two LEDs I think the simplest thing is as part of the migration we will rename the field in the device configuration. This way we can have more informative names in the instrument metadata.

### DIFF
--- a/examples/isi_instrument.py
+++ b/examples/isi_instrument.py
@@ -1,0 +1,273 @@
+"""Generates an example JSON file for an ISI (Intrinsic Signal Imaging) instrument
+
+Based on the ISI Full Rig BOM (AIBS part 0113_000-00, Rev A).
+The rig consists of:
+  - ISI Table Assembly (0113_100-00)
+  - ISI System Structure (0113_200-00)
+  - Camera, Stage, and Mount Assembly (0113_300-00)
+  - Headframe Clamp Assembly (0113_400-00)
+  - ISI Eye Tracking Assembly (0123_500-00)
+  - Camera/Lens Assembly (0113_550-00) with tandem Nikon lenses and Andor Zyla sCMOS
+  - LED Light Ring Assembly (0113_530-00)
+"""
+
+import argparse
+from datetime import date
+
+from aind_data_schema_models.coordinates import AnatomicalRelative
+from aind_data_schema_models.devices import (
+    CameraChroma,
+    CameraTarget,
+    DataInterface,
+    DetectorType,
+    FilterType,
+)
+from aind_data_schema_models.modalities import Modality
+from aind_data_schema_models.organizations import Organization
+from aind_data_schema_models.units import FrequencyUnit, SizeUnit
+
+from aind_data_schema.components.coordinates import CoordinateSystemLibrary
+from aind_data_schema.components.devices import (
+    Camera,
+    CameraAssembly,
+    Computer,
+    DAQDevice,
+    Device,
+    Filter,
+    Lens,
+    LightEmittingDiode,
+    Monitor,
+    MotorizedStage,
+)
+from aind_data_schema.core.instrument import Instrument
+
+acquisition_computer = Computer(
+    name="Acquisition Computer",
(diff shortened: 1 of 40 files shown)


examples/isi_instrument.py
```diff
@@ -0,0 +1,248 @@
...
+    name="Andor Zyla 5.5 sCMOS",
+    manufacturer=Organization.OXFORD_INSTRUMENTS,
+    model="Zyla 5.5 sCMOS",
+    detector_type=DetectorType.CAMERA,
+    data_interface=DataInterface.USB,
+    chroma=CameraChroma.BW,
+)
+
+isi_lens_35mm = Lens(
+    name="Nikon NIKKOR 35mm f/1.4",
```
saskiad: this is a nice to have: I think having the name contain the full model number isn't best practice. It's not really bad, but it sets a precedent of doing this, and we want to avoid people cramming metadata into names instead of using other fields. I know if this case, we need to distinguish the two nikon lens, but we don't need the full manufacturer/model in the device name. Could be "lens 35mm" or something. 
Kind of the same with the camera above too.

examples/isi_instrument.py
```diff
@@ -0,0 +1,248 @@
...
+
+isi_lens_105mm = Lens(
+    name="Nikon Micro-NIKKOR 105mm f/2.8",
+    manufacturer=Organization.NIKON,
+    model="Micro-NIKKOR 105mm f/2.8",
+    notes="Rear lens of tandem-lens assembly, facing the camera.",
+)
+
+isi_bandpass_filter = Filter(
+    name="Semrock FF01-630/92-50 Bandpass Filter",
```
saskiad: this is prime example of above. We already have the bandapass and center wavelength in real fields. Make this simpler.

examples/isi_instrument.py
```diff
@@ -0,0 +1,248 @@
...
+eye_tracking_lens = Lens(
+    name="InfiniStix 130mm W.D. 0.73x",
+    manufacturer=Organization.INFINITY_PHOTO_OPTICAL,
+    model="Proximity Series 130mm W.D./0.73x",
+)
+
+eye_tracking_camera_assembly = CameraAssembly(
+    name="Eye Tracking Camera Assembly",
+    target=CameraTarget.EYE,
+    relative_position=[AnatomicalRelative.ANTERIOR],
```
saskiad, line 133: usually eye tracking isn't done from the anterior...

examples/isi_instrument.py
```diff
@@ -0,0 +1,248 @@
...
+)
+
+eye_tracking_collimating_lens = Lens(
+    name="Thorlabs LB1092-B-ML Bi-Convex Lens",
+    manufacturer=Organization.THORLABS,
+    model="LB1092-B-ML",
+)
+
+led_ring_green = LightEmittingDiode(
+    name="LED Ring Green 527nm",
```
saskiad: keep the color, remvoe the wavelength from the name.

examples/isi_instrument.py
```diff
@@ -0,0 +1,248 @@
...
+    name="LED Ring Red 635nm",
+    manufacturer=Organization.OTHER,
+    model="HLMP-EG08-Y2000",
+    wavelength=635,
+    wavelength_unit=SizeUnit.NM,
+    notes="9 units in LED light ring assembly (0113_530-00).",
+)
+
+stimulus_monitor = Monitor(
+    name="ASUS PA248Q Stimulus Monitor",
```
saskiad: just "Stimulus Monitor" works fine here

saskiad (approved):
I'll allow it, but I'd love to make the device names simpler so that people don't see this as how this should be done. Also makes it harder to work with during acqusition when the device names are long and detailed.

PR #1835, opened 2026-05-16
refactor: remove all Organization.X groups

### LINKED ISSUE #1834: Replace all Organization.X requirements with the more general .ONE_OF

dbirman:
PR removes all Organization.X groups, which will allow us to remove them from aind-data-schema-models and stop supporting them. This simplifies maintenance of the schema substantially. 

It also makes some Device subclasses like Lens superfluous now that there isn't a LENS_MANUFACTURERS group. We can't remove them until 3.X though for backward compatibility.

### DIFF
--- a/docs/source/components/devices.md
+++ b/docs/source/components/devices.md
@@ -57,7 +57,6 @@ Camera Detector
 | Field | Type | Title (Description) |
 |-------|------|-------------|
 | `detector_type` | [DetectorType](../aind_data_schema_models/devices.md#detectortype) |   |
-| `manufacturer` | [Organization](../aind_data_schema_models/organizations.md#organization) |   |
 | `data_interface` | [DataInterface](../aind_data_schema_models/devices.md#datainterface) | Data interface  |
 | `cooling` | [Cooling](../aind_data_schema_models/devices.md#cooling) | Cooling  |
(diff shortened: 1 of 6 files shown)


docs/source/components/devices.md
```diff
@@ -85,6 +84,7 @@ Camera Detector
 | `driver_version` | `Optional[str]` | Driver version  |
 | `name` | `str` | Device name  |
 | `serial_number` | `Optional[str]` | Serial number  |
+| `manufacturer` | Optional[[Organization](../aind_data_schema_models/organizations.md#organization)] | Manufacturer  |
```
saskiad: why are we making this optional?

saskiad: ditto all the edits below as well

dbirman: Oh I see, we did a "bad" stylistic thing by overriding an Optional field in the parent class as a required field in the subclasses. That's fine, I will change it back to that.

dbirman: Fixed

saskiad: alternatively, we can make manufacturer required for the base Device class and remove it from all these sub classes?

dbirman: That's a V3 breaking change

saskiad: yes, that's true.

saskiad (approved):

PR #1836, opened 2026-05-19
feat: add name field to DataAsset (#1787)

### LINKED ISSUE #1787: fix DataProcess.input_data to match DataDescription.source_data
we may be able to start aligning these representations without breaking anything - input_data is a DataAsset object with a s3 url, while source_data is just an asset name. 
maybe we want to allow either url or name in the DataAsset as a non-breaking change and move in practice to name? 
we should also check more thoroughly on input_data usage - did the upgrader drop any input_data represented in 1.0? does the transfer service use this field?

tmchartrand:
infers name from url if in s3://aind-open-data/

I chose not to add a validation pattern to the name field since I assume we want to allow inputs matching old or new name patterns. Let me know if that would be preferred instead.


dbirman (approved):
Looks good, small lint fix you need to do still

PR #1858, opened 2026-06-16
added joystick config

### LINKED ISSUE #1829: Joystick and force plate
NBA issue

Follow up with Vivek and Ines. 

The only thing these might require is isometric vs non-isometric. Need more information though.

saskiad:
closes #1829

### DIFF
--- a/src/aind_data_schema/components/configs.py
+++ b/src/aind_data_schema/components/configs.py
@@ -460,6 +460,12 @@ class OlfactometerConfig(DeviceConfig):
     )
 
 
+class JoystickConfig(DeviceConfig):
+    """Configuration of joystick"""
+
+    is_isometric: bool = Field(..., title="Is isometric")
+
+
 # EPHYS CONFIGS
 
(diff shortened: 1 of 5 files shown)


saskiad:
I don't think there's a need for a specific joystick device, I think the generic device is sufficient - and if they have position info the position device. 
There is one config setting which is whether it is being used in isometric or non-isometric mode. So I added the config to acquisition for that.

src/aind_data_schema/components/configs.py
```diff
@@ -460,6 +460,12 @@ class OlfactometerConfig(DeviceConfig):
     )
 
 
+class JoystickConfig(DeviceConfig):
+    """Configuration of joystick"""
+
+    is_isometric: bool = Field(..., title="Is isometric")
```
dbirman: Can you just add a description here, something like "Whether the joystick is locked in place"

dbirman (approved):

PR #1864, opened 2026-06-27
docs: schema tree function and interactive digrams

dbirman:
(1) PR adds a function in `schema_tree.py` that generates a stylized schema tree, looks like this basically:

- **Acquisition** — Description of data acquisition metadata including streams, stimuli, and experimental setup.
  - `protocol_id` (Optional[LIST OF: str]) — Protocol ID: DOI for protocols.io
  - `subject_id` (str, required) — Subject ID: Unique identifier for the subject
  - `specimen_id` (Optional[ONE OF: str, LIST OF: str]) — Specimen ID: Required for in vitro modalities. Standard format is {subject_id} with a _### suffix, as needed
  - `acquisition_start_time` (AwareDatetime, required) — Acquisition start time: During validation, timezone information will be moved into the acquisition_start_tz field.
  - `acquisition_start_tz` (Optional[ONE OF: int, TimeZoneName]) — Acquisition start timezone: Automatically populated by a validator based on acquisition_start_time. Will be a TimeZoneName (IANA name) when the datetime uses a ZoneInfo timezone, or an integer UTC offset in hours for fixed-offset timezones. Use ZoneInfo (from the zoneinfo standard library) to preserve the named timezone.

But it keeps going for the entire schema. This will be used for the appendix in the data-schema paper, as well as for the aind-data-mcp.

(2) Using the new schema tree the code now adds a new section to the docs with an interactive diagram, see preview: https://aind-data-schema--1864.org.readthedocs.build/en/1864/


dougollerenshaw (approved):
The result looks good. Very cool! But I didn't even try to review code because I don't know how to read typescript.

PR #1866, opened 2026-06-30
feat: thermistor assembly and implant

### LINKED ISSUE #1585: Thermistor Device and Implant
- [ ] Add Thermistor Device
- [ ] Add ThermistorImplant
- [ ] Add to Surgery procedures union

Info to be provided by Tiffany

dbirman:
Adds a new DeviceImplant to support thermistor wire implants, and adds an example showing how to use it. Molex is not an available manufacturer, we'll have to add it separately (we can do that before merging, if we want).


examples/thermistor_procedures.py
```diff
@@ -0,0 +1,87 @@
...
+    name="Thermistor wire",
+    manufacturer=Organization.TE_CONNECTIVITY,
+    model="GAG22K7MCD419",
+    notes="Digi-Key part number: GAG22K7MCD419. "
+    "https://www.digikey.com/en/products/detail/te-connectivity-measurement-specialties/GAG22K7MCD419/5277253",
+)
+
+connector = Device(
+    name="Connector",
+    manufacturer=Organization.OTHER,
```
saskiad: the manufacturer is on the digikey page. If we're adding this, it'd be great to use these fields rather than put it in the notes.

examples/thermistor_procedures.py
```diff
@@ -0,0 +1,87 @@
...
+    model="0533984002",
+    notes="Molex connector. Digi-Key part number: 900-0533984002CT-ND. "
+    "https://www.digikey.com/en/products/detail/molex/0533984002/15622916",
+)
```
saskiad: let's not put links to parts in the notes.

src/aind_data_schema/components/devices.py
```diff
@@ -779,3 +779,11 @@ class MyomatrixArray(Device):
...
 class Microscope(Device):
     """Description of a microscope"""
+
+
+class ThermistorAssembly(DataModel):
+    """Description of a thermistor assembly for temperature measurement"""
+
+    name: str = Field(..., title="Thermistor assembly name")
+    thermistor: Device = Field(..., title="Thermistor device")
+    connector: Device = Field(..., title="Connector device")
```
saskiad: how important is the connector as a device? I don't think we put this kind of info in a lot of other places. It'd expect this to be in a protocol on making the thermistor assembly, or as a subsection in a protocol on the thermistor surgery. I don't know that it makes sense to track this in the metadata.

src/aind_data_schema/components/surgery_procedures.py
```diff
@@ -134,6 +141,15 @@ def validate_size(self):
         return self
 
 
+class DeviceImplant(ProtocolMixin, DevicePosition, DataModel):
+    """Description of a generic device implant procedure"""
+
+    implanted_device: str = Field(
```
saskiad: won't we always need the implanted_device to take a Device object? So a generic device should take Device rather than be a string.

saskiad: And then is it possible to make it a discriminated list that can be Device or ThermistorAssembly and if we have other implanted devices that can't just be a device we could add them in the future?

src/aind_data_schema/components/devices.py
saskiad: and note: without connector, we don't need an assembly and can just use Device

saskiad (changes requested):
first, I don't think we need the connector for the assembly. But I'm willing to be persuaded.
second, I think we can just use the generic DeviceImplant with a discriminated list

src/aind_data_schema/components/surgery_procedures.py
dbirman: There's some reason that I made it a str... but now I don't remember what it was. I simplified to just DeviceImplant with `implanted_device: Device`. Hopefully that doesn't break anything

examples/thermistor_procedures.py
```diff
@@ -0,0 +1,73 @@
...
+from aind_data_schema.core.procedures import Procedures, Surgery
+from aind_data_schema_models.coordinates import AnatomicalRelative, AxisName, Direction, Origin
+from aind_data_schema_models.organizations import Organization
+from aind_data_schema_models.units import SizeUnit
+
+thermistor_wire = Device(
+    name="Thermistor wire",
+    manufacturer=Organization.TE_CONNECTIVITY,
+    model="GAG22K7MCD419",
+    notes="Digi-Key part number: GAG22K7MCD419. "
```
saskiad: can we remove having the part numebr in the notes with the link like this? I don't like setting bad precedence in our examples

examples/thermistor_procedures.py
```diff
@@ -0,0 +1,73 @@
...
+    "https://www.digikey.com/en/products/detail/te-connectivity-measurement-specialties/GAG22K7MCD419/5277253",
+)
+
+thermistor_coordinate_system = CoordinateSystem(
+    name="THERMISTOR",
+    origin=Origin.ORIGIN,
```
saskiad: how do we interpret the origin being origin?

saskiad (approved):
I think this is good. I still feel like ProbeConfig and DevicePosition are so similar that there could be a way to combine them, but I think it's not obvious...

Just clean up the stuff in the example so people don't copy that.

examples/thermistor_procedures.py
dbirman: Changed to tip

PR #1871, opened 2026-07-10
cloud compatible path usage

### LINKED ISSUE #1867: cloud-compatible Path usage
It should be easy to write metadata directly to a Path-like object supporting cloud storage, like from universal-pathlib. A few small tweaks are needed to allow this, like not directly creating a new Path inside recursive_check_paths. We may at some point want to change type annotations to reflect this support using https://github.com/barneygale/pathlib-abc, but for now we can just use the existing Path type annotations and rely on duck typing.

tmchartrand:
closes #1867 by using Path instance methods that are compatible with other Path-like objects, rather than stdlib functions that aren't. retains compatibility with path strings even though the type annotations don't support that usage.


dbirman (approved):
Lgtm

PR #1877, opened 2026-09-11
fix: allow merging empty string acquisition_type

### LINKED ISSUE #1876: Allow acquisition_type merging with ""

dbirman:
This is a hotfix to allow merging two acquisition_type fields when one is an empty string.


dougollerenshaw (approved):
Thanks for the quick turn around on this!

